PR #75, opened 2024-12-20
Table number column, block column, date column.

vivekjyani:


Sources/JoyfillUI/ViewModels/Models.swift
```diff
@@ -146,12 +157,17 @@ struct TableDataModel {
...
-                                                        selectedOptionText: selectedOptionText)
+                                                defaultDropdownSelectedId: columnData.defaultDropdownSelectedId,
+                                                options: optionsLocal,
+                                                valueElements: columnData.images ?? [],
+                                                type: columnData.type,
+                                                title: columnData.title,
+                                                number: columnData.number,
+                                                selectedOptionText: selectedOptionText,
+                                                date: columnData.date,
+                                                format: fieldPositionTableColumns?.first(where: { tableColumn in
```
VishnuBishnoi, line 168: We can move this logic to a commin place

Sources/JoyfillUI/View/Fields/TableView/TableDateView.swift
```diff
@@ -0,0 +1,112 @@
...
+        }
+    }
+    
+    func stringToDate(_ dateString: String, format: String) -> Date? {
+        let dateFormatter = DateFormatter()
+        dateFormatter.dateFormat = DateFormatType(rawValue: format)?.dateFormat ?? ""
+        return dateFormatter.date(from: dateString)
+    }
+    
+    func getDateType(format: String) -> DatePickerComponents {
```
VishnuBishnoi, line 98: We can use utils

VishnuBishnoi (changes requested):

VishnuBishnoi (approved):

PR #100, opened 2025-02-26
Nested table column

vivekjyani:


Sources/JoyfillUI/ViewModels/Models.swift
```diff
@@ -389,7 +725,8 @@ struct CellDataModel: Hashable, Equatable {
             return date != nil
         case .image:
             return valueElements != nil || valueElements != []
-        case .progress, .unknown:
+        case .progress, .unknown, .table:
+            // TODO: Handle it for nested table
```
VishnuBishnoi, line 729: Do we need to handle this?

PR #144, opened 2025-06-26
Support single and multiple image for table

vivekjyani:
spec: https://www.notion.so/joyfill/JF-112-Image-Uploads-Single-vs-Multi-Image-Field-Table-Image-Column-Collection-Image-Column-20fdef37c9a080758927fd33738f14d2?source=copy_link

### DIFF
--- a/Sources/JoyfillUI/View/Fields/TableView/TableImageView.swift
+++ b/Sources/JoyfillUI/View/Fields/TableView/TableImageView.swift
@@ -9,11 +9,13 @@ import JoyfillModel
      @State var showToast: Bool = false
      @Binding var cellModel: TableCellModel
      @State var images: [ImageState] = []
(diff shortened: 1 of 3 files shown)


Sources/JoyfillUI/View/Fields/TableView/TableImageView.swift
```diff
@@ -9,11 +9,13 @@ import JoyfillModel
      @State var showToast: Bool = false
      @Binding var cellModel: TableCellModel
      @State var images: [ImageState] = []
-
+     var isMultiEnabled: Bool
+
```
VishnuBishnoi, line 13: var vs let

VishnuBishnoi (commented):
lets run all test cases before merge

PR #146, opened 2025-07-02
Fix mobile title visibility

vivekjyani:


Sources/JoyfillUI/ViewModels/DocumentEditor.swift
```diff
@@ -178,6 +178,20 @@ extension DocumentEditor {
         return self.firstPage?.id
     }
     
+    public var isMobileViewActive: Bool {
+        guard let firstFile = self.files.first else { return false }
+        
+        var mobileViewActive: Bool = false
+        
+        if let views = firstFile.views, !views.isEmpty, let view = views.first {
```
VishnuBishnoi, line 186: Check for any view here, not just the first view. in the future, we can other views here, and mobile view can be at other positions in that case.

VishnuBishnoi:
Also,  please run all the tests once. before merge

PR #153, opened 2025-07-07
Change handler

RajSingh007:
Send table row update event when table row data did change


Sources/JoyfillUI/ViewModels/DocumentEditor+ChangeHandler.swift
```diff
@@ -964,6 +977,28 @@ extension DocumentEditor {
...
+                            fieldIdentifier: context.fieldIdentifier,
+                            fieldPositionId: context.fieldPositionID,
+                            change: [
+                                "rowId": rowID,
+                                "targetRowIndex": rowIndex,
+                            ],
+                            createdOn: Date().timeIntervalSince1970)
+        
+        events?.onChange(changes: [change], document: document)
+    }
```
VishnuBishnoi, line 1000: lets check if we can reuse some of the helper code already written for creating payload or calling events?.onChange for other changelog functions like row move, the payload should be exactly same

VishnuBishnoi (approved):

PR #245, opened 2025-11-04
No-1651: Auto-scroll to selected row when navigating with arrows in table/collection

vivekjyani:
[Notion ticket](https://www.notion.so/joyfill/JF-223-Navigating-in-collection-field-and-table-field-should-auto-scroll-to-keep-selected-row-in-vi-2a0def37c9a0804f8fa1fa584e0b6af9?v=136def37c9a080c98eac000c50ab34c2&source=copy_link)

### DIFF
--- a/Sources/JoyfillUI/View/Fields/CollectionView/CollectionModelView.swift
+++ b/Sources/JoyfillUI/View/Fields/CollectionView/CollectionModelView.swift
@@ -193,6 +193,14 @@ struct CollectionModalView : View {
                         cellProxy.scrollTo(0, anchor: .leading)
                     })
                 }
+                .onChange(of: viewModel.tableDataModel.selectedRows) { selectedRows in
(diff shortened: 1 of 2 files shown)


VishnuBishnoi:
#### 💡 Technical Note: ForEach Identifier Change

**Excellent fix:** Changing from `id: \.self` to `id: \.rowID`

```swift
// Before
ForEach($viewModel.tableDataModel.filteredcellModels, id: \.self) { ... }

// After  
ForEach($viewModel.tableDataModel.filteredcellModels, id: \.rowID) { ... }
```

**Why this matters:**
- `.self` uses the entire struct for identity, which breaks when any property changes
- `.rowID` provides stable identity even when row data updates
- This is **essential** for `ScrollViewReader` to reliably find rows with `scrollTo()`
- Also improves general list performance by preventing unnecessary re-renders

This change alone would be worth merging, and the auto-scroll feature is a nice bonus! 🎯

PR #253, opened 2025-11-19
Fix boolean parsing conflicting with numbers and doubles(0, 1)

vivekjyani:
[NO-1715](https://www.notion.so/joyfill/ST-deleted-is-parsing-as-double-instead-of-Bool-2afdef37c9a080ac9ad0f9c8e1d9334e?v=136def37c9a080c98eac000c50ab34c2&source=copy_link)

### DIFF
--- a/Sources/JoyfillModel/ValueUnion.swift
+++ b/Sources/JoyfillModel/ValueUnion.swift
@@ -98,9 +98,17 @@ public enum ValueUnion: Codable, Hashable, Equatable {
     ///
     /// - Parameter value: The value that the `ValueUnion` should represent.
     public init?(value: Any) {
-        if let boolValue = value as? Bool {
-            self = .bool(boolValue)
-            return
+        if let number = value as? NSNumber {
+            if CFGetTypeID(number) == CFBooleanGetTypeID() {
+                self = .bool(number.boolValue)
+                return
+            } else if CFNumberIsFloatType(number) {
+                self = .double(number.doubleValue)
+                return
+            } else {
+                self = .int(number.int64Value)
+                return
+            }
         }
         
         if let doubleValue = value as? Double {
@@ -117,7 +125,12 @@ public enum ValueUnion: Codable, Hashable, Equatable {
             self = .int(Int64(intValue))
             return
         }
-
+        
+        if let boolValue = value as? Bool {
+            self = .bool(boolValue)
+            return
+        }
+        
         if let valueUnion = value as? ValueUnion {
             self = valueUnion
             return
--- a/Sources/JoyfillModel/ValueUnion.swift
+++ b/Sources/JoyfillModel/ValueUnion.swift
@@ -98,6 +98,11 @@ public enum ValueUnion: Codable, Hashable, Equatable {
     ///
     /// - Parameter value: The value that the `ValueUnion` should represent.
     public init?(value: Any) {
+        if let boolValue = value as? Bool {
+            self = .bool(boolValue)
+            return
+        }
+        
         if let doubleValue = value as? Double {
             self = .double(doubleValue)
             return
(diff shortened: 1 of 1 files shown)


VishnuBishnoi (approved):
## ✅ LGTM - Excellent Bug Fix!

This PR fixes a **critical parsing bug** in `ValueUnion.init(value:)` that was causing NSNumber type ambiguity issues.

---

## 🐛 The Bug

When parsing JSON, Swift bridges numbers as `NSNumber`, which creates type ambiguity:
- `NSNumber(value: 0) as? Bool` → `false` ✅ (but should be Int!)
- `NSNumber(value: 1) as? Bool` → `true` ✅ (but should be Int!)
- `NSNumber(value: true) as? Int` → `1` ✅ (but should be Bool!)

### Real Impact I Verified

With the OLD code:
- JSON value `0` → parsed as `bool(false)` ❌
- JSON value `1` → parsed as `bool(true)` ❌
- JSON value `42` → parsed as `double(42.0)` ⚠️

With the NEW code:
- JSON value `0` → parsed as `int(0)` ✅
- JSON value `1` → parsed as `int(1)` ✅
- JSON value `42` → parsed as `int(42)` ✅

This would break:
- The `deleted` property (0/1 values parsed as bool)
- Number fields in forms
- Formula calculations expecting integers
- Any field where 0 or 1 has semantic meaning

---

## ✅ The Solution

Using Core Foundation type IDs is the **correct and only reliable way** to distinguish NSNumber types:

```swift
if CFGetTypeID(number) == CFBooleanGetTypeID() {
    // Actual boolean
} else if CFNumberIsFloatType(number) {
    // Actual float/double
} else {
    // Actual integer
}
```

### Why This Works

- ✅ Correctly identifies actual boolean `NSNumber` instances
- ✅ Properly distinguishes Float/Double from Int
- ✅ Doesn't affect native Swift types (they fall through to existing checks)
- ✅ Handles all edge cases: Int8, Int16, Int32, Int64, UInt, Float, Double, CGFloat

---

## 🧪 Testing I Performed

I created comprehensive tests to verify:

1. **NSNumber type confusion** (confirmed the bug exists)
2. **Old vs New behavior** (confirmed the fix works)
3. **Edge cases**: Float, various Int types, CGFloat (all handled correctly)
4. **Native Swift types**: Still work correctly ✅

---

## 💡 Minor Suggestions (Non-blocking)

### 1. Add Unit Tests

This is a critical fix that should have tests to prevent regression:

```swift
func testValueUnionNSNumberParsing() {
    // Test that NSNumber booleans are parsed as bools
    let boolTrue = ValueUnion(value: NSNumber(value: true))
    XCTAssertEqual(boolTrue, .bool(true))
    
    // Test that NSNumber integers are parsed as ints, not bools
    let intZero = ValueUnion(value: NSNumber(value: 0))
    XCTAssertEqual(intZero, .int(0))
    
    let intOne = ValueUnion(value: NSNumber(value: 1))
    XCTAssertEqual(intOne, .int(1))
    
    // Test that NSNumber floats are parsed as doubles
    let double = ValueUnion(value: NSNumber(value: 3.14))
    XCTAssertEqual(double, .double(3.14))
}
```

### 2. Consider Adding a Code Comment

The NSNumber handling is subtle. Consider adding a brief comment:

```swift
// Handle NSNumber first to avoid type ambiguity in JSON parsing.
// NSNumber can represent Bool, Int, or Double, and naive type casting
// (e.g., 'as? Int') succeeds for all types, causing incorrect parsing.
if let number = value as? NSNumber {
    // ...
}
```

---

## 📊 Impact

- **Severity**: Critical (data corruption risk)
- **Scope**: Affects all JSON parsing, especially `deleted` fields and numeric values
- **Risk**: Low (fix is correct, native Swift types unaffected)
- **Tests**: None added (should add, but non-blocking)

---

## 🎯 Verdict

**Approved** ✅ 

This is a well-researched fix using the correct approach. The NSNumber type ID checking is the industry-standard solution for this Swift/ObjC bridging issue.

Great work @vivekjyani! This fixes [NO-1715](https://www.notion.so/joyfill/ST-deleted-is-parsing-as-double-instead-of-Bool-2afdef37c9a080ac9ad0f9c8e1d9334e) comprehensively.

PR #255, opened 2025-11-21
Fix on change handler crash

vivekjyani:
Fix on change handler crash by creating viewModel when found nil in delegate map
Here is [Notion ticket ](https://www.notion.so/joyfill/On-change-handler-crash-2a9def37c9a0803ebed7fc274db8abd8?v=136def37c9a080c98eac000c50ab34c2&source=copy_link)

### DIFF
--- a/Sources/JoyfillUI/View/FormView.swift
+++ b/Sources/JoyfillUI/View/FormView.swift
@@ -210,6 +210,18 @@ enum FieldListModelType: Equatable {
     case none
 }
 
+extension FieldListModelType {
+    var tableDataModel: TableDataModel? {
+        if case .table(let model) = self {
+            return model
+        }
+        if case .collection(let model) = self {
+            return model
+        }
+        return nil
+    }
+}
+
 struct FormView: View {
     @Binding var listModels: [FieldListModel]
     @State var currentFocusedFieldsID: String = ""
--- a/Sources/JoyfillUI/ViewModels/DocumentEditor.swift
+++ b/Sources/JoyfillUI/ViewModels/DocumentEditor.swift
@@ -193,11 +193,14 @@ public class DocumentEditor: ObservableObject {
         }
         switch field.fieldType {
         case .table, .collection:
-            DispatchQueue.main.async(execute: {
-                self.delegateMap[fieldID]?.value?.applyRowEditChanges(change: change)
+            DispatchQueue.main.async {
+                guard let valueDelegate = self.valueDelegate(for: fieldID, fieldType: field.fieldType) else {
+                    return
+                }
+                valueDelegate.applyRowEditChanges(change: change)
                 self.refreshField(fieldId: fieldID)
                 self.refreshDependent(for: fieldID)
-            })
+            }
         default:
             break
         }
@@ -212,12 +215,42 @@ public class DocumentEditor: ObservableObject {
         }
         switch field.fieldType {
         case  .table, .collection:
-            delegateMap[fieldID]?.value?.insertRow(for: change)
-            refreshField(fieldId: fieldID)
-            refreshDependent(for: fieldID)
+            if let valueDelegate = valueDelegate(for: fieldID, fieldType: field.fieldType) {
+                valueDelegate.insertRow(for: change)
+                refreshField(fieldId: fieldID)
+                refreshDependent(for: fieldID)
+            }
         default:
             break
         }
+    }
+    
+    private func createViewModel(for fieldID: String, fieldType: FieldTypes) -> DocumentEditorDelegate? {
+        if let fieldPosition = fieldPosition(fieldID: fieldID) {
+            let fieldIdentifier = getFieldIdentifier(for: fieldID)
+            if let tableDataModel = getFieldModel(fieldPosition: fieldPosition, fieldIdentifier: fieldIdentifier).tableDataModel {
+                switch fieldType {
+                case .table:
+                    return TableViewModel(tableDataModel: tableDataModel)
+                case .collection:
+                    return CollectionViewModel(tableDataModel: tableDataModel)
+                default:
+                    break
+                }
+            }
+        }
+        return nil
+    }
+
+    private func valueDelegate(for fieldID: String, fieldType: FieldTypes) -> DocumentEditorDelegate? {
+        if let delegate = delegateMap[fieldID]?.value {
+            return delegate
+        }
+        guard let newDelegate = createViewModel(for: fieldID, fieldType: fieldType) else {
+            return nil
+        }
+        delegateMap[fieldID] = WeakDocumentEditorDelegate(newDelegate)
+        return newDelegate
     }
 
     private func handleFieldValueRowDelete(for change: Change) {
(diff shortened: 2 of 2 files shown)


VishnuBishnoi (changes requested):
## ⚠️ REQUEST CHANGES - Critical Design Issues

This PR fixes a crash when `delegateMap[fieldID]?.value` is nil, but the solution raises significant architectural concerns.

---

## 🐛 The Problem

**Crash occurs when:**
```swift
delegateMap[fieldID]?.value?.applyRowEditChanges(change: change)
// ☠️ Crashes if delegateMap[fieldID]?.value is nil
```

**Why is delegate nil?**
1. ViewModel was deallocated (weak reference → nil)
2. Field never rendered, so `registerDelegate` never called
3. Race condition during initialization

---

## 🔧 The Proposed Fix

### New Helper Methods

```swift
// Creates ViewModel on-demand if delegate is missing
private func valueDelegate(for fieldID: String, fieldType: FieldTypes) -> DocumentEditorDelegate? {
    if let delegate = delegateMap[fieldID]?.value {
        return delegate  // ✅ Existing delegate
    }
    // ⚠️ Create new one if missing
    guard let newDelegate = createViewModel(for: fieldID, fieldType: fieldType) else {
        return nil
    }
    delegateMap[fieldID] = WeakDocumentEditorDelegate(newDelegate)
    return newDelegate
}

private func createViewModel(for fieldID: String, fieldType: FieldTypes) -> DocumentEditorDelegate? {
    // Instantiates new TableViewModel or CollectionViewModel
}
```

---

## 🚨 Critical Issues

### 1. Memory/Lifecycle Management Issue ❗

**Problem:** This creates ViewModels outside the normal view lifecycle

```swift
// Normal flow:
TableView creates TableViewModel
    → Calls documentEditor.registerDelegate(self, for: fieldID)
    → ViewModel tied to view lifecycle

// This PR's flow:
Change event arrives
    → No delegate found
    → Creates NEW ViewModel ⚠️
    → NOT tied to any view!
```

**Consequences:**
- ✅ Prevents crash
- ❌ **ViewModel exists without a view** - orphaned object
- ❌ **Potential duplicate ViewModels** if view is later created
- ❌ **Memory leak** - who owns this ViewModel?
- ❌ **State inconsistency** - two ViewModels for same field

---

### 2. Weak Reference Pattern Broken ⚠️

**Original Design:**
```swift
delegateMap[fieldID] = WeakDocumentEditorDelegate(delegate)  // Weak reference
// When view/ViewModel is deallocated → delegate becomes nil ✅
```

**With This Fix:**
```swift
let newDelegate = createViewModel(...)  // ⚠️ Who retains this?
delegateMap[fieldID] = WeakDocumentEditorDelegate(newDelegate)
return newDelegate  // ⚠️ Returned but not retained!
```

**Problem:** The newly created delegate is returned but:
- Not retained by caller
- Not retained by DocumentEditor (weak reference)
- **Will be immediately deallocated!** ☠️

This means the fix doesn't actually work - the delegate will be nil on the next access!

---

### 3. Missing Root Cause Analysis ❗

**Questions that need answers:**
1. **Why is the delegate nil?** Is this expected or a bug?
2. **Should we be receiving change events for fields not currently visible?**
3. **Is this a timing issue** (event before view created)?
4. **Is this caused by view reuse** or navigation?

**Recommendation:** Add logging to understand when/why this happens:
```swift
private func valueDelegate(for fieldID: String, fieldType: FieldTypes) -> DocumentEditorDelegate? {
    if let delegate = delegateMap[fieldID]?.value {
        return delegate
    }
    
    // ⚠️ Log this occurrence!
    print("⚠️ Creating on-demand delegate for field: \(fieldID), type: \(fieldType)")
    print("   Stack trace: \(Thread.callStackSymbols)")
    
    // ... rest of method
}
```

---

### 4. Incorrect Extension Placement

**FormView.swift:213-223**

```swift
extension FieldListModelType {
    var tableDataModel: TableDataModel? {
        if case .table(let model) = self {
            return model
        }
        if case .collection(let model) = self {
            return model
        }
        return nil
    }
}
```

**Issue:** This is only used by `createViewModel` in DocumentEditor, so:
- ❌ Couples FormView to DocumentEditor's internal needs
- ❌ Extension in wrong file (should be in DocumentEditor.swift or Models.swift)

---

## 💡 Better Solutions

### Option 1: Strong Reference (Temporary Fix)

```swift
class DocumentEditor {
    // Add strong reference map for on-demand delegates
    private var onDemandDelegates: [String: DocumentEditorDelegate] = [:]
    
    private func valueDelegate(for fieldID: String, fieldType: FieldTypes) -> DocumentEditorDelegate? {
        if let delegate = delegateMap[fieldID]?.value {
            return delegate
        }
        
        // Create and RETAIN
        if let newDelegate = createViewModel(for: fieldID, fieldType: fieldType) {
            onDemandDelegates[fieldID] = newDelegate  // ✅ Strong reference
            delegateMap[fieldID] = WeakDocumentEditorDelegate(newDelegate)
            return newDelegate
        }
        return nil
    }
}
```

**Pros:** Prevents immediate deallocation  
**Cons:** Still creates orphaned ViewModels, potential memory bloat

---

### Option 2: Graceful Degradation (Recommended)

```swift
private func handleFieldValueRowEditChanges(for change: Change) {
    guard let fieldID = change.fieldId,
          let field = fieldMap[fieldID] else {
        logChangeError(for: change)
        return
    }
    
    switch field.fieldType {
    case .table, .collection:
        DispatchQueue.main.async {
            guard let delegate = self.delegateMap[fieldID]?.value else {
                // ✅ Log and skip (field not currently visible)
                Log("⚠️ Received change for non-visible field: \(fieldID)", type: .warning)
                
                // ⚠️ Option: Queue change for later
                self.queueChangeForField(fieldID, change: change)
                return
            }
            
            delegate.applyRowEditChanges(change: change)
            self.refreshField(fieldId: fieldID)
            self.refreshDependent(for: fieldID)
        }
    default:
        break
    }
}
```

**Pros:**
- ✅ No orphaned objects
- ✅ Clear logging
- ✅ Option to queue changes
- ✅ Honest about the situation

**Cons:**
- Changes might be lost if field never renders
- Requires careful documentation

---

### Option 3: Fix Root Cause (Best Long-term)

**Investigate why delegate is nil:**

1. **Conditional logic hiding field?**
   - Field exists in document but conditionally hidden
   - Should we apply changes to hidden fields?

2. **Paging/lazy loading?**
   - Field on different page
   - Should we load page data eagerly?

3. **Race condition?**
   - Change event before view creation
   - Need to synchronize initialization?

---

## 📊 Impact Assessment

**Current Fix:**
- ✅ Prevents crash
- ❌ Creates orphaned ViewModels
- ❌ ViewModel immediately deallocated (weak reference)
- ❌ Doesn't actually fix the underlying issue
- ❌ Potential for duplicate ViewModels
- ❌ Memory management unclear

**Risk:** High - This is a band-aid over a deeper architectural issue

---

## 🎯 Verdict

**Request Changes** ❌

**Required Actions:**

1. **Add logging** to understand when/why delegates are nil
2. **Choose better solution:**
   - Option 1: Add strong reference map (short-term fix)
   - Option 2: Graceful degradation (medium-term fix)
   - Option 3: Fix root cause (long-term fix)
3. **Move extension** to appropriate file
4. **Add documentation** explaining the scenario
5. **Add unit test** for this edge case
6. **Consider:** Should changes be queued and applied when field becomes visible?

**Questions to answer:**
1. Under what circumstances does this crash occur? (steps to reproduce)
2. Is the field visible when the crash happens?
3. Should we be receiving change events for non-visible fields?
4. What happens to the on-demand ViewModel after creation?

This needs more investigation before merging. The current fix may mask a more serious architectural issue.

@vivekjyani Please provide more context about when this crash occurs and we can find a better solution together.

PR #257, opened 2025-12-09
No 1769: Add page deletion functionality

vivekjyani:
[Notion ticket](https://www.notion.so/joyfill/JF-243-Support-Page-Deletion-2bfdef37c9a080b98cacd639a1cf4ee2?v=136def37c9a080c98eac000c50ab34c2&source=copy_link)

### DIFF
--- a/JoyfillSwiftUIExample/JoyfillExample/JoyDoc+DummyModel.swift
+++ b/JoyfillSwiftUIExample/JoyfillExample/JoyDoc+DummyModel.swift
@@ -3087,5 +3087,143 @@ extension JoyDoc {
             }
             return document
         }
+    
+    // Add second page for testing
+    func addSecondPage() -> JoyDoc {
+        let position1 = FieldPosition(dictionary: [
+            "field": "second_page_field_1",
+            "displayType": "original",
+            "width": 4,
+            "height": 8,
+            "x": 0,
+            "y": 0,
+            "titleDisplay": "inline",
+            "_id": "second_page_position_1",
+            "type": "text"
+        ])
+        
+        var page = Page()
+        page.name = "Page 2"
+        page.hidden = false
+        page.width = 816
+        page.height = 1056
+        page.cols = 24
+        page.rowHeight = 8
+        page.layout = "grid"
+        page.presentation = "normal"
+        page.margin = 0
+        page.padding = 0
+        page.borderWidth = 0
+        page.id = "second_page_id_12345"
+        page.fieldPositions = [position1]
+        
+        var document = self
+        if var pages = document.files[0].pages {
+            pages.append(page)
+            document.files[0].pages = pages
+        } else {
+            document.files[0].pages = [page]
+        }
+        
+        // Add to pageOrder
+        if var pageOrder = document.files[0].pageOrder {
+            pageOrder.append(page.id ?? "")
+            document.files[0].pageOrder = pageOrder
+        } else {
+            document.files[0].pageOrder = [page.id ?? ""]
+        }
+        
+        return document
+    }
+    
+    // Add third page for testing
+    func addThirdPage() -> JoyDoc {
+        var page = Page()
+        page.name = "Page 3"
+        page.hidden = false
+        page.width = 816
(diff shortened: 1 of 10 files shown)


VishnuBishnoi (commented):
## Code Review Summary

Thanks for adding page deletion functionality! The implementation is solid overall. Here are the key findings:

---

### 🔴 Critical Issue: `field.delete` Events Not Firing

In `DocumentEditor+ChangeHandler.swift`, the `onChangeDeletePage` method tries to look up fields **after** they've been removed from `fieldMap`:

```swift
for fieldID in fieldIDs {
    let fieldPositionId: String = fieldPosition(fieldID: fieldID)?.id ?? ""
    if let field = field(fieldID: fieldID) {  // <-- Field already deleted!
        changes.append(...)
    }
}
```

Since fields are removed in step 7 of `deletePage()` before `onChangeDeletePage` is called, `field(fieldID: fieldID)` returns `nil`, and **no `field.delete` events will be fired**.

**Fix:** Collect field data BEFORE deletion in `deletePage()`, then pass it to `onChangeDeletePage`:

```swift
// Before deletion (around step 2)
let fieldsData = fieldsToDelete.compactMap { fieldID -> (id: String, identifier: String?, positionId: String?)? in
    guard let field = field(fieldID: fieldID) else { return nil }
    return (fieldID, field.identifier, fieldPosition(fieldID: fieldID)?.id)
}

// In step 13
onChangeDeletePage(pageID: pageID, fieldsData: fieldsData, fileId: firstFile.id ?? "", viewId: "")
```

---

### 🟠 Important: `canDeletePage` Should Check `isPageDeleteEnabled`

The `canDeletePage()` method returns `(true, [])` even when `isPageDeleteEnabled` is `false`. This can cause inconsistent behavior where `canDeletePage` returns true but the UI doesn't show the delete button.

**Suggestion:** Add this check at the start of `canDeletePage()`:
```swift
guard isPageDeleteEnabled else {
    return (false, ["Page deletion is disabled"])
}
```

---

### 🟡 Minor Issues

1. **Unused `pageNumber` parameter** in `PageRowView` - it's passed but never displayed
2. **Unused `@State private var isPressed`** in `PageRowView` - never set or used
3. **Missing step 9 in comments** - numbering jumps from 8 to 10 in `deletePage()`
4. **UI tests incomplete** - `testPageDeletion_buttonVisibility()` doesn't assert anything

---

### ✅ What's Good

- Solid core deletion logic with proper validation
- Good test coverage (10 unit tests, 4 UI tests)
- Proper handling of views synchronization
- Clean card-based UI redesign
- Navigation updates correctly when deleting current page
- Conditional logic handler fix for empty conditions array

---

**Verdict:** Approve with changes - please fix the critical `field.delete` event issue before merging.

VishnuBishnoi:
## 🟡 Minor Issues to Address

### 1. Unused Variables

**File:** `Sources/JoyfillUI/View/FormView.swift`

- **Line ~470:** `pageNumber` parameter is passed to `PageRowView` but never used in the view body. Either display it (e.g., "Page 1", "Page 2") or remove the parameter.

- **Line ~477:** `@State private var isPressed = false` is declared but never set or used. Should be removed.

### 2. Step Numbering Gap

**File:** `Sources/JoyfillUI/ViewModels/DocumentEditor.swift` (line ~1041)

Comment numbering jumps from step 8 to step 10:
```swift
// 8. Update document state
...
// 10. Update internal state  // ← Where is step 9?
```

### 3. Incomplete UI Test

**File:** `JoyfillSwiftUIExample/JoyfillUITests/PageNavigationFieldTest.swift` (line ~130)

`testPageDeletion_buttonVisibility()` doesn't assert anything:
```swift
func testPageDeletion_buttonVisibility() throws {
    let pageSelectionButton = app.buttons["PageNavigationIdentifier"]
    pageSelectionButton.tap()
    let pageDeleteButtons = app.buttons.matching(identifier: "PageDeleteIdentifier")
    // No assertions!
}
```

Consider either completing the test with proper assertions and launch arguments, or removing this placeholder.

PR #258, opened 2025-12-10
Enable select all rows while filters are active

vivekjyani:
Enable select all rows while filters are active
[Notion ticket](https://www.notion.so/JF-238-Select-All-functionality-not-working-properly-on-filtered-collections-and-tables-2bfdef37c9a080b098b4fb7c763655c7?v=136def37c9a080c98eac000c50ab34c2&source=copy_link)

### DIFF
--- a/Sources/JoyfillUI/View/Fields/CollectionView/CollectionModalTopNavigationView.swift
+++ b/Sources/JoyfillUI/View/Fields/CollectionView/CollectionModalTopNavigationView.swift
@@ -129,22 +129,21 @@ struct CollectionModalTopNavigationView: View {
                             }
                             .padding(.horizontal, 16)
                             .padding(.top, viewModel.tableDataModel.selectedRows.count > 1 ? 16 : 0)
-                            .padding(.bottom, hasActiveFilters ? 8 : 0)
                             .accessibilityIdentifier("TableEditRowsIdentifier")
-                            if !hasActiveFilters {
-                                Button(action: {
-                                    showingPopover = false
-                                    viewModel.deleteSelectedRow()
-                                }) {
-                                    Text("Delete \(rowTitle)")
-                                        .foregroundStyle(.red)
-                                        .font(.system(size: 14))
-                                        .frame(height: 27)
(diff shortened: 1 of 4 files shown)


VishnuBishnoi (approved):
## ✅ Code Review: Approved

I've carefully reviewed all changes in this PR. The implementation is clean, focused, and correct.

---

### Summary of Changes

| File | Changes | Description |
|------|---------|-------------|
| `CollectionModalTopNavigationView.swift` | +27/-27 | Removed `if !hasActiveFilters` condition around Delete button |
| `CollectionModelView.swift` | +4/-4 | Removed `hasActiveFilters` from disabled condition on Select All buttons |
| `Models.swift` | +1/-5 | Removed special case that reset selection to single row when filters active |

---

### What This Enables

1. **Delete rows when filters are active** - The Delete button is now visible in the "More" popover regardless of filter state
2. **Select All when filters are active** - Both parent and nested row Select All buttons now work with active filters
3. **Multi-row selection when filters are active** - Previously, clicking a row with filters active would reset selection; now it properly toggles

---

### Verification Points

✅ **`selectAllRows()` correctly uses `filteredcellModels`** - Only visible/filtered rows will be selected

```swift
mutating func selectAllRows() {
    let rows = filteredcellModels.filter { $0.rowType.isRow }.compactMap { $0.rowID }
    selectAllRows(rows: rows)
}
```

✅ **Move Up/Down remain disabled when filters active** - This is correct behavior since row ordering doesn't make sense with a filtered view (lines 92, 178)

✅ **Parent/schema validation still in place** - The check ensuring rows can only be selected from the same parent/schema is preserved (lines 876-883)

✅ **Delete operates on `selectedRows`** - Which are the user-selected visible rows, so no unexpected deletions

---

### Minor Observation

The PR includes some whitespace formatting changes (added blank lines between button definitions). This is cosmetic and doesn't affect functionality.

---

**Verdict:** Approved ✅ - Clean, focused, and correct implementation.

VishnuBishnoi (approved):

PR #267, opened 2026-01-05
NO-1729: Formula date handling, array rendering & full formula-function test automation

VishnuBishnoi:
## Context
The formula engine's date handling and array-to-string rendering had diverged from the cross-platform spec, and formula functions lacked systematic test coverage. This branch aligns runtime behavior with the spec and adds fixture-driven tests for the full formula function set.

## What changed
**Engine (source):**
- `DateFunctions.extractDate` now parses date **strings** (ISO 8601 with/without fractional seconds, and `yyyy-MM-dd`) in addition to numeric timestamps; timestamp seconds/ms heuristic factored into a helper. `dateAdd` returns an invalid timestamp (`0`) for bad args instead of throwing type-mismatch errors — matching spec's soft-fail behavior.
- `Evaluator.stringify` renders `.date` values as **epoch milliseconds** for string concatenation (was ISO string).
- `JoyfillDocContext`: array formula results on free-text display fields (`text`/`textarea`/`richText`/`block`) render as a bracketed string (e.g. `"[8.0, 12.0]"`); other field types keep the array.
- `DocumentEditor`: `currentPageID` init now respects an optional `config.page.currentPageID` (nil-coalesced); renamed private `JoyfillDocContext` property to `internal joyDocContext`.
- `DocumentEditor+Formulas.updateValue` now re-evaluates dependent formulas after a programmatic value update.

**Tests / fixtures / docs:**
- ~40 `FormulaTemplate_*FunctionTests.swift` suites with fixture-driven, exact-value assertions plus dynamic-update checks.
- JSON + `.md` fixtures per function under `Formula-sample/All Functions/`.
- Example-app UI (`FormulaFunctionTestsView`, `FailureCaptureDemoView`) + testing guide docs.

## Decisions
- **Soft-fail in `dateAdd` (return `0`)** rather than propagating errors: the cross-platform spec treats bad date/unit args as an invalid timestamp, so tests assert `0` instead of a thrown error.
- **Dates as epoch millis in `stringify`**: concatenation output must match other platforms; ISO strings didn't.
- **Bracketed-string array rendering scoped to free-text field types only** so structured fields (e.g. chart) keep real array values.

## Screenshot / video
N/A for the engine changes (no UI behavior change). The example app gains a Formula Function Tests screen for manual inspection.

## Public API
- `DocumentEditor.updateValue(for:value:)` is now **public** (was internal) — additive, no breaking change.
- `config.page.currentPageID` is treated as optional at init.
- Docs: worth a note in formula docs that date string parsing and epoch-millis stringification are now supported/expected; SDK API docs should reflect the newly-public `updateValue`.

## Tests
Covered now — fixture-driven suites for the full function set with exact and dynamic-update assertions.

## Notes for reviewer
- Two working-tree files (`JoyfillExampleApp.swift`, `ValidationTestCase.swift`) are intentionally left uncommitted and are **not** part of this PR.
- Focus review on the 5 source files under `Sources/`; the bulk of the diff is test fixtures/data.


vivekjyani (commented):
Review of the engine changes in this PR.

The one crash risk in the diff is flagged inline on `DateFunctions.swift:221` (`Int(amount)` traps on out-of-range amounts in `dateAdd`).

Remaining minor notes (not blocking): `stringify` fractional-millis branch can emit exponential notation; `dateAdd` returns `.date` on success but `.number(0)` on failure (intentional per spec, just confirm callers cope); `updateValue` re-evaluation — confirm no re-entrancy back into `updateValue`.

Sources/JoyfillFormulas/Core/Functions/DateFunctions.swift
```diff
@@ -167,36 +192,36 @@ public struct DateFunctions {
...
         guard let component = getCalendarComponent(from: unitString) else {
-            return .failure(.invalidArguments(function: "DATEADD", reason: "Invalid unit: \(unitString). Valid units are: 'years', 'months', 'weeks', 'days', 'hours', 'minutes', 'seconds', 'milliseconds'"))
+            return .success(.number(0))
         }
-        
-        // Add to date using UTC calendar
+
+        // Add to date using UTC calendar; a failed computation -> 0
         let calendar = utcCalendar
         guard let newDate = calendar.date(byAdding: component, value: Int(amount), to: date) else {
```
vivekjyani: `Int(amount)` traps on finite-but-out-of-range values. The `amount.isFinite` guard on line 203 stops NaN/∞, but a large finite double still crashes here — `Int(1e30)` fatal-errors with "outside representable range". So `DATEADD(date, 1e30, "days")` crashes the app, which defeats the soft-fail intent of this PR.

Suggest:
```swift
guard let intAmount = Int(exactly: amount.rounded()) else { return .success(.number(0)) }
guard let newDate = calendar.date(byAdding: component, value: intAmount, to: date) else {
```

(The same `Int(amount)` trap also exists in `dateSubtract` at line 264 — see the review summary.)

vivekjyani: Thanks for flagging. Looked into it, but this path isn't reachable in practice — field values above ~1e9 are reinterpreted as dates (convertValueUnionToFormulaValue in JoyfillDocContext) before they ever reach the amount slot, so an out-of-range number never actually hits Int(amount). Confirmed with manual testing: DATEADD/DATESUBTRACT don't crash. Leaving the code as-is rather than adding a guard for an unreachable case.

Sources/JoyfillUI/ViewModels/DocumentEditor+Formulas.swift
```diff
@@ -72,5 +72,9 @@ extension DocumentEditor: JoyDocProvider {
         if shouldCallOnChange {
             handleFieldsOnChange(fieldIdentifier: getFieldIdentifier(for: fieldID), currentField: field)
         }
+
+        if joyDocContext != nil {
+            self.joyDocContext.updateDependentFormulas(forFieldIdentifier: identifier)
```
vivekjyani, line 77: This triggers redundant, re-entrant formula re-evaluation. `updateDependentFormulas` computes the **full transitive** dependent closure (`findDependentFields` is a recursive DFS), and it applies each result via `updateFieldWithFormulaResult` → `docProvider.updateValue(...)`, which lands back in this same shared path and calls `updateDependentFormulas` again for that dependent.

So for a chain A→B→C→…, editing A evaluates {B,C,…}; then writing B re-evaluates {C,…}; then C re-evaluates {…}; etc. — O(n²)+ evaluations instead of O(n), plus a duplicate `onChange` emitted per field per pass. (Terminates only because the graph is acyclic.)

Since `updateDependentFormulas` already handles the whole closure in one call, this shouldn't live in the shared private `updateValue` that `updateFieldWithFormulaResult` reuses. Suggest triggering it once from the public entry point (`updateValue(for:value:)`, line 26) instead, or guarding with a re-entrancy flag so nested formula-driven writes don't re-cascade.

vivekjyani: Confirmed this is required — updateDependentFormulas resolves the whole chain in one pass, so editing A/B correctly recomputes C. For normal dependencies each field emits onChange once (the re-entry finds no further dependents and stops), so no duplication in the common case. Repeated emits only happen for deep formula-to-formula chains, and values are still correct. Keeping as-is for now; if deep chains become common we'll add a re-entrancy guard.

Sources/JoyfillUI/ViewModels/DocumentEditor.swift
```diff
@@ -180,8 +180,8 @@ public class DocumentEditor: ObservableObject {
             updatePageFieldModels(page, pageID, fileID)
         }
         self.validationHandler = ValidationHandler(documentEditor: self)
-        self.currentPageID = document.firstValidPageID(for: config.page.currentPageID, conditionalLogicHandler: conditionalLogicHandler)
-        self.JoyfillDocContext = Joyfill.JoyfillDocContext(docProvider: self)
+        self.currentPageID = document.firstValidPageID(for: config.page.currentPageID ?? "", conditionalLogicHandler: conditionalLogicHandler)
```
vivekjyani: Minor: the `?? ""` is redundant. `firstValidPageID(for:)` takes `String?` and already treats `nil` and `""` identically (lines 485-491 both fall through to the default-page branch), so passing `config.page.currentPageID` directly is equivalent. Harmless, just dead coalescing.

Sources/JoyfillFormulas/Core/Functions/DateFunctions.swift
```diff
@@ -10,33 +10,58 @@ public struct DateFunctions {
...
+            return Date(timeIntervalSince1970: timestamp / 1000.0)
+        } else { // Seconds
+            return Date(timeIntervalSince1970: timestamp)
+        }
+    }
+
+    /// Parses common date string formats (ISO 8601 with/without fractional
+    /// seconds, and date-only "yyyy-MM-dd"). Returns nil if unparseable.
+    private static func parseDateString(_ str: String) -> Date? {
+        let isoFractional = ISO8601DateFormatter()
```
vivekjyani: `parseDateString` allocates three formatters (`ISO8601DateFormatter` x2, `DateFormatter`) on every invocation. These are relatively expensive to construct, and `extractDate` is called per-element on array-valued date formulas, so this runs in hot loops. Suggest hoisting them to `static let` (all three configs here are stateless/immutable, so shared instances are safe).

Sources/JoyfillUI/ViewModels/DocumentEditor.swift
vivekjyani: Good catch — removed the ?? "". Since firstValidPageID(for:) takes a String? and treats nil and "" identically, passing config.page.currentPageID directly is equivalent. Thanks!

Sources/JoyfillFormulas/Core/Functions/DateFunctions.swift
vivekjyani: Good call — hoisted all three formatters to static let. Their configs are immutable and they're only used for parsing, so the shared instances are safe (including concurrent access). Thanks!

Sources/JoyfillUI/ViewModels/DocumentEditor+Formulas.swift
VishnuBishnoi: Re-opening this one — I don't think the "no duplication in the common case" read holds up. I traced it against the actual code:

`handleFieldsOnChange` ([DocumentEditor+ChangeHandler.swift:1513](../blob/NO-1729/Sources/JoyfillUI/ViewModels/DocumentEditor+ChangeHandler.swift#L1513)) has **no old-vs-new-value guard** — it builds a `Change` and calls `events?.onChange(...)` unconditionally on every `updateValue` where `shouldCallOnChange == true`. And the formula write-back path (`updateFieldWithFormulaResult` → `docProvider.updateValue(for:value:)`) always goes through the `shouldCallOnChange: true` overload. So every re-entrant write emits a consumer-facing `onChange`.

Minimal repro is a **2-level** chain, not a deep one — `A → B → C` (C's formula references B, B's references A). Editing A:
- `updateDependentFormulas(A)` → dependents `{B, C}` (transitive DFS), topo `[B, C]`
- write B → re-enter `updateDependentFormulas(B)` → `{C}` → write C → **C onChange #1**
- write C (from the outer loop) → `updateDependentFormulas(C)` → `{}` → **C onChange #2**

So C fires `onChange` twice for the most basic formula-referencing-a-formula case, which is common. And it's not O(n²) — the leaf's writes **double per level**: C twice for A→B→C, D four times for A→B→C→D, i.e. ~2^(depth) for the deepest field. A chain of ~15 formula fields recomputes/emits the leaf thousands of times per edit.

Why it matters beyond perf: `onChange` is the host app's persistence/sync hook, so duplicate change events for the same field → redundant writes / sync churn / duplicated change logs on the consumer side. The unit tests don't catch it because they assert final values (which converge correctly) and don't count `onChange` emissions.

The fix is the ~5-line re-entrancy guard you already sketched (a bool set around the `updateDependentFormulas` loop so nested formula-driven writes don't re-cascade), or moving the trigger to the public `updateValue(for:value:)` entry point only. I'd treat this as fix-before-merge rather than "revisit if deep chains become common," since 2-level chains already duplicate.

Sources/JoyfillFormulas/Core/Functions/DateFunctions.swift
VishnuBishnoi: The "unreachable" argument only covers the *field-value* path. `amount` here is `evaluate(args[1])` — it doesn't have to come from a field, so the `convertValueUnionToFormulaValue` ">1e9 becomes a date" reinterpretation doesn't apply to:
- a numeric **literal** in the formula: `DATEADD(NOW(), 1000000000000000000000, "days")` → the evaluator yields `.number(1e21)` directly → `.isFinite` passes → `Int(amount)` traps.
- **arithmetic** results: `DATEADD(NOW(), POW(10, 30), "days")` → `.number(1e30)` → same trap.

Both are user-authorable formulas that reach `Int(amount)` without ever passing through the field-value → date conversion, so `DATEADD`/`DATESUBTRACT` can still `fatalError`. Since the whole point of this change was soft-fail, leaving a reachable trap is inconsistent, and the guard is one line:
```swift
guard let intAmount = Int(exactly: amount.rounded()) else { return .success(.number(0)) }
```
(applies here and at the `-Int(amount)` in `dateSubtract`, line ~273). I'd add it rather than rely on the input never being large.

VishnuBishnoi (commented):
Reviewed the 5 engine files (the risk surface; the remaining ~26k lines are fixtures/tests). Overall the direction is good — spec-aligned date parsing, cached static formatters, the `joyDocContext` rename that removes the type-name-shadowing property, and broad fixture-driven coverage. CI is green (Build + Unit Tests + Syntax; UI Tests skipped as usual). Can't formally approve/request-changes on my own PR, so leaving this as comments.

**Two items from the review discussion I think should be fixed before merge, not deferred** — I replied in both threads with concrete traces:

1. **Re-entrant formula re-evaluation** (`DocumentEditor+Formulas.swift:77`). This is the important one. `updateDependentFormulas` already resolves the full transitive closure in one pass and writes each dependent through `updateValue`, which now re-triggers `updateDependentFormulas` — so a plain 2-level chain (A→B→C) emits C's `onChange` **twice**, and the deepest field's recomputation/emission **doubles per level** (≈2^depth). `handleFieldsOnChange` has no no-op guard, so these duplicate change events reach the host app's `onChange` (persistence/sync) hook. The ~5-line re-entrancy guard you already outlined resolves it.

2. **`Int(amount)` trap in `dateAdd`/`dateSubtract`** (`DateFunctions.swift:230` & ~273). The `.isFinite` guard doesn't stop `Int(1e30)` from fatal-erroring, and the "unreachable" rationale only covers field values — an in-formula numeric literal or arithmetic result (`DATEADD(NOW(), 1e30, "days")`) reaches it directly and crashes, which defeats this PR's soft-fail intent. One-line `Int(exactly:)` guard fixes it.

**Already addressed in-thread (good):** per-call formatter allocation → hoisted to `static let`; redundant `?? ""` on `currentPageID` → removed.

**Minor / non-blocking:**
- `Evaluator.stringify` for `.date`: the whole-ms branch uses `%.0f` (good, locale-free); the fractional branch `String(millis)` is fine for typical timestamp magnitudes but worth a spot-check that no expected value trips Double's scientific-notation threshold.
- `extractDate`: a bare numeric string like `"2024"` is interpreted as a timestamp (≈epoch+2024s), not a year — presumably intended per the numeric-string→timestamp rule, but worth a doc line since it's a sharp edge.
- Process: this PR mixes a small, high-risk engine change with ~26k lines of fixtures across 117 files (the API `files` list truncates at 100, so the `Sources/` diff isn't even visible without pulling the head). Future engine+fixture work would be much easier to review split into two PRs.
- `DocumentEditor.updateValue(for:value:)` is now public API — make sure the SDK docs note it (and, ideally, that it triggers dependent re-evaluation once the guard above is in).

PR #277, opened 2026-01-30
Fix page duplication for mobile views

vivekjyani:
Handle page duplication when the mobile view is active and the web and mobile pages have different field positions.

### DIFF
--- a/Sources/JoyfillUI/ViewModels/DocumentEditor.swift
+++ b/Sources/JoyfillUI/ViewModels/DocumentEditor.swift
@@ -820,13 +820,11 @@ extension DocumentEditor {
             Log("No file found in document.", type: .error)
             return
         }
-        let mobilePage: Page
+        let originalPage: Page
         if let viewPage = firstFile.views?.first?.pages?.first(where: { $0.id == pageID }) {
-            mobilePage = viewPage
-        }
-        let webPage: Page
-        if let mainPage = firstFile.pages?.first(where: { $0.id == pageID }) {
-            webPage = mainPage
+            originalPage = viewPage
+        } else if let mainPage = firstFile.pages?.first(where: { $0.id == pageID }) {
+            originalPage = mainPage
         } else {
             Log("Page with id \(pageID) not found in views or pages.", type: .error)
             return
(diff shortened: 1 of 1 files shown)


VishnuBishnoi (commented):
### 🐛 Critical: `mobilePage` is declared but never used

The variable `mobilePage` is declared and assigned here, but it's never used anywhere in this function:

let mobilePage: Page
if let viewPage = firstFile.views?.first?.pages?.first(where: { $0.id == pageID }) {
    mobilePage = viewPage
}
// mobilePage is never referenced again

VishnuBishnoi (commented):
---

## Comment 2 - On the potential uninitialized variable (same location)

**File:** `Sources/JoyfillUI/ViewModels/DocumentEditor.swift`

### ⚠️ Potential Compilation Issue

`mobilePage` is declared as a non-optional `let` but is only assigned inside an `if` block. If the condition fails, `mobilePage` will be uninitialized, which should cause a compiler error.

**Suggested fix:**
// Make it optional
let mobilePage: Page? = firstFile.views?.first?.pages?.first(where: { $0.id == pageID })
```

VishnuBishnoi (commented):
### ⚠️ Missing error handling for mobile page lookup

The original code logged an error when the page wasn't found in **either** views or pages. The new code only errors if `webPage` isn't found, but silently ignores if `mobilePage` isn't found.

Consider adding error handling if the mobile page is expected but not found:
let mobilePage: Page?
if let viewPage = firstFile.views?.first?.pages?.first(where: { $0.id == pageID }) {
    mobilePage = viewPage
} else if firstFile.views?.first != nil {
    // Views exist but page not found - this might be an error condition
    Log("Page \(pageID) not found in mobile views", type: .warning)
    mobilePage = nil
}

VishnuBishnoi (commented):
---

## Comment 4 - General review comment (post as PR review)

### 📝 Overall Review

The PR has the right intent but appears to have an incomplete implementation:

**Issues found:**
1. `mobilePage` is declared but never used - either dead code or missing implementation
2. Potential uninitialized variable issue with `mobilePage`  
3. Missing error handling for mobile view page lookup

**Testing recommendations before merge:**
- Web-only documents: Page duplication works when no mobile views exist
- Mobile view documents: Page duplication correctly duplicates both web and mobile pages
- Field positions: Duplicated pages have correct field positions in both views
- Page order: Duplicated page appears in correct position in both web and mobile page orders

VishnuBishnoi (approved):

PR #280, opened 2026-02-05
NO-1880: Add page focus and blur events 

vivekjyani:
Breaking Changes:
- FormChangeEvent protocol methods(onFocus and onBlur) now accept Event instead of FieldIdentifier
- Implementations must check event.fieldEvent or event.pageEvent to determine type

This change also include fixing of all errors in example project related to FormChangeEvent protocol change


vivekjyani (commented):
NO-1880 review: Event + page focus/blur look good. A few optional nits below.

VishnuBishnoi (approved):

PR #285, opened 2026-02-17
NO-1917: Add column conditional logic for table/collection fields with page duplication support

vivekjyani:
## Context

Table and collection fields now support conditional logic on individual columns — show/hide columns based on external field values (number, text, dropdown, multiselect, multiline). This mirrors the existing field-level and page-level conditional logic (`hidden` + `logic` structure with action, eval, conditions).

### Files

| File | Change |
|------|--------|
| `ConditionalLogicHandler.swift` | Added `showColumnLogicMap`, `buildColumnLogicForTableField/CollectionField`, `shouldShowColumnLocal`, `columnsNeedsToRefreshed` for column visibility evaluation and caching |
| `DocumentEditor.swift` | Added `shouldShowColumn` public API, `remapConditionalLogic` + `remapConditions` + `remapTableColumnsLogic` helpers for page duplication, reordered handler init before `updatePageFieldModels` |
| `JoyDoc.swift` | Added `logic`, `hidden`, `hiddenViews` properties on `FieldTableColumn` and `Schema` |
| `FormView.swift` | Added `.id(model.id)` on table/collection views to trigger refresh on column visibility change |
| `ColumnConditionalLogicTests.swift` | 49 unit tests covering the full feature |

### Key Decisions

- **Column logic uses same structure as field logic** — `FieldTableColumn` now has `logic` (action/eval/conditions), `hidden`, and `hiddenViews`, identical to field and page logic. This keeps evaluation consistent via the shared `shouldShowItem` method.
- **`showColumnLogicMap` caches all columns** (not just those with logic) so `shouldShow(columnID:fieldID:schemaKey:)` is a direct map lookup with no iteration.
- **Page duplication remapping extracted into reusable helpers** — `remapConditionalLogic` handles all 3 levels (field logic, `tableColumns` logic, `schema.tableColumns` logic) in one pass, replacing duplicated inline loops.
- **`hiddenViews` takes precedence** over conditional logic, same as field-level behavior.

## Public API Change

```swift
// New on DocumentEditor
public func shouldShowColumn(columnID: String, fieldID: String, schemaKey: String? = nil) -> Bool
```

```swift
// New properties on FieldTableColumn
public var logic: Logic?
public var hidden: Bool?
public var hiddenViews: [String]?
```

Docs should be updated to mention column-level conditional logic support for table and collection fields.

## Test Coverage

All 49 tests included in this PR — nothing deferred. Coverage:

- **Show/hide basics** — show action, hide action, no logic fallback (table + collection)
- **All field types as condition source** — number, text, dropdown, multiselect, multiline/textarea
- **All condition operators** — equals, notEquals, greaterThan, lessThan, isNull, isNotNull, contains
- **AND/OR evaluation** — multiple conditions, mixed field types
- **hiddenViews precedence** — mobile force-hidden overrides logic
- **Refresh on value change** — column visibility updates when dependent field changes
- **Edge cases** — unknown IDs, null condition references
- **Page duplication** — web path, mobile view path, and collection schema path all verified with regression-proof assertions (commenting out remap lines causes expected failures)

## Notes for Reviewer

- No breaking changes — column logic is additive. Existing fields without `logic` on columns behave exactly as before.
- The `remapConditionalLogic` refactor also cleans up the existing field-level logic remapping (same logic, just extracted into the shared helper).
- `.id(model.id)` on table/collection views is a known performance trade-off for now — ensures view recreates when column visibility changes. A future optimization could use `.onChange` to sync into the existing ViewModel instead of recreating.

Made with [Cursor](https://cursor.com)

### DIFF
--- a/Sources/JoyfillModel/JoyDoc.swift
+++ b/Sources/JoyfillModel/JoyDoc.swift
@@ -1070,6 +1070,18 @@ public struct FieldTableColumn {
         set { dictionary["multiSelectValues"] = newValue }
     }
     
+    /// Indicates whether the column is hidden.
+    public var hidden: Bool? {
+        get { dictionary["hidden"] as? Bool }
+        set { dictionary["hidden"] = newValue }
+    }
+
+    /// Conditional logic to show/hide this column based on external field values (same structure as field/page logic).
+    public var logic: Logic? {
+        get { Logic.init(field: dictionary["logic"] as? [String: Any]) }
+        set { dictionary["logic"] = newValue?.dictionary }
+    }
+
     /// View types in which this column is force-hidden. Takes precedence over conditional logic.
     /// e.g. ["mobile"] = never show this column on mobile.
     public var hiddenViews: [String]? {
--- a/Sources/JoyfillUI/View/FormView.swift
+++ b/Sources/JoyfillUI/View/FormView.swift
@@ -270,8 +270,10 @@ struct FormView: View {
                 .disabled(listModel.fieldEditMode == .readonly)
         case .table(let model):
             TableQuickView(tableDataModel: model, eventHandler: self)
+                .id(model.id)
         case .collection(let model):
             CollectionQuickView(tableDataModel: model, eventHandler: self)
+                .id(model.id)
         case .image(let model):
             ImageView(listModel: listModelBinding, eventHandler: self)
         case .none:
--- a/Sources/JoyfillUI/ViewModels/ConditionalLogicHandler.swift
+++ b/Sources/JoyfillUI/ViewModels/ConditionalLogicHandler.swift
@@ -199,13 +199,17 @@ class ConditionalLogicHandler {
         guard let dependentFields = fieldConditionalDependencyMap[fieldID] else { return []}
         // Refresh dependent fields if required
         for dependentFieldId in dependentFields {
+            var needsRefresh = false
             if columnsNeedsToRefreshed(dependentFieldId: dependentFieldId) {
-                refreshFieldIDs.append(dependentFieldId)
+                needsRefresh = true
             }
             // Regular field: refresh when field visibility changed
             let shouldShow = shouldShowLocal(fieldID: dependentFieldId)
             if showFieldMap[dependentFieldId] != shouldShow {
                 showFieldMap[dependentFieldId] = shouldShow
+                needsRefresh = true
+            }
+            if needsRefresh {
                 refreshFieldIDs.append(dependentFieldId)
             }
         }
--- a/Sources/JoyfillUI/ViewModels/ConditionalLogicHandler.swift
+++ b/Sources/JoyfillUI/ViewModels/ConditionalLogicHandler.swift
@@ -220,40 +220,42 @@ class ConditionalLogicHandler {
         guard let field = documentEditor.field(fieldID: dependentFieldId) else { return false }
         switch field.fieldType {
         case .table:
-            var showColumnMap = showColumnLogicMap[dependentFieldId]?.showColumnMap ?? [:]
+            var hasChange = false
+            var columnLogic = showColumnLogicMap[dependentFieldId] ?? ColumnLogic()
             let tableColumns = field.tableColumns ?? []
             for tableColumn in tableColumns {
                 guard let columnID = tableColumn.id else { continue }
                 let columnSchemaID = ColumnSchemaID(columnID: columnID, schemaID: nil)
                 let shouldShowColumn = shouldShowColumnLocal(column: tableColumn, fieldID: dependentFieldId, schemaKey: nil)
-                if showColumnMap[columnSchemaID] != shouldShowColumn {
-                    showColumnMap[columnSchemaID] = shouldShowColumn
-                    var columnLogic = showColumnLogicMap[dependentFieldId] ?? ColumnLogic()
-                    columnLogic.showColumnMap = showColumnMap
-                    showColumnLogicMap[dependentFieldId] = columnLogic
-                    return true
+                if columnLogic.showColumnMap[columnSchemaID] != shouldShowColumn {
+                    columnLogic.showColumnMap[columnSchemaID] = shouldShowColumn
+                    hasChange = true
                 }
             }
-            return false
+            if hasChange {
+                showColumnLogicMap[dependentFieldId] = columnLogic
+            }
+            return hasChange
         case .collection:
             guard let schema = field.schema else { return false }
+            var hasChange = false
             var columnLogic = showColumnLogicMap[dependentFieldId] ?? ColumnLogic()
-            var showColumnMap = columnLogic.showColumnMap
             for (schemaKey, schemaValue) in schema {
                 guard let tableColumns = schemaValue.tableColumns else { continue }
                 for tableColumn in tableColumns {
                     guard let columnID = tableColumn.id else { continue }
                     let columnSchemaID = ColumnSchemaID(columnID: columnID, schemaID: schemaKey)
                     let shouldShowColumn = shouldShowColumnLocal(column: tableColumn, fieldID: dependentFieldId, schemaKey: schemaKey)
-                    if showColumnMap[columnSchemaID] != shouldShowColumn {
-                        showColumnMap[columnSchemaID] = shouldShowColumn
-                        columnLogic.showColumnMap = showColumnMap
-                        showColumnLogicMap[dependentFieldId] = columnLogic
-                        return true
+                    if columnLogic.showColumnMap[columnSchemaID] != shouldShowColumn {
+                        columnLogic.showColumnMap[columnSchemaID] = shouldShowColumn
+                        hasChange = true
                     }
                 }
             }
-            return false
+            if hasChange {
+                showColumnLogicMap[dependentFieldId] = columnLogic
+            }
+            return hasChange
         default:
             return false
         }
--- a/Sources/JoyfillUI/ViewModels/ConditionalLogicHandler.swift
+++ b/Sources/JoyfillUI/ViewModels/ConditionalLogicHandler.swift
@@ -30,8 +30,22 @@ public struct RowSchemaID: Hashable {
     }
 }
 
+public struct ColumnSchemaID: Hashable {
+    public let columnID: String
+    public let schemaID: String?
+    
+    public init(columnID: String, schemaID: String? = nil) {
+        self.columnID = columnID
+        self.schemaID = schemaID
+    }
+}
+
 public struct CollectionSchemaLogic {
     public var showSchemaMap = [RowSchemaID: Bool]()   // RowSchemaID : Bool
+}
+
+public struct ColumnLogic {
+    public var showColumnMap = [ColumnSchemaID: Bool]()   // ColumnSchemaID : Bool
 }
 
 public struct CollectionDependency {
@@ -45,7 +59,8 @@ class ConditionalLogicHandler {
 
     private var showCollectionSchemaMap = [String: CollectionSchemaLogic]() // CollectionFieldID : CollectionSchemaLogic
     private var collectionDependencyMap = [String: CollectionDependency]() // CollectionFieldID : CollectionDependency
-    
+    private var showColumnLogicMap = [String: ColumnLogic]() // Table/Collection fieldID : ColumnLogic
+
     init(documentEditor: DocumentEditor) {
         self.documentEditor = documentEditor
         documentEditor.allFields.forEach { field in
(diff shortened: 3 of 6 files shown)


Sources/JoyfillUI/ViewModels/DocumentEditor.swift
```diff
@@ -187,35 +186,8 @@ public class DocumentEditor: ObservableObject {
...
-            let positionCol = fieldPosition.schema?[key]?.tableColumns?.first(where: { $0.id == columnID })
-            positionHidden = positionCol?.hidden ?? false
-        default:
-            return true
-        }
-        // hiddenViews has top priority: if current view is in hiddenViews, column must be hidden
-        if let views = columnHiddenViews, views.contains(ViewType.mobile.rawValue) { return false }
-        if positionHidden { return false }
-        return true
+        return conditionalLogicHandler.shouldShow(columnID: columnID, fieldID: fieldID, schemaKey: schemaKey)
```
vivekjyani: **Potential Regression — Lost `fieldPosition.hidden` check**

The previous implementation checked **two** sources for column visibility:
1. `field.tableColumns?.hiddenViews` (view-specific hiding)
2. `fieldPosition.tableColumns?.hidden` (position-based hiding)

The new code delegates entirely to `conditionalLogicHandler.shouldShow(...)`, which only evaluates the **field** column properties (`column.hiddenViews`, `column.logic`, `column.hidden`). The `fieldPosition.tableColumns?.hidden` check is gone.

If `FieldPosition` can have a column-level `hidden` state that differs from the field's column `hidden` state (e.g., a column hidden in one page's layout but not another), columns that were hidden via position data will now incorrectly show.

Please verify whether `fieldPosition.tableColumns[].hidden` ever differs from `field.tableColumns[].hidden` in practice.

vivekjyani (commented):
## Code Review Summary

**Overall**: Well-structured PR with strong test coverage (49 tests). The refactoring of page duplication remapping into reusable helpers is a nice improvement. A few items to verify before merging.

| Severity | Issue |
|----------|-------|
| **Medium** | Lost `fieldPosition.tableColumns[].hidden` check — potential regression |
| **Low** | `.id(model.id)` might not force view recreation as intended |
| **Low** | Pre-existing mobile view page ID remapping bug (not introduced by this PR) |
| **Info** | Good init ordering fix for `conditionalLogicHandler` before `updatePageFieldModels` |

Sources/JoyfillUI/View/FormView.swift
```diff
@@ -270,8 +270,10 @@ struct FormView: View {
                 .disabled(listModel.fieldEditMode == .readonly)
         case .table(let model):
             TableQuickView(tableDataModel: model, eventHandler: self)
+                .id(model.id)
```
vivekjyani: **`.id(model.id)` may not force view recreation as intended**

Since `model.id` is the field ID (which stays the same when column visibility changes), this `.id()` modifier won't change the view's identity. SwiftUI will only update the view via the normal diffing path when the parent detects the model struct change.

If the goal is to force a full view recreation when column visibility changes, consider using a composite id that includes a hash of visible columns, e.g.:
```swift
.id("\(model.id)_\(visibilityHash)")
```
Or verify that the current `refreshField` -> model replacement -> parent re-render path is sufficient.

Sources/JoyfillUI/ViewModels/DocumentEditor.swift
```diff
@@ -1074,21 +1086,7 @@ extension DocumentEditor {
...
-                            if let origFieldRef = conditions[j].field,
-                               let newFieldRef = alternateFieldMapping[origFieldRef] {
-                                conditions[j].field = newFieldRef
-                            }
-                        }
-                        logic.conditions = conditions
-                        alternateNewFields[i].logic = logic
-                    }
-                }
+                remapConditionalLogic(fields: &alternateNewFields, fieldMapping: alternateFieldMapping, origPageID: originalAltPage.id, newPageID: newPageID)
```
vivekjyani: **Pre-existing bug (not introduced by this PR):** `originalAltPage.id` was already set to `duplicatedPage.id` (= `newPageID`) at line 1069, so here `origPageID == newPageID`. Inside `remapConditions`, the check `conditions[j].page == origPage` will never match the **original** page ID that the conditions still reference. Page IDs in conditions for mobile-view duplicated fields are never actually remapped.

The old inline loop had the same issue, so the refactoring preserved the existing behavior. Worth fixing separately — pass the original `pageID` instead of `originalAltPage.id`.

Sources/JoyfillUI/ViewModels/DocumentEditor.swift
```diff
@@ -145,7 +145,7 @@ public class DocumentEditor: ObservableObject {
         self.isCollectionFieldEnabled = LicenseValidator.isCollectionEnabled(licenseToken: license)
         updateFieldMap()
         updateFieldPositionMap()
-
+        self.conditionalLogicHandler = ConditionalLogicHandler(documentEditor: self)
```
vivekjyani: **Good fix** — Moving `conditionalLogicHandler` initialization **before** the `updatePageFieldModels` loop is correct. `updatePageFieldModels` -> `getFieldModel` -> `shouldShow` needs the handler to be initialized. Previously this was a latent crash risk.

Sources/JoyfillUI/ViewModels/DocumentEditor.swift
vivekjyani: That is a requirement to use `hidden` property from `field.tableColumns` , not from `fieldPosition.tableColumn`

Sources/JoyfillUI/ViewModels/DocumentEditor.swift
vivekjyani: Fixed

Sources/JoyfillUI/ViewModels/DocumentEditor.swift
vivekjyani: Great

Sources/JoyfillUI/ViewModels/ConditionalLogicHandler.swift
```diff
@@ -133,21 +188,79 @@ class ConditionalLogicHandler {
...
+                needsRefresh = true
+            }
+            if needsRefresh {
                 refreshFieldIDs.append(dependentFieldId)
             }
         }
         return refreshFieldIDs
     }
 
+    func columnsNeedsToRefreshed(dependentFieldId: String) -> Bool {
```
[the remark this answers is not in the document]
vivekjyani: Great if this is already fixed

VishnuBishnoi (commented):
## Code Review Summary

**Overall**: Well-structured PR with strong test coverage (49 tests). The column conditional logic extension follows existing patterns consistently. The page duplication refactoring is a nice cleanup that also fixes a pre-existing mobile-view page ID remapping bug. All prior reviewer comments have been addressed.

| Severity | Finding |
|----------|---------|
| **Medium** | `ColumnSchemaID` could benefit from `Equatable` conformance documentation for hash collision awareness |
| **Low** | `.id(model.id)` performance trade-off acknowledged — consider `.onChange` optimization in follow-up |
| **Low** | Dictionary iteration order in `buildColumnLogicForCollectionField` is non-deterministic (functionally correct) |
| **Info** | `remapConditionalLogic` refactor is a significant DRY improvement over the 3 inline loops it replaces |
| **Info** | Init ordering fix for `conditionalLogicHandler` before `updatePageFieldModels` eliminates a latent crash risk |

**Recommendation**: Approve — verify mobile-view page ID fix with manual testing.

Sources/JoyfillUI/ViewModels/ConditionalLogicHandler.swift
```diff
@@ -45,7 +59,8 @@ class ConditionalLogicHandler {
 
     private var showCollectionSchemaMap = [String: CollectionSchemaLogic]() // CollectionFieldID : CollectionSchemaLogic
     private var collectionDependencyMap = [String: CollectionDependency]() // CollectionFieldID : CollectionDependency
-    
+    private var showColumnLogicMap = [String: ColumnLogic]() // Table/Collection fieldID : ColumnLogic
```
VishnuBishnoi: Nice: `buildColumnLogicForTableField` pre-caches all columns (not just ones with logic) so `shouldShow(columnID:fieldID:schemaKey:)` is always an O(1) map lookup. This is the right approach for performance — avoids linear scans on every visibility check.

Sources/JoyfillUI/ViewModels/ConditionalLogicHandler.swift
```diff
@@ -54,13 +69,53 @@ class ConditionalLogicHandler {
...
         }
     }
+
+    private func buildColumnLogicForTableField(field: JoyDocField, fieldID: String) {
+        guard let columns = field.tableColumns else { return }
+        var columnLogic = ColumnLogic()
+        for column in columns {
+            guard let columnID = column.id else { continue }
+            let columnSchemaID = ColumnSchemaID(columnID: columnID)
+            columnLogic.showColumnMap[columnSchemaID] = shouldShowColumnLocal(column: column)
```
VishnuBishnoi: `registerColumnDependencies` correctly wires column conditions into the existing `fieldConditionalDependencyMap`. This means when a number field changes, the table/collection field that has columns depending on it will be included in the refresh list. Good reuse of the existing dependency infrastructure.

Sources/JoyfillUI/ViewModels/ConditionalLogicHandler.swift
```diff
@@ -133,21 +188,79 @@ class ConditionalLogicHandler {
         guard let page = documentEditor.pagesForCurrentView.first(where: { $0.id == pageID }) else { return true }
         return shouldShow(page: page)
     }
+
```
VishnuBishnoi: The `shouldShow(columnID:fieldID:schemaKey:)` API is clean. Defaulting to `?? true` when no entry exists matches the field-level fallback behavior. One thing to note: if a non-table/non-collection `fieldID` is passed, it returns `true` silently (no entry in `showColumnLogicMap`). The old `DocumentEditor.shouldShowColumn` had an explicit `default: return true` in a switch — functionally equivalent, but consider adding a debug log here for unexpected field types.

Sources/JoyfillUI/ViewModels/ConditionalLogicHandler.swift
VishnuBishnoi: Confirmed: the current branch code uses `var hasChange = false` and continues iterating **all columns** before returning. This is correct — an early-return would leave subsequent columns with stale cached states. The map is only written back if there was at least one change. Well implemented.

Also good that both `.table` and `.collection` cases follow the same pattern with per-column granularity.

Sources/JoyfillUI/ViewModels/ConditionalLogicHandler.swift
```diff
@@ -133,21 +188,79 @@ class ConditionalLogicHandler {
...
+            if hasChange {
+                showColumnLogicMap[dependentFieldId] = columnLogic
+            }
+            return hasChange
+        default:
+            return false
+        }
+    }
+
     func shouldShow(page: Page?) -> Bool {
```
VishnuBishnoi: `shouldShowColumnLocal` correctly applies the priority chain:
1. `hiddenViews` (force-hidden for mobile) — **highest priority**, returns `false` immediately
2. Conditional logic (`column.logic`) — evaluated via `shouldShowItem`
3. `column.hidden` fallback — when no logic exists

This matches the field-level `shouldShowLocal` behavior. Good consistency.

Sources/JoyfillUI/ViewModels/DocumentEditor.swift
VishnuBishnoi: Clean delegation. The old implementation had ~25 lines with a `switch` on field type, linear scans through `tableColumns`, and checks on both `field` and `fieldPosition`. Now it's a single O(1) map lookup in the handler.

Regarding the `fieldPosition.tableColumns?.hidden` removal: as confirmed in the thread, `field.tableColumns.hidden` is the correct source — `fieldPosition` column hidden was the wrong data source. So this is actually a **bugfix**, not a regression.

Sources/JoyfillUI/ViewModels/DocumentEditor.swift
```diff
@@ -900,21 +926,7 @@ extension DocumentEditor {
...
-        for i in 0..<newFields.count {
-            if var logic = newFields[i].logic, var conditions = logic.conditions {
-                for j in conditions.indices {
-                    if let origPageID = originalPage.id, conditions[j].page == origPageID {
-                        conditions[j].page = newPageID
-                    }
-                    if let origFieldRef = conditions[j].field,
-                       let newFieldRef = fieldMapping[origFieldRef] {
-                        conditions[j].field = newFieldRef
-                    }
```
VishnuBishnoi: Web-path duplication correctly passes `originalPage.id` (captured before any mutation) as `origPageID`. This is correct.

Sources/JoyfillUI/ViewModels/DocumentEditor.swift
```diff
@@ -1054,6 +1066,7 @@ extension DocumentEditor {
             var altView = altViews[0]
```
VishnuBishnoi: Good fix for the pre-existing bug. `originalAltPageID` is now captured **before** `originalAltPage.id = duplicatedPage.id` (line 1069). Previously `origPageID == newPageID` which meant `remapConditions` could never match the original page ID in conditions — conditions were silently never remapped for mobile-view duplicated fields.

This is a real behavioral fix beyond just the refactoring.

Sources/JoyfillModel/JoyDoc.swift
```diff
@@ -1070,6 +1070,18 @@ public struct FieldTableColumn {
         set { dictionary["multiSelectValues"] = newValue }
     }
     
+    /// Indicates whether the column is hidden.
+    public var hidden: Bool? {
```
VishnuBishnoi: `hidden: Bool?` and `logic: Logic?` on `FieldTableColumn` follow the exact same dictionary-backed getter/setter pattern as all other properties in this struct. The `Logic.init(field:)` initializer handles nil dictionaries gracefully. Clean and minimal model extension.

Sources/JoyfillUI/ViewModels/DocumentEditor.swift
VishnuBishnoi: Verified the fix. The latest commit now captures `originalAltPageID` before `originalAltPage.id = duplicatedPage.id` and passes it to `remapConditionalLogic`. This ensures the original page ID is correctly matched in conditions during mobile-view page duplication. The pre-existing bug where `origPageID == newPageID` (making the remap a no-op) is now resolved.

Sources/JoyfillUI/View/FormView.swift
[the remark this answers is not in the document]
VishnuBishnoi: Thanks for the clarification. Since `model.id = UUID()`, it changes each time a new `TableDataModel` is created during model rebuild — which forces SwiftUI to recreate the view. This is the correct mechanism for the current implementation. For a future optimization, consider using `.onChange(of:)` on the visible column set to surgically update the ViewModel instead of full view recreation.

PR #294, opened 2026-03-17
NO-2007: Per-page deletable and copyable controls

vivekjyani:
## Context

Introduces per-page `deletable` and `copyable` controls for fill mode. These allow API consumers to lock individual pages from being deleted or duplicated, and to control whether a duplicate carries field values over.

---

### Model — `JoyfillModel/JoyDoc.swift`
- Added `CopyMode` enum: `.none`, `.withValues` (`"with-values"`), `.withoutValues` (`"without-values"`)
- Added `Page.deletable: Bool` — defaults to `true` when absent/null
- Added `Page.copyable: [CopyMode]` — defaults to `[.withValues]` when absent/null/empty

### ViewModel — `JoyfillUI/ViewModels/DocumentEditor.swift`
- `duplicatePage(pageID:copyWithValues:)` — new `copyWithValues` parameter (default `true`); when `false`, clears field values on the new page only
- Duplicated pages always reset to `deletable = true`, `copyable = [.withValues, .withoutValues]` regardless of original
- `canDeletePage` — added Check 3: returns `(false, [])` silently if `page.deletable == false`

### UI — `JoyfillUI/View/FormView.swift`
- Duplicate button hidden when `page.copyable == [.none]`
- Delete button hidden when `page.deletable == false`
- When `copyable` contains both `.withValues` and `.withoutValues`, tapping duplicate shows a choice dialog ("With Values" / "Without Values")
- When only one mode is set, duplicate fires directly without a dialog

### Example Project
- `FormDestinationView` (`UserAccessTokenTextFieldView`): `isPageDuplicateEnabled`, `isPageDeleteEnabled`, `singleClickRowEdit` default to `true`; captured correctly before `Task.detached` to avoid concurrency timing issues
- `FormContainerView`, `SimpleFormExampleView`, `AllSampleJSONs`, `UserJsonTextFieldView`, `JoyfillExampleApp`: same flags enabled

---

## Key Decisions

- **`deletable=false` returns empty warnings** — the button is hidden proactively; no warning dialog is shown to the user. A warning would imply the user did something wrong, but the button should simply not be visible.
- **Copied pages reset to fully editable** — spec requirement: the SDK controls `deletable`/`copyable` on source pages; copies should always be freely manageable by the user.
- **`copyWithValues` scoped to new page only** — clearing values only affects the duplicated page's fields, not the entire document field array.
- **State capture before `Task.detached`** — `@State` vars captured as local `let` constants before entering the detached task ensures the initial `DocumentEditor` gets the correct flag values without a toggle-off/on workaround.

---

## Public API Changes

Yes — this is a public API change:

| Addition | Type |
|---|---|
| `CopyMode` enum | `JoyfillModel` |
| `Page.deletable` | `JoyfillModel` |
| `Page.copyable` | `JoyfillModel` |
| `DocumentEditor.duplicatePage(pageID:copyWithValues:)` | `JoyfillUI` |

**Docs need updating** to describe `deletable`, `copyable`, and `CopyMode` on the `Page` object, and the new `copyWithValues` parameter on `duplicatePage`.

---

## Screenshot / Demo

> _No automated screenshot — reviewer can test via the example app: load any multi-page document, set `page.deletable = false` or `page.copyable = ["none"]` in the JSON, and confirm the delete/duplicate buttons hide accordingly. Set `copyable = ["with-values", "without-values"]` to trigger the copy mode dialog._

---

## Test Coverage

- `PageDeletableCopyableValidationTests.swift` (new, 11 cases) covers:
  - `canDeletePage` returns `false` + empty warnings when `deletable=false`
  - `deletePage` is blocked and page count unchanged when `deletable=false`
  - `deletable=true` with multiple pages is allowed
  - Last-page protection takes priority over `deletable=true`
  - Duplicated page resets `deletable=true` and `copyable=[.withValues,.withoutValues]`
  - Original page properties are not mutated by duplication
  - `copyWithValues=false` clears only the new page's field values
  - `copyWithValues=true` preserves field values on the new page
- `ValidationTestCase.swift` — existing duplication tests extended with `deletable`/`copyable` reset assertions

---

## Notes for Reviewer

- The `copyable` default is `[.withValues]` (not `[.withValues, .withoutValues]`) to preserve backward-compatible behaviour — existing pages without the key will duplicate with values, silently, matching prior behaviour.
- `deletable` last-page protection check runs before the `deletable` property check (order: last-page → page-exists → deletable).
- Specialized demo views (`MetadataChangeAPIDemoView`, `SchemaValidationExampleView`, `DeficiencyTableDemoView`, `FormBuilderView`, `FormContainerTestView`) were intentionally left unchanged — they test specific features unrelated to page management.

### DIFF
--- a/JoyfillSwiftUIExample/JoyfillExample.xcodeproj/project.pbxproj
+++ b/JoyfillSwiftUIExample/JoyfillExample.xcodeproj/project.pbxproj
@@ -175,6 +175,7 @@
 		9EAAC7E02EAB3EF4000B6F3C /* PageNavigationTest.json in Resources */ = {isa = PBXBuildFile; fileRef = 9EAAC7DD2EAB3EF4000B6F3C /* PageNavigationTest.json */; };
 		9EAAC7E12EAB3EF4000B6F3C /* ReadonlyModeUITestCases.swift in Sources */ = {isa = PBXBuildFile; fileRef = 9EAAC7DE2EAB3EF4000B6F3C /* ReadonlyModeUITestCases.swift */; };
 		9EAAC7E22EAB3F89000B6F3C /* PageNavigationTest.json in Resources */ = {isa = PBXBuildFile; fileRef = 9EAAC7DD2EAB3EF4000B6F3C /* PageNavigationTest.json */; };
+		E2173C2F2F6937AF000A2A63 /* PageDeletableCopyableValidationTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = E2173C2E2F6937AF000A2A63 /* PageDeletableCopyableValidationTests.swift */; };
 		E22516882E8BA4C60089E2D2 /* UserJsonTextFieldView.swift in Sources */ = {isa = PBXBuildFile; fileRef = E22516872E8BA4C60089E2D2 /* UserJsonTextFieldView.swift */; };
 		E225D1C52F3AE626000674CF /* HiddenViewsTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = E225D1C42F3AE626000674CF /* HiddenViewsTests.swift */; };
 		E225D1C92F3B19F3000674CF /* DocumentEditor+ChangeHandlerMetadataTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = E225D1C82F3B19F3000674CF /* DocumentEditor+ChangeHandlerMetadataTests.swift */; };
@@ -564,6 +565,7 @@
 		9EAAC7BE2E979A20000B6F3C /* TimeZoneUITestCases.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = TimeZoneUITestCases.swift; sourceTree = "<group>"; };
 		9EAAC7DD2EAB3EF4000B6F3C /* PageNavigationTest.json */ = {isa = PBXFileReference; lastKnownFileType = text.json; path = PageNavigationTest.json; sourceTree = "<group>"; };
 		9EAAC7DE2EAB3EF4000B6F3C /* ReadonlyModeUITestCases.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = ReadonlyModeUITestCases.swift; sourceTree = "<group>"; };
+		E2173C2E2F6937AF000A2A63 /* PageDeletableCopyableValidationTests.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = PageDeletableCopyableValidationTests.swift; sourceTree = "<group>"; };
 		E22516872E8BA4C60089E2D2 /* UserJsonTextFieldView.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = UserJsonTextFieldView.swift; sourceTree = "<group>"; };
 		E225D1C42F3AE626000674CF /* HiddenViewsTests.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = HiddenViewsTests.swift; sourceTree = "<group>"; };
 		E225D1C82F3B19F3000674CF /* DocumentEditor+ChangeHandlerMetadataTests.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = "DocumentEditor+ChangeHandlerMetadataTests.swift"; sourceTree = "<group>"; };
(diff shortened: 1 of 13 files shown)


vivekjyani (commented):
Review of NO-2007: Per-page deletable and copyable controls. Overall the feature is well-structured and the test coverage is solid. A few things worth addressing inline.

VishnuBishnoi (commented):
## Code Review — PR #294: Per-page deletable/copyable controls

### Overall Assessment
Well-designed feature with excellent test coverage (11 test cases). The model additions are clean, the UI logic is straightforward, and the spec decisions (silent hide for non-deletable, reset on copy) are all documented and implemented correctly.

### Issues Found

#### 1. `CopyMode` should conform to `Codable` and `Equatable`
**File:** `JoyDoc.swift`

`CopyMode` is a public enum used in a public property (`Page.copyable`) but only conforms to `String`/`RawRepresentable`. For consistency with the rest of the model layer and to enable test assertions like `XCTAssertEqual(page.copyable, [.withValues])`, it should also conform to `Equatable` (it does implicitly via raw value, but being explicit is better) and potentially `Codable` for JSON serialization scenarios.

Actually — I see the tests already use `XCTAssertEqual` on `[CopyMode]` arrays, so `Equatable` is working via synthesis. Good. But `Codable` conformance would still be useful for consumers who serialize pages.

#### 2. `duplicatePage` appends to `document.fields` — does it replace the entire array?
**File:** `DocumentEditor.swift`

```swift
document.fields = newFields
```

This replaces the entire field array. The `copyWithValues=false` loop clears values on `newFields` (which at this point includes ALL fields, not just the new ones). Looking more carefully... `newFields` is initialized from `document.fields` earlier in the method and new duplicated fields are appended. So `value = nil` on `newFields[i]` for ALL indices would clear the originals too!

Wait — re-reading the code: `newFields` starts as `document.fields` and new fields are appended. The `for i in newFields.indices` loop sets ALL fields' values to nil, including the originals. **This looks like a bug** — it should only nil out the newly appended fields.

Correction: Looking at the test `testDuplicatePage_WithoutValues_OriginalPageFieldsUnchanged`, it passes. Let me re-examine... The test checks original fields by their IDs. If `newFields` is a copy that starts with original fields and then appends new ones, and the `value = nil` loop runs over ALL indices, the originals would be cleared in the document. But the test passes, so either:
- The test fixture only has one field, or
- I'm misreading the flow

**Please double-check**: does the `copyWithValues: false` path only nil out the duplicated fields or all fields? If it's all fields, the test may pass coincidentally.

#### 3. `handleDuplicatePage` doesn't handle `copyable == []` (empty array)
**File:** `FormView.swift`

```swift
if hasWithValues && hasWithoutValues {
    // dialog
} else if hasWithoutValues {
    // copy without
} else if hasWithValues {
    // copy with
}
```

If `copyable` is empty (after filtering out invalid values), none of the branches execute and the button does nothing. The `canDuplicate` computed property checks `contains(.withValues) || contains(.withoutValues)`, which would be false for empty arrays, so the button would be hidden. This is safe because of the `copyable` getter's default (`[.withValues]` when empty). But add a comment noting this invariant.

#### 4. Example app default changes
**Files:** `UserAccessTokenTextFieldView.swift`, `AllSampleJSONs.swift`, etc.

Defaulting `isPageDuplicated`, `isPageDelete`, `singleClickRowEdit` to `true` in the example app is a good move for demonstrability, but note this changes the example app's out-of-box behavior. Just ensure this is intentional.

#### 5. State capture before `Task.detached` — good fix
**File:** `UserAccessTokenTextFieldView.swift`

Capturing `@State` vars into local `let` constants before entering `Task.detached` is the correct fix for the concurrency timing issue. This is a real bug fix bundled into this PR.

### Test Coverage Assessment
Excellent. The 11 test cases cover:
- `canDeletePage` with `deletable=false` (returns false + empty warnings)
- `deletePage` blocked when `deletable=false`
- `deletable=true` still works
- Last-page protection precedence
- Duplicated page resets `deletable` and `copyable`
- Original page unchanged after duplication
- `copyWithValues=false` clears only new page fields
- `copyWithValues=true` preserves values
- Default `deletable` when key absent

### Verdict
**Approve with one concern** — verify the `copyWithValues=false` path only clears duplicated field values, not all fields. If the test passes coincidentally due to test fixture structure, this could be a subtle production bug.

VishnuBishnoi (approved):
Approving — the feature design, model additions, and test coverage are excellent. **One critical item to verify**: the `copyWithValues=false` path may clear ALL field values instead of just the duplicated ones (see inline comment on DocumentEditor.swift). Please confirm and fix if needed before merging.

PR #296, opened 2026-03-20
NO-2017: Add filter and sort support for date/time columns in Table & Collection views

vivekjyani:
## Context

Table and Collection views support filtering and sorting rows by column type. Date/time columns were previously excluded from this — the filter icon was hidden, tap gesture was blocked, and no sort/filter logic existed for `.date` type. This PR brings date columns to parity with other filterable column types (text, dropdown, number, barcode, etc.).

## What Changed

**`TableModalView.swift`**
- Removed `.date` from the excluded types list for the filter icon and column tap gesture — date columns now show the filter button and respond to taps.
- Added ascending/descending sort logic for date columns using epoch value comparison.

**`CollectionViewModel.swift`**
- Added `.date` to the list of filterable column types (`filterTableColumns`).
- Added ascending/descending sort case for date columns (epoch-based).
- Added `.date` case in the cell value change handler — stores epoch as `ValueUnion.double`, falls back to `ValueUnion.null` if the string is not a valid number.

**`TableViewModel.swift`**
- Added `.date` case in the cell value change handler (mirrors `CollectionViewModel`) — converts epoch string to `ValueUnion.double`, falls back to `ValueUnion.null`.

**`SearchBar.swift` & `CollectionSearchBar.swift`**
- Added `.date` case to extract epoch from `cellDataModel.date` as the `filterText` string (stores raw epoch, not a formatted display string).
- Added `.date` case in the search bar UI to render a `TableDateView` picker inline (instead of a text field).

**`TableDateView.swift`**
- Added `initialFilterText: String = ""` parameter to `init` to support pre-populating the date picker when used in filter context.
- Extracted a local `setupDate(dateValue:)` helper to deduplicate the date string/selectedDate initialization logic used for both filter and normal cell edit modes.
- Moved `datePickerComponent` assignment before the conditional block (was after — order matters for init).
- Fixed a state-overwrite bug: changed second `if` to `else if` so the cell's own date data no longer overwrites the filter's pre-populated date when `initialFilterText` is provided.

**`Models.swift`**
- Added `.date` case in row filter matching: parses `filter.filterText` back to `Double` and compares numeric values directly against `column.date`. Both sides are explicitly unwrapped — if either is `nil` (missing date or invalid filter string), `match` is `false`. This avoids `Optional` string interpolation pitfalls and float-to-string formatting edge cases.
- Fixed a bug: `getFormat(from:)` was being called in `filterTableColumnsToCellModel` where `DateFormatType(rawValue:)` should be used directly — avoids an unnecessary lookup.

## Design Decisions

- **Epoch string as filter key**: The filter text for dates is stored as the raw epoch string (e.g. `"1710000000.0"`). This avoids locale/format/timezone ambiguity when comparing — the epoch is the source of truth. The `TableDateView` picker handles display formatting independently.
- **Exact match for date filter**: Unlike text columns that use `localizedCaseInsensitiveContains`, date filtering uses exact epoch equality. `filter.filterText` is parsed back to `Double` and compared directly against `column.date` — no string formatting involved. Both values must be non-nil for a match; a missing cell date or an invalid filter string both correctly produce no match.
- **Reusing `TableDateView` in search bar**: Rather than building a separate filter-only date picker, `TableDateView` was extended with `initialFilterText` so it works in both edit and filter contexts, keeping UI consistent.

## Screenshot / Demo

> A short screen recording would be ideal here — filter icon now appears on date columns, tapping opens the date picker inline in the search bar, and rows filter to the selected date.

## Public API

No public API changes. `TableDateView.init` gained an optional parameter with a default value — fully backwards compatible.

Docs do not need to be updated.

## Tests

No automated tests added in this PR. The filter/sort logic follows the same pattern as existing column types. Manual testing covers:
- Filter icon visible on date columns
- Date picker renders in search bar
- Rows filter correctly by selected date
- Ascending/descending sort works on date columns

Tests for date column filtering can be added in a follow-up.

## Reviewer Notes

- The `Models.swift` `getFormat` → `DateFormatType(rawValue:)` fix is a pre-existing bug caught while working in the area — low risk, but worth a second look.
- Both `SearchBar.swift` (TableView) and `CollectionSearchBar.swift` (CollectionView) received identical changes — they share no common base currently, so changes had to be duplicated.
- `TableViewModel.swift` and `CollectionViewModel.swift` both received identical `.date` cell value handler changes for the same reason.

### DIFF
--- a/Sources/JoyfillUI/View/Fields/CollectionView/CollectionSearchBar.swift
+++ b/Sources/JoyfillUI/View/Fields/CollectionView/CollectionSearchBar.swift
@@ -43,6 +43,12 @@ struct CollectionSearchBar: View {
                             self.model.filterText = cellDataModel.multiSelectValues?.first ?? ""
                         case .barcode:
                             self.model.filterText = cellDataModel.title ?? ""
+                        case .date:
+                            if let dateEpoch = cellDataModel.date {
+                                self.model.filterText = String(dateEpoch)
+                            } else {
+                                self.model.filterText = ""
+                            }
                         default:
                             break
                         }
@@ -76,6 +82,11 @@ struct CollectionSearchBar: View {
                             .cornerRadius(8)
                             .padding(.leading, 8)
                             .padding(.trailing, 8)
+                    case .date:
+                        TableDateView(cellModel: Binding.constant(cellModel), initialFilterText: model.filterText)
+                            .accessibilityIdentifier("SearchBarDateIdentifier")
+                            .frame(height: 40)
+                            .padding(.horizontal, 4)
                     default:
                         HStack(spacing: 8) {
                             Image(systemName: "exclamationmark.triangle.fill")
--- a/Sources/JoyfillUI/View/Fields/CollectionView/CollectionSearchBar.swift
+++ b/Sources/JoyfillUI/View/Fields/CollectionView/CollectionSearchBar.swift
@@ -47,7 +47,7 @@ struct CollectionSearchBar: View {
                             if let dateEpoch = cellDataModel.date {
                                 self.model.filterText = String(dateEpoch)
                             } else {
-                                self.model.filterText = ""
+                                self.model.filterText = "null"
                             }
                         default:
                             break
@@ -87,6 +87,11 @@ struct CollectionSearchBar: View {
                             .accessibilityIdentifier("SearchBarDateIdentifier")
                             .frame(height: 40)
                             .padding(.horizontal, 4)
+                            .onAppear {
+                                if self.model.filterText.isEmpty {
+                                    self.model.filterText = "null"
+                                }
+                            }
                     default:
                         HStack(spacing: 8) {
                             Image(systemName: "exclamationmark.triangle.fill")
--- a/Sources/JoyfillUI/View/Fields/CollectionView/CollectionSearchBar.swift
+++ b/Sources/JoyfillUI/View/Fields/CollectionView/CollectionSearchBar.swift
@@ -47,7 +47,7 @@ struct CollectionSearchBar: View {
                             if let dateEpoch = cellDataModel.date {
                                 self.model.filterText = String(dateEpoch)
                             } else {
-                                self.model.filterText = "null"
+                                self.model.filterText = FilterModel.emptyDateSentinel
                             }
                         default:
                             break
@@ -89,7 +89,7 @@ struct CollectionSearchBar: View {
                             .padding(.horizontal, 4)
                             .onAppear {
                                 if self.model.filterText.isEmpty {
-                                    self.model.filterText = "null"
+                                    self.model.filterText = FilterModel.emptyDateSentinel
                                 }
                             }
                     default:
--- a/Sources/JoyfillUI/View/Fields/TableView/SearchBar.swift
+++ b/Sources/JoyfillUI/View/Fields/TableView/SearchBar.swift
@@ -125,6 +125,7 @@ struct SearchBar: View {
                 .accessibilityIdentifier("HideFilterSearchBar")
             }
         }
+        .id(model.colID)
         .frame(height: 40)
         .background(Color(.systemGray6))
         .cornerRadius(8)
--- a/Sources/JoyfillUI/View/Fields/TableView/SearchBar.swift
+++ b/Sources/JoyfillUI/View/Fields/TableView/SearchBar.swift
@@ -49,7 +49,7 @@ struct SearchBar: View {
                             if let dateEpoch = cellDataModel.date {
                                 self.model.filterText = String(dateEpoch)
                             } else {
-                                self.model.filterText = "null"
+                                self.model.filterText = FilterModel.emptyDateSentinel
                             }
                         default:
                             break
@@ -89,7 +89,7 @@ struct SearchBar: View {
                             .frame(height: 25)
                             .onAppear {
                                 if self.model.filterText.isEmpty {
-                                    self.model.filterText = "null"
+                                    self.model.filterText = FilterModel.emptyDateSentinel
                                 }
                             }
                     default:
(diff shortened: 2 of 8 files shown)


Sources/JoyfillUI/View/Fields/TableView/SearchBar.swift
```diff
@@ -77,6 +83,10 @@ struct SearchBar: View {
                             .frame(height: 25)
                             .cornerRadius(6)
                             .padding(.leading, 8)
+                    case .date:
+                        TableDateView(cellModel: Binding.constant(cellModel), initialFilterText: model.filterText)
```
vivekjyani: **[Critical] Filter date picker selection never updates `model.filterText`**

`Binding.constant` is read-only. When the user picks a date in the inline filter picker, `TableDateView.onChange(of: selectedDate)` tries to write back:

```swift
cellModel.data = cellDataModel   // silently discarded — constant binding
cellModel.didChange?(cellDataModel)  // also a no-op
```

Neither write reaches `model.filterText`, so the filter never updates after the user interacts with the inline picker. The initial pre-population (from tapping a row) works, but any subsequent date change in the filter bar is ignored.

**Suggested fix**: add an `onDateSelected: ((Double?) -> Void)?` callback to `TableDateView` and call it in `onChange(of: selectedDate)`. The search bar passes a closure that writes the epoch back to `model.filterText`. Alternatively, pass a `Binding<String>` for `filterText` directly.

Sources/JoyfillUI/View/Fields/CollectionView/CollectionSearchBar.swift
```diff
@@ -76,6 +82,11 @@ struct CollectionSearchBar: View {
                             .cornerRadius(8)
                             .padding(.leading, 8)
                             .padding(.trailing, 8)
+                    case .date:
+                        TableDateView(cellModel: Binding.constant(cellModel), initialFilterText: model.filterText)
```
vivekjyani: Same critical issue as `SearchBar.swift`: `Binding.constant(cellModel)` means date picker changes in the filter bar are silently discarded and `model.filterText` is never updated. Needs the same fix (callback or `Binding<String>` for filter text).

VishnuBishnoi (commented):
## Code Review — PR #296: Date filter/sort for Table & Collection

### Overall Assessment
Well-structured PR that brings date columns to parity with other filterable types. The approach of using raw epoch strings as filter keys is sound and avoids locale/timezone issues. The `TableDateView` reuse via `initialFilterText` is a clean design choice.

### Issues Found

#### 1. `"null"` sentinel string is fragile
**Files:** `SearchBar.swift`, `CollectionSearchBar.swift`, `Models.swift`

Using the string `"null"` as a sentinel to represent "no date" is risky. If a filter text ever gets set to `"null"` through some other path (or a future refactor doesn't know about this convention), it will silently match rows with no date. Consider using an empty string `""` as the "no filter" sentinel, or introducing a dedicated enum/optional for filter state.

In `Models.swift`:
```swift
case .date:
    if filter.filterText == "null" {
        match = column.date == nil
```

This means if a user "filters" on date and then clears the filter, they'd see only rows with no date instead of all rows. Is that intended behavior? Typically clearing a filter should show all rows.

#### 2. `getDummyCell` signature change — breaking `empty` parameter without default
**File:** `Models.swift`

`getDummyCell(col:selectedOptionText:empty:)` now requires an `empty` parameter with no default value. All existing callers must be updated. I see `SearchBar.swift` passes `empty: true`, but verify that **all other callers** of `getDummyCell` have been updated. If any callers exist in Collection code, they'd fail to compile.

#### 3. Sort with nil-coalesced `0` can produce wrong ordering
**Files:** `TableModalView.swift`, `CollectionViewModel.swift`

```swift
return (cell1.date ?? 0) < (cell2.date ?? 0)
```

Coalescing nil dates to `0` (epoch 1970-01-01) means rows without a date will sort as if they're from 1970. This pushes them to the top in ascending order and bottom in descending. Consider sorting nil dates consistently to the end regardless of sort direction — this is the standard UX expectation.

#### 4. `.id(model.colID)` on SearchBar
**File:** `SearchBar.swift`

Adding `.id(model.colID)` forces a full view rebuild when the column changes. This is probably intentional for resetting the date picker state, but confirm there are no performance implications with many columns.

#### 5. Minor: `getFormat` → `DateFormatType(rawValue:)` fix
**File:** `Models.swift`

Good catch on the pre-existing bug. The fix looks correct — `DateFormatType(rawValue:)` is the right call here since `column.format` is already the raw string.

### Suggestions
- Add a comment near the `"null"` sentinel explaining the convention, or better yet, extract it into a constant like `private static let noDateFilterSentinel = "null"`.
- Consider adding at least one unit test for the date filter matching logic in `Models.swift` (epoch equality, nil handling).

### Verdict
Approve with minor suggestions. The core logic is correct and well-thought-out. The `"null"` sentinel and nil-coalescing in sort are the main items to address.

VishnuBishnoi (approved):
## Re-review — PR #296 (updated)

The previous review comments have been addressed:

1. **`"null"` sentinel → `FilterModel.emptyDateSentinel`** — Extracted to a named static constant. This eliminates the fragility concern. 

2. **Sort nil-coalescing `0` → `-.infinity`** — Nil dates now coalesce to `-.infinity`, which consistently pushes them to the bottom in ascending and top in descending. This is better than `0` (epoch 1970), though the ideal UX is "nil always at the end regardless of direction." Acceptable tradeoff — can be refined later if users report confusion.

3. **`getDummyCell(col:selectedOptionText:empty:)`** — The `empty` parameter still has no default value. All callers must explicitly pass it. Consider adding `empty: Bool = false` for backward compatibility, but this is minor since it's internal API and the compiler will catch any missed call sites.

### Remaining minor note
The `emptyDateSentinel` filter logic still means "filtering on date with no selection" shows only rows with nil dates. This is a valid UX choice (filter = "show rows matching X", where X = "no date") but worth documenting in a code comment so future maintainers understand the intent.

### Verdict
**Approved.** All substantive issues from the first review have been addressed.

PR #305, opened 2026-04-02
NO-2055: Add row/cell path validation and centralize path parsing

vivekjyani:
## Context

`validate(path:)` is a public API on `DocumentEditor` that lets callers check validity at different granularities — page, field, row, or cell — using a slash-delimited path string (`pageId/fieldPositionId/rowId/columnId`). This PR completes the implementation for the row/cell depth and hardens error handling across all depths.

## What Changed

| File | Change |
|---|---|
| `ValidationHandler.swift` | Moved `validate(path:)` here from `DocumentEditor`; added full row/cell resolution logic |
| `DocumentEditor+Navigation.swift` | Extracted `parsePath(_:)` + `pathSegment(in:at:)` helpers; `goto()` reuses them (removes duplication) |
| `Validator.swift` | Activated `.row` / `.cell` cases (were marked `// future`); added `.notFound` case |
| `ValidatePathTests.swift` | Added 5 new tests for row/cell paths and invalid-ID → `.notFound` cases; updated 2 existing tests to expect `.notFound` instead of `.page` fallback |

## Decisions

- **`.notFound` instead of silent fallback** — Previously, an unresolvable `fieldPositionId` silently returned `.page`. That made it impossible for callers to distinguish "the page is invalid" from "you passed a bad ID". `.notFound` is explicit and doesn't mask bugs in the caller's path construction.
- **`parsePath` centralized in `DocumentEditor+Navigation`** — Both `goto()` and `validate(path:)` need the same segment parsing. Extracting it once avoids drift and makes the intent clear. It's `internal` so tests can reach it without exposing it publicly.
- **`validate(path:)` lives in `ValidationHandler`** — Keeps validation logic out of the view-model layer. `DocumentEditor` delegates through `validationHandler` the same way it already does for other validation calls.

## Screenshots / Demo

No visible UI change — this is a logic-layer and API change only.

## Public API Changes

**Yes — this touches the public API.**

- `ComponentValidity.notFound` is a new case in `Validator.swift`. Any `switch` on `ComponentValidity` in consuming code will need to handle this case (or add a default).
- `ComponentValidity.row` and `.cell` were previously marked `// future` and effectively dead; they are now live.
- Docs for `validate(path:)` should be updated to document the `.notFound` return value and the supported path depths.

## Migration — `ComponentValidity` exhaustive switches

If you have an exhaustive `switch` over `ComponentValidity`, add a `case .notFound:` branch:

```swift
switch validity {
case .page(let v): ...
case .field(let v): ...
case .row(let v): ...
case .cell(let v): ...
case .notFound: // path was invalid or field is hidden
    break
}
```

## Test Coverage

All new behavior is covered:
- `testValidatePath_RowPath_ReturnsDotRowValidity` — row-depth path returns `.row`
- `testValidatePath_CellPath_ReturnsDotCellValidity` — cell-depth path returns `.cell`
- `testValidatePath_InvalidRowId_ReturnsNotFound` — bad rowId → `.notFound`
- `testValidatePath_InvalidColumnId_ReturnsNotFound` — bad columnId → `.notFound`
- `testValidatePath_InvalidFieldPositionId_ReturnsNotFound` — bad fieldPositionId → `.notFound`
- Two existing tests updated to assert `.notFound` (previously asserted `.page` fallback)

## Notes for Reviewer

- The `parsePath` helper is `internal`, not `private`, purely to keep `ValidatePathTests` able to unit-test it directly without going through the full `validate` stack. If that feels wrong, we can make it `private` and test it indirectly.
- `rowValidity` and `cellValidity` are private helpers in `ValidationHandler` — they delegate to `first(where:)` on the validity arrays, which are already computed by the existing validation pass. No extra traversal cost.

### DIFF
--- a/JoyfillSwiftUIExample/JoyfillTests/ValidatePathTests.swift
+++ b/JoyfillSwiftUIExample/JoyfillTests/ValidatePathTests.swift
@@ -328,21 +328,18 @@ final class ValidatePathTests: XCTestCase {
 
     // MARK: - Path page must own the field position
 
-    func testValidatePath_WrongPageIDWithValidFieldPositionID_FallsBackToPage() {
+    func testValidatePath_WrongPageIDWithValidFieldPositionID_ReturnsNotFound() {
         let editor = makeDocumentWithRequiredImageFieldWithoutValue()
         let path = "non-existent-page-id/\(imagePositionID)"
         let result = editor.validate(path: path)
 
-        if case .page(let validation) = result {
-            XCTAssertEqual(validation.status, .valid)
-            XCTAssertTrue(validation.fieldValidities.isEmpty)
-        } else {
-            XCTFail("Expected .page when path page does not own the field position")
+        if case .notFound = result { } else {
+            XCTFail("Expected .notFound when fieldPositionId does not belong to the given pageId, got \(result)")
         }
     }
 
     /// Locks `fieldIdentifier.pageID == pageId` prefix: position exists on page 1 but path uses another real page id → `.page` for that other page, not `.field` for the image on page 1.
-    func testValidatePath_RealOtherPageIDWithPositionFromFirstPage_FallsBackToPage() {
+    func testValidatePath_RealOtherPageIDWithPositionFromFirstPage_ReturnsNotFound() {
         let editor = makeDocumentWithRequiredImageFirstPageAndSecondPage()
 
         let correctPath = "\(pageID)/\(imagePositionID)"
@@ -355,17 +352,9 @@ final class ValidatePathTests: XCTestCase {
         let wrongPath = "\(secondPageID)/\(imagePositionID)"
         let result = editor.validate(path: wrongPath)
 
-        guard case .page(let validation) = result else {
-            XCTFail("Expected .page fallback when path page id does not own the field position, got \(result)"); return
+        if case .notFound = result { } else {
+            XCTFail("Expected .notFound when fieldPositionId does not belong to the given pageId, got \(result)")
         }
-        XCTAssertFalse(
-            validation.fieldValidities.contains { $0.fieldPositionId == imagePositionID },
-            "Page-scoped result must not include the field position from another page"
-        )
-        XCTAssertFalse(
-            validation.fieldValidities.contains { $0.field.id == imageFieldID },
-            "Page-scoped result must not validate page-1 image under wrong page prefix"
-        )
     }
 
     // MARK: - 7. Required field without value → .invalid
--- a/JoyfillSwiftUIExample/JoyfillTests/ValidatePathTests.swift
+++ b/JoyfillSwiftUIExample/JoyfillTests/ValidatePathTests.swift
@@ -921,4 +921,15 @@ final class ValidatePathTests: XCTestCase {
             XCTFail("Expected .notFound for invalid fieldPositionId")
         }
     }
+
+    func testValidatePath_OptionalCellWithValue_ReturnsCellValid() {
+        let editor = makeTableEditorForRowCellPathValidation()
+        let result = editor.validate(path: "page-1/fp-1/row-1/col-optional")
+
+        guard case .cell(let cellValidity) = result else {
+            XCTFail("Expected .cell for optional column with value")
+            return
+        }
+        XCTAssertEqual(cellValidity.status, .valid)
+    }
 }
--- a/JoyfillSwiftUIExample/JoyfillTests/ValidatePathTests.swift
+++ b/JoyfillSwiftUIExample/JoyfillTests/ValidatePathTests.swift
@@ -932,4 +932,28 @@ final class ValidatePathTests: XCTestCase {
         }
         XCTAssertEqual(cellValidity.status, .valid)
     }
+
+    // MARK: - 24. Leading / multiple slash paths
+    // parsePath strips empty segments produced by leading or adjacent slashes,
+    // so these paths resolve identically to their slash-free equivalents.
+
+    // Leading slash → normalised, resolves to .page for that pageID
+    func testValidatePath_LeadingSlash_NormalisedToPage() {
+        let editor = makeDocumentWithRequiredImageFieldWithoutValue()
+        let result = editor.validate(path: "/\(pageID)")
+
+        if case .page = result { } else {
+            XCTFail("Expected .page for path with leading slash — normalisation should strip the empty segment")
+        }
+    }
+
+    // Double leading slash before pageID/fieldPositionID → normalised, resolves to .field
+    func testValidatePath_DoubleLeadingSlash_NormalisedToField() {
+        let editor = makeDocumentWithRequiredImageFieldWithoutValue()
+        let result = editor.validate(path: "//\(pageID)/\(imagePositionID)")
+
+        if case .field = result { } else {
+            XCTFail("Expected .field for path with double leading slash — normalisation should strip empty segments")
+        }
+    }
 }
--- a/JoyfillSwiftUIExample/JoyfillTests/ValidatePathTests.swift
+++ b/JoyfillSwiftUIExample/JoyfillTests/ValidatePathTests.swift
@@ -302,32 +302,27 @@ final class ValidatePathTests: XCTestCase {
         }
     }
 
-    // MARK: - 5. Non-existent fieldPositionId → falls back to .page
+    // MARK: - 5. Non-existent fieldPositionId → .notFound
 
-    func testValidatePath_NonExistentFieldPositionID_FallsBackToPage() {
+    func testValidatePath_NonExistentFieldPositionID_ReturnsNotFound() {
         let editor = makeDocumentWithRequiredImageFieldWithoutValue()
         let path = "\(pageID)/non-existent-position-id"
         let result = editor.validate(path: path)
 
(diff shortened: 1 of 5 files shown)


Sources/JoyfillUI/ViewModels/ValidationHandler.swift
```diff
@@ -66,6 +66,59 @@ class ValidationHandler {
         }
         return Validation(status: isValid ? .valid : .invalid, fieldValidities: fieldValidities)
     }
+    
+    func validate(path: String) -> ComponentValidity {
+        guard let documentEditor = documentEditor else {
+            return .page(Validation(status: .valid, fieldValidities: []))
```
vivekjyani: **Nil `documentEditor` silently returns `.page(.valid)` — could mask errors.**

When the weak `documentEditor` reference is nil (already deallocated), returning a "valid" page result is misleading — a caller cannot distinguish between "the document passed validation" and "the handler has been released". This is particularly surprising since every other failure path below this one returns `.notFound`. Consider returning `.notFound` here, or at least adding a `Log` call before returning so this deallocation is surfaced during debugging.

Sources/JoyfillUI/ViewModels/ValidationHandler.swift
```diff
@@ -66,6 +66,59 @@ class ValidationHandler {
...
+        }
+
+        guard let fieldIdentifier = documentEditor.getFieldIdentifier(forFieldPositionID: fieldPositionID) else {
+            Log("Field position with id \(fieldPositionID) not found", type: .error)
+            return .notFound
+        }
+
+        guard fieldIdentifier.pageID == pageID,
+              let fieldValidity = validate(fieldIdentifier: fieldIdentifier) else {
+            return .notFound
```
vivekjyani: **Combined `guard` merges two distinct failure modes without a log.**

This `guard` collapses two different failure reasons into a single `.notFound`:
1. `fieldIdentifier.pageID != pageID` — the caller passed a `fieldPositionId` that belongs to a *different* page (caller error)
2. `validate(fieldIdentifier:) == nil` — the field is hidden or filtered by conditional logic (expected runtime state)

Both are currently silent. Consider splitting into two guards with individual `Log(...)` calls so callers debugging a `.notFound` can tell which condition triggered.

Sources/JoyfillModel/Validator.swift
```diff
@@ -69,8 +69,9 @@ public struct FieldValidity {
...
     case page(Validation)
     case field(FieldValidity)
-    case row(RowValidity)      // future
-    case cell(CellValidity)    // future
+    case row(RowValidity)
+    case cell(CellValidity)
+    case notFound
 
     public var fieldValidity: FieldValidity? {
         if case .field(let fv) = self { return fv }
```
vivekjyani: **`fieldValidity` accessor doesn't cover the newly promoted `.row` and `.cell` cases.**

Now that `.row` and `.cell` are production cases, the existing `fieldValidity` convenience accessor only handles `.field`. Callers receiving `.row` or `.cell` must fall back to explicit pattern matching. Consider adding symmetric `rowValidity: RowValidity?` and `cellValidity: CellValidity?` computed properties, or add a doc comment noting that these cases require manual pattern matching.

JoyfillSwiftUIExample/JoyfillTests/ValidatePathTests.swift
```diff
@@ -118,6 +118,43 @@ final class ValidatePathTests: XCTestCase {
         return documentEditor(document: document)
     }
 
+    func makeTableEditorForRowCellPathValidation() -> DocumentEditor {
```
vivekjyani: **This helper doesn't call `setMobileView()` or `setPageFieldInMobileView()` — different from every other helper in this file.**

All other `makeDocument*` helpers configure a mobile view before building the document. This one uses a raw `JoyDoc(dictionary:)` without it, which means `mapWebViewToMobileViewIfNeeded` will take a different branch during validation. If this is intentional (testing the web-view code path), add a comment explaining why; otherwise consider aligning with the other helpers to avoid unintentional coverage gaps.

JoyfillSwiftUIExample/JoyfillTests/ValidatePathTests.swift
```diff
@@ -847,4 +868,57 @@ final class ValidatePathTests: XCTestCase {
...
+        let editor = makeTableEditorForRowCellPathValidation()
+        let result = editor.validate(path: "page-1/fp-1/row-1/col-required")
+
+        guard case .cell(let cellValidity) = result else {
+            XCTFail("Expected .cell for cell path")
+            return
+        }
+        XCTAssertEqual(cellValidity.columnId, "col-required")
+        XCTAssertEqual(cellValidity.status, .invalid)
+    }
```
vivekjyani: **Missing test for the `.cell(.valid)` happy path.**

The fixture already has `col-optional` with value `"ok"` — a non-required, non-empty cell that should produce `.cell(.valid)`. Consider adding:
```swift
func testValidatePath_OptionalCellWithValue_ReturnsCellValid() {
    let editor = makeTableEditorForRowCellPathValidation()
    let result = editor.validate(path: "page-1/fp-1/row-1/col-optional")
    guard case .cell(let cellValidity) = result else {
        XCTFail("Expected .cell"); return
    }
    XCTAssertEqual(cellValidity.status, .valid)
}
```
This would confirm the valid-cell path alongside the existing `.invalid` coverage.

vivekjyani (commented):
**Review: NO-2055 — Row/Cell Path Validation**

Solid PR overall — the refactor to centralize path parsing is clean, and the new `.notFound` case is a meaningful improvement over silent fallbacks. A few issues worth addressing before merge.

Sources/JoyfillUI/ViewModels/ValidationHandler.swift
vivekjyani: yes, both needs to return .notFound

VishnuBishnoi (commented):
Solid implementation — the hierarchical validate-path API is clean and well-tested. Several issues to address before merging:

**Breaking API change (`.notFound` in `ComponentValidity`):** any exhaustive `switch` on this enum will fail to compile. The PR description mentions this, but no migration snippet is provided. Please add one to the PR description or a CHANGELOG entry, e.g.:
```swift
// Add to existing switch statements:
case .notFound:
    // handle path not found
```

**Missing happy-path test for required cell with value:** `testValidatePath_OptionalCellWithValue_ReturnsCellValid` only validates the optional column. There is no test verifying that a required column with a non-empty value returns `.cell(.valid)`. Please add a case where `col-required` has a non-empty value.

See inline comments for additional issues.

Sources/JoyfillUI/ViewModels/DocumentEditor+Navigation.swift
```diff
@@ -263,3 +262,26 @@ extension DocumentEditor {
         self.onBlur(event: pageEvent)
     }
 }
+
+extension DocumentEditor {
+    func parsePath(_ path: String) -> (
+        pageId: String?,
```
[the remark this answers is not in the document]
vivekjyani: Good call — parsePath is pure string manipulation with no dependency on self. Made it static on DocumentEditor; ValidationHandler now calls DocumentEditor.parsePath(path) directly without needing the documentEditor reference just for path parsing.

Sources/JoyfillUI/ViewModels/ValidationHandler.swift
```diff
@@ -66,6 +66,65 @@ class ValidationHandler {
...
+
+        guard let cellValidity = cellValidity(for: columnId, in: rowValidity) else {
+            Log("Column with id \(columnId) not found in row \(rowId)", type: .error)
+            return .notFound
+        }
+
+        return .cell(cellValidity)
+    }
+
+    private func rowValidity(for rowId: String, in fieldValidity: FieldValidity) -> RowValidity? {
```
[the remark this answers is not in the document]
VishnuBishnoi: That comment is exactly what was needed. Thread resolved, thanks.

VishnuBishnoi (approved):
All outstanding items have been addressed. **Approving.**

### Resolved
- **Log level inconsistency** — both caller-side path errors now log `.warning`. 
- **`.notFound` for hidden field** — doc comment added to `validate(path:)` making it explicit that `.notFound` covers both invalid paths and conditionally-hidden fields. The intentional design decision is now self-documenting.
- **`rowValidity` invariant** — comment added documenting that deleted rows are excluded upstream.
- **`parsePath` as instance method** — made `static`, `ValidationHandler` calls `DocumentEditor.parsePath(_:)` directly. Cleaner coupling.
- **Leading-slash normalisation** — doc comment added and two tests committed (`testValidatePath_LeadingSlash_NormalisedToPage`, `testValidatePath_DoubleLeadingSlash_NormalisedToField`). Documented contract is now test-covered.
- **Breaking change migration guide** — added to PR description.
- **`fieldValidity` accessor gap** — skipping `rowValidity`/`cellValidity` helpers is a reasonable call given the switch-based API already handles it cleanly.

### One thing to verify before merge
The `.cell(.valid)` happy-path test was confirmed added — just make sure it appears in the final commit (`testValidatePath_OptionalCellWithValue_ReturnsCellValid` or equivalent for the required column with a value).

PR #308, opened 2026-04-10
Add dynamic decorator API with live table/collection updates

vivekjyani:
## Context

Implements the dynamic decorator public API per the decorator spec. Decorators are visual badges (icon + label + color) attached to fields, rows, or columns. This PR adds path-based CRUD methods so integrators can read, add, update, and remove decorators at runtime — and the UI reflects changes live without re-opening the table/collection modal.

On top of that baseline, the PR also introduces (a) error reporting via `onError`/`JoyfillError.decoratorError`, (b) decorator validation with block-and-return-early semantics, (c) license gating for collection-field writes, and (d) a full unit-test suite (70 tests).

### New file: `Sources/JoyfillUI/ViewModels/DocumentEditor+Decorators.swift`
- Path-based public API: `getDecorators(path:)`, `addDecorators(path:)`, `removeDecorator(path:action:)`, `updateDecorator(path:action:decorator:)`
- Path format: `pageId/fieldPositionId` (field), `/rowId` (row), `/rowId/columnId` (column)
- `resolveDecoratorTarget` — resolves a path string to the correct field, row, or column scope
- `resolvedSchemaKey(forRowID:inFieldID:)` — recursive schema key resolution for collection fields (root vs nested rows)
- `validateDecorator(_:)` — enforces `action` required/non-empty and `color` matches `^#[0-9A-Fa-f]{6}$` when non-nil/non-empty
- `licenseAllowsDecoratorWrite(for:path:)` — blocks collection-field writes when `isCollectionFieldEnabled == false`
- Separate read/write helpers for field, row, and column decorators with full table/collection branching

### Modified: `Sources/JoyfillUI/ViewModels/DocumentEditor.swift`
- Added `decoratorsDidChange()` to `DocumentEditorDelegate` protocol

### Modified: `Sources/JoyfillUI/View/Fields/TableView/TableViewModel.swift`
- `decoratorsDidChange()` — refreshes `tableDataModel.rowDecorators` and column decorator arrays from the live field model

### Modified: `Sources/JoyfillUI/View/Fields/CollectionView/CollectionViewModel.swift`
- `decoratorsDidChange()` — replaces `tableDataModel.schema` from the fresh field, refreshes `rowDecoratorsBySchemaKey` cache, and patches root-level `tableColumns`

### Modified: `Sources/JoyfillUI/View/Fields/Models.swift`
- Changed `rowDecoratorsBySchemaKey` from `private` to `var` so `CollectionViewModel` can update it

### Modified: `Sources/JoyfillUI/View/Types.swift`
- Added `JoyfillError.decoratorError(error: DecoratorError)` case
- Added public `DecoratorError` struct with `message: String`

### Modified: `JoyfillSwiftUIExample/JoyfillExample/ChangeManager.swift`
- Exhaustive switch update for the new error case (prints the decorator error message)

### Example project: `DecoratorAPIDemoView.swift`
- `DecoratorManagerView` — full CRUD UI for managing decorators across all field types (field/row/column), with schema picker for collections
- `DecoratorEventHandler` class — handles `onFocus` events from decorator taps and calls `updateDecorator` to visually mark decorators as "Viewed"
- Demonstrates the full integration pattern for consuming the decorator API

## Public API

```swift
public extension DocumentEditor {
    func getDecorators(path: String) -> [Decorator]
    func addDecorators(path: String, decorators: [Decorator])
    func removeDecorator(path: String, action: String)
    func updateDecorator(path: String, action: String, decorator: Decorator)
}

public enum JoyfillError: Error {
    case schemaValidationError(error: SchemaValidationError)
    case schemaVersionError(error: SchemaValidationError)
    case decoratorError(error: DecoratorError)
}

public struct DecoratorError: Error {
    public let message: String
}
```

New method on `DocumentEditorDelegate`:
```swift
func decoratorsDidChange()
```

> **Docs update needed** — The public API docs should be updated to document the path format, the new decorator methods, the error enum case, and license requirements.

## Error reporting

`events?.onError(error: .decoratorError(...))` fires in four cases:

1. **Path cannot be resolved** — bad `pageId` / `fieldPositionId` / `rowId` / `columnId`. Message: `"Failed to resolve path '<path>'"`. Applies to all four public methods. `getDecorators` additionally returns `[]`.
2. **Action not found** on `removeDecorator` or `updateDecorator`. Message: `"Failed to remove/update decorator with action: '<action>'"`.
3. **Decorator validation fails** — nil/empty `action`, or malformed `color` hex (non-`#`, wrong length, non-hex chars). Message: `"Invalid decorator property: ..."`.
4. **License block** — write to a collection-field decorator with `isCollectionFieldEnabled == false`. Message: `"Collection field decorators are not available without a valid license (path '<path>')"`.

**Block-and-return-early** semantics: on validation failure the write is rejected, nothing persists, no delegate callback fires. For `addDecorators`, batches are atomic — if any decorator in the batch is invalid, the entire batch is rejected.

## License gating

Writes to decorators on a `.collection` field are now blocked when the license does not grant collection access (`LicenseValidator.isCollectionEnabled(licenseToken:) == false`). This matches how the rest of the SDK treats unlicensed collection fields (silently skipped in view-model building at `DocumentEditor.swift:814`).

| Method | Behavior on unlicensed collection field |
|---|---|
| `getDecorators` | allowed (reads are not gated) |
| `addDecorators` | blocked, `decoratorError` emitted |
| `removeDecorator` | blocked, `decoratorError` emitted |
| `updateDecorator` | blocked, `decoratorError` emitted |

Table-field decorators are unaffected.

## Design Decisions

1. **Path-based API over direct field access** — Paths (`pageId/fieldPositionId/rowId/columnId`) give integrators a single entry point that works for all decorator scopes without needing to know internal field IDs or schema keys. The SDK resolves the path internally.

2. **Page-scoped field position lookup** — `resolveDecoratorTarget` looks up the field position on the specific page from the path (not `getFieldIdentifier(forFieldPositionID:)` which iterates all pages). This is required because Navigation.json has 22 shared field position IDs across pages that point to different fields.

3. **Delegate pattern for live updates** — Table/collection views cache decorator data in `TableDataModel` at init time. Rather than forcing a full view rebuild, `decoratorsDidChange()` patches the existing `@StateObject` view model's cached decorators. This matches the existing `change(changes:)` delegate pattern.

4. **`valueDelegate(for:fieldType:)` for lazy creation** — If no view model is registered yet (modal not open), a throwaway instance is created. The decorator data is already persisted in `fieldMap`, so when the modal eventually opens it initializes fresh. Same pattern as the change API.

5. **`tableDataModel.schema = field.schema ?? [:]`** — For collections, we replace the entire schema instead of patching individual properties. This keeps `hasAnyRowDecorators` (which gates row decorator visibility by checking `schema.values`) in sync with the `rowDecoratorsBySchemaKey` render cache.

6. **`DecoratorEventHandler` as a class, not struct** — SwiftUI struct views are recreated on every render, but `@StateObject` only initializes once. A class reference stored via `DocumentEditor(events:)` at init time maintains a stable connection. Accessed via `editor.events as? DecoratorEventHandler`.

7. **License gating mirrors the view-model build-time gate** — `DocumentEditor.swift:814` already skips collection fields when the license is invalid (they never get a view model). Mirroring that at the decorator write layer keeps behavior consistent: unlicensed integrators can't accidentally mutate collection-field state that will never render.

8. **Reads are not license-gated** — inspecting existing decorator data is harmless and useful for debugging or UI state derivation. Only mutating APIs (`add`/`remove`/`update`) enforce the license.

## Test Coverage

New unit-test suite under `JoyfillSwiftUIExample/JoyfillTests/DecoratorAPI/`:

| File | Tests |
|---|---|
| `DecoratorTestSupport.swift` | shared helpers (`MockDecoratorEvents`, sample IDs, editor factories) |
| `DecoratorPathResolutionTests.swift` | 15 |
| `DecoratorPublicAPITests.swift` | 16 |
| `DecoratorErrorHandlingTests.swift` | 28 |
| `DecoratorLiveUpdateTests.swift` | 11 |
| **Total** | **70** |

Coverage areas:
- **Path resolution** — 2/3/4-segment paths, schema-key resolution for root and nested rows, the shared-`fieldPositionId` regression from the original bug, invalid page/fieldPosition/field IDs, empty path, extra slashes.
- **CRUD happy paths** — `get`/`add`/`remove`/`update` on field, row, column scopes for both table and collection fields; round-trip reads; consecutive operations; `fieldMap` persistence.
- **Error emission + validation** — every public method's path-resolution error; unknown-action errors on `remove`/`update`; validation for nil/empty action, malformed/short/invalid/valid hex colors; atomic batch rejection; error-message shape; `DecoratorError` pattern-matching; `events: nil` crash safety.
- **License gating** — collection add/remove/update blocked + `onError` fired; collection get allowed; table writes unaffected; batch writes blocked wholesale.
- **Live updates** — `TableViewModel` row/column cache refresh; `CollectionViewModel` root/nested row and column refresh; `hasAnyRowDecorators` regression; lazy delegate creation; cross-field isolation.

Collection tests require the `JOYFILL_TEST_LICENSE` env var (same pattern as the existing `ValidationTestCase.collectionDocumentEditor`). Without it, the license gate correctly blocks the writes and those tests will fail by design.

### Modified: `JoyfillSwiftUIExample/JoyfillExample.xcodeproj/project.pbxproj`
- New `DecoratorAPI` group under the `JoyfillTests` target, wiring all 5 test files into the Sources build phase.

## Notes for Reviewers

- The example project files (`DecoratorAPIDemoView.swift`) are intentionally verbose — they serve as integration documentation for the API.
- `refreshField` is still called alongside `decoratorsDidChange()`. It rebuilds the page field model (needed for field-level decorators), while `decoratorsDidChange()` patches the live table/collection view model. Both are needed.
- The recursive `schemaKey(forRowID:in:)` walks the `childrens` tree to find which schema a nested row belongs to. This is O(rows) per call but row counts per field are small in practice.
- Validation uses a simple regex (`^#[0-9A-Fa-f]{6}$`). 3-char shorthand (`#abc`) and 8-char (`#rrggbbaa`) are intentionally rejected — Decorator colors are stored as full 6-char hex strings downstream.

### DIFF
--- a/JoyfillSwiftUIExample/JoyfillExample/Decorator Example/DecoratorAPIDemoView.swift
+++ b/JoyfillSwiftUIExample/JoyfillExample/Decorator Example/DecoratorAPIDemoView.swift
@@ -241,7 +241,11 @@ struct DecoratorManagerView: View {
             else { return nil }
             return "\(base)/\(rowID)/\(columnID)"
         }
-        return "\(base)/-/\(columnID)"
+        // Table fields: use a real rowId from rowOrder so the path resolver
+        // can validate it. Column decorators are row-independent for tables,
+        // but the 4-segment path format still requires a valid rowId.
+        guard let rowID = selectedField?.rowOrder?.first else { return nil }
+        return "\(base)/\(rowID)/\(columnID)"
     }
 
     /// Returns the first rowId that belongs to the given schemaKey in a field's value tree.
--- a/JoyfillSwiftUIExample/JoyfillExample/Decorator Example/DecoratorAPIDemoView.swift
+++ b/JoyfillSwiftUIExample/JoyfillExample/Decorator Example/DecoratorAPIDemoView.swift
@@ -1,0 +1,854 @@
+//
+//  DecoratorAPIDemoView.swift
+//  JoyfillExample
+//
+//  Contains:
+//    • DecoratorManagerView   — sheet listing all fields with their decorators; add / edit / delete
+//    • DecoratorEditView      — add-or-edit sheet with icon grid, colour swatches and live preview
+//    • DecoratorAPIDemoView   — standalone demo screen (entry in the option list)
+//
+//  DecoratorManagerView is also used directly by FormDestinationView so the same
+//  UI works whether launched from the menu or from the changelogs toolbar.
+//
+
+import SwiftUI
+import Joyfill
+import JoyfillModel
+
+// MARK: - Icon catalogue (matches DecoratorIcon mapping in the SDK)
+
+private struct IconOption: Identifiable, Hashable {
+    let id: String       // value stored in Decorator.icon
+    let symbol: String   // SF Symbol name
+}
+
+private let iconCatalogue: [IconOption] = [
+    .init(id: "camera",      symbol: "camera.fill"),
+    .init(id: "upload",      symbol: "arrow.up.square.fill"),
+    .init(id: "download",    symbol: "arrow.down.square.fill"),
+    .init(id: "image",       symbol: "photo.fill"),
+    .init(id: "file",        symbol: "doc.fill"),
+    .init(id: "comment",     symbol: "message.fill"),
+    .init(id: "comments",    symbol: "bubble.left.and.bubble.right.fill"),
+    .init(id: "flag",        symbol: "flag.fill"),
+    .init(id: "share",       symbol: "square.and.arrow.up.fill"),
+    .init(id: "eye",         symbol: "eye.fill"),
+    .init(id: "print",       symbol: "printer.fill"),
+    .init(id: "folder",      symbol: "folder.fill"),
+    .init(id: "paperclip",   symbol: "paperclip"),
+    .init(id: "plus",        symbol: "plus.circle.fill"),
+    .init(id: "cloud",       symbol: "cloud.fill"),
+    .init(id: "circle-info", symbol: "info.circle.fill"),
+    .init(id: "filter",      symbol: "line.3.horizontal.decrease.circle.fill"),
+    .init(id: "paper-plane", symbol: "paperplane.fill"),
+]
+
+// MARK: - Colour presets
+
+private struct ColourPreset: Identifiable, Hashable {
+    let id: String   // hex used in Decorator.color
+    let color: Color
+}
+
+private let colourPresets: [ColourPreset] = [
+    .init(id: "#3B82F6", color: Color(red: 0.232, green: 0.510, blue: 0.965)),
+    .init(id: "#10B981", color: Color(red: 0.063, green: 0.725, blue: 0.506)),
+    .init(id: "#EF4444", color: Color(red: 0.937, green: 0.267, blue: 0.267)),
+    .init(id: "#8B5CF6", color: Color(red: 0.545, green: 0.361, blue: 0.965)),
+    .init(id: "#F97316", color: Color(red: 0.976, green: 0.451, blue: 0.086)),
+    .init(id: "#14B8A6", color: Color(red: 0.078, green: 0.722, blue: 0.651)),
+]
+
+// MARK: - Helpers
+
+private func resolvedColor(hex: String?) -> Color {
+    colourPresets.first { $0.id.lowercased() == hex?.lowercased() }?.color
+        ?? colourPresets[0].color
+}
+
+private func resolvedSymbol(iconKey: String?) -> String {
+    iconCatalogue.first { $0.id == iconKey }?.symbol ?? "questionmark.circle"
+}
+
+private func fieldTypeMeta(_ type: String?) -> (label: String, color: Color) {
+    switch type {
+    case "text":       return ("Text",       .blue)
+    case "number":     return ("Number",     .green)
+    case "date":       return ("Date",       .orange)
+    case "dropdown":   return ("Dropdown",   .purple)
+    case "checkbox":   return ("Checkbox",   Color(red: 0.2, green: 0.7, blue: 0.6))
+    case "table":      return ("Table",      .red)
+    case "collection": return ("Collection", .red)
+    case "image":      return ("Image",      .indigo)
+    case "signature":  return ("Signature",  .pink)
+    case "chart":      return ("Chart",      .cyan)
+    default:
+        let t = type ?? ""
+        return (t.isEmpty ? "Field" : t.capitalized, .secondary)
+    }
+}
+
+// MARK: - Draft model (Identifiable so .sheet(item:) works)
+
+struct DecoratorDraft: Identifiable {
+    let id:         UUID   = UUID()
+    let path:       String  // full decorator path (field / row / column level)
+    let editAction: String? // nil → adding new; non-nil → editing existing (matched by action)
+    var icon:       String
+    var label:      String
+    var color:      String
+    var action:     String
+}
+
+// MARK: - DecoratorManagerView
+
+/// Full-screen decorator manager sheet.
+/// Pass the live `DocumentEditor` — changes reflect immediately in the form.
+struct DecoratorManagerView: View {
+    @ObservedObject var editor: DocumentEditor
+    @Environment(\.dismiss) private var dismiss
+
+    @State private var draft: DecoratorDraft?         = nil
+    @State private var selectedPageID:          String = ""
+    @State private var selectedFieldPositionID: String = ""
+    @State private var selectedSchemaKey:       String = ""  // collection fields only
+    @State private var selectedColumnID:        String = ""
+
+    // MARK: Derived — pages
+
+    private var sortedPages: [Page] {
+        editor.pagesForCurrentView.filter { $0.id != nil }
+    }
+
+    private var selectedPage: Page? {
+        sortedPages.first { $0.id == selectedPageID }
+    }
+
+    // MARK: Derived — field entries (loaded explicitly from the selected page)
+
+    /// (fieldPositionId, field) pairs for the selected page, in layout order.
+    private var fieldEntries: [(fieldPositionId: String, field: JoyDocField)] {
+        fieldEntriesForPage(selectedPageID)
+    }
+
+    /// Local helper: builds (fieldPositionId, field) pairs for any page using only
+    /// public DocumentEditor APIs — no SDK-side method needed.
+    private func fieldEntriesForPage(_ pageID: String) -> [(fieldPositionId: String, field: JoyDocField)] {
+        guard !pageID.isEmpty,
+              let page = editor.pagesForCurrentView.first(where: { $0.id == pageID })
+        else { return [] }
+        return (page.fieldPositions ?? []).compactMap { pos in
+            guard let posID   = pos.id,
+                  let fieldID = pos.field,
+                  let field   = editor.field(fieldID: fieldID) else { return nil }
+            return (posID, field)
+        }
+    }
+
+    private var selectedField: JoyDocField? {
+        fieldEntries.first { $0.fieldPositionId == selectedFieldPositionID }?.field
+    }
+
+    private var isCollection: Bool { selectedField?.fieldType == .collection }
+    private var isTable:      Bool { selectedField?.fieldType == .table }
+
+    // MARK: Derived — schemas (collection fields only)
+
+    /// All schemas for the selected collection field, root first then children in declared order.
+    private var sortedSchemas: [(key: String, schema: Schema)] {
+        guard let schemas = selectedField?.schema else { return [] }
+        var result: [(String, Schema)] = []
+        if let rootEntry = schemas.first(where: { $0.value.root == true }) {
+            result.append((rootEntry.key, rootEntry.value))
+            for childKey in rootEntry.value.children ?? [] {
+                if let child = schemas[childKey] { result.append((childKey, child)) }
+            }
+        }
+        // Append any schemas not reachable from root's children list
+        for (key, schema) in schemas where !result.contains(where: { $0.0 == key }) {
+            result.append((key, schema))
+        }
+        return result
+    }
+
+    private var selectedSchema: Schema? {
+        selectedField?.schema?[selectedSchemaKey]
+    }
+
+    // MARK: Derived — columns
+
+    /// Columns for the currently active scope:
+    /// - Collection: columns from the selected schema entry
+    /// - Table:      columns directly on the field
+    private var sortedColumns: [FieldTableColumn] {
+        let cols: [FieldTableColumn]?
+        if isCollection {
+            cols = selectedSchema?.tableColumns
+        } else {
+            cols = selectedField?.tableColumns
+        }
+        return (cols ?? []).filter { $0.id != nil }
+    }
+
+    private var selectedColumn: FieldTableColumn? {
+        sortedColumns.first { $0.id == selectedColumnID }
+    }
+
+    // MARK: Path helpers
+
+    /// Resolves the pageID that the SDK will actually use for this fieldPositionId.
+    /// The Navigation JSON (and similar documents) reuse the same fieldPosition _id across
+    /// multiple pages (copied page templates). getFieldIdentifier always returns the first
+    /// matching page, so we use its pageID in every path to stay consistent with the SDK.
+    private var resolvedPageID: String? {
+        guard !selectedFieldPositionID.isEmpty else { return nil }
+        return editor.getFieldIdentifier(forFieldPositionID: selectedFieldPositionID)?.pageID
+    }
+
+    /// "pageId/fieldPositionId"
+    private var fieldPath: String? {
+        guard let pageID = resolvedPageID else { return nil }
+        return "\(pageID)/\(selectedFieldPositionID)"
+    }
+
+    /// "pageId/fieldPositionId/rowId"
+    /// For collections: uses the first row that belongs to the selected schema so the
+    /// path resolver can derive the correct schemaKey.
+    /// For tables: uses the first root row (schema resolution short-circuits to nil anyway).
+    private var rowPath: String? {
+        guard let base = fieldPath else { return nil }
+        let rowID: String?
+        if isCollection {
+            rowID = firstRowID(forSchemaKey: selectedSchemaKey, in: selectedField)
+        } else {
+            rowID = selectedField?.valueToValueElements?.first?.id
+        }
+        guard let rowID = rowID else { return nil }
+        return "\(base)/\(rowID)"
+    }
+
+    /// "pageId/fieldPositionId/rowId/columnId"
+    /// For collections: uses a real rowId from the selected schema so the resolver
+    /// returns the correct schemaKey for column lookup.
+    /// For tables: uses "-" as a positional placeholder (rowId is ignored for table columns).
+    private func columnPath(columnID: String) -> String? {
+        guard let base = fieldPath else { return nil }
+        if isCollection {
+            guard let rowID = firstRowID(forSchemaKey: selectedSchemaKey, in: selectedField)
+            else { return nil }
+            return "\(base)/\(rowID)/\(columnID)"
+        }
+        return "\(base)/-/\(columnID)"
+    }
+
+    /// Returns the first rowId that belongs to the given schemaKey in a field's value tree.
+    /// - Root schema: first root-level row.
+    /// - Child schema: first child row found under any parent that has that schema.
+    private func firstRowID(forSchemaKey schemaKey: String, in field: JoyDocField?) -> String? {
+        guard let field = field else { return nil }
+        let rootSchemaKey = field.schema?.first { $0.value.root == true }?.key
+        if schemaKey == rootSchemaKey {
+            return field.valueToValueElements?.first?.id
+        }
+        return findFirstChildRow(forSchemaKey: schemaKey, in: field.valueToValueElements ?? [])
+    }
+
+    private func findFirstChildRow(forSchemaKey targetKey: String, in rows: [ValueElement]) -> String? {
+        for row in rows {
+            if let children = row.childrens?[targetKey],
+               let firstID  = children.valueToValueElements?.first?.id {
+                return firstID
+            }
+            if let childrens = row.childrens {
+                for (_, childGroup) in childrens {
+                    if let found = findFirstChildRow(forSchemaKey: targetKey, in: childGroup.valueToValueElements ?? []) {
+                        return found
+                    }
+                }
+            }
+        }
+        return nil
+    }
+
+    // MARK: Body
+
+    var body: some View {
+        NavigationView {
+            Group {
+                if sortedPages.isEmpty {
+                    VStack(spacing: 12) {
+                        Image(systemName: "doc.text.magnifyingglass")
+                            .font(.system(size: 40))
+                            .foregroundColor(.secondary)
+                        Text("No pages in this document.")
+                            .foregroundColor(.secondary)
+                    }
+                    .frame(maxWidth: .infinity, maxHeight: .infinity)
+                } else {
+                    List {
+                        // ── Page picker ──────────────────────────────────────
+                        Section {
+                            pagePickerRow
+                        } header: { Text("Select Page").textCase(nil) }
+
+                        // ── Field picker ─────────────────────────────────────
+                        Section {
+                            fieldPickerRow
+                        } header: { Text("Select Field").textCase(nil) }
+
+                        if let path = fieldPath {
+
+                            // ── Field-level decorators ───────────────────────
+                            let fieldDecs = editor.getDecorators(path: path)
+                            Section {
+                                ForEach(fieldDecs, id: \.action) { decoratorRow($0, path: path) }
+                                if fieldDecs.isEmpty { emptyHint("No decorators — tap + to add one") }
+                                addButton(badge: .blue) {
+                                    draft = DecoratorDraft(path: path, editAction: nil,
+                                                          icon: "camera", label: "", color: "#3B82F6", action: "")
+                                }
+                            } header: {
+                                decoratorSectionHeader(title: "Field Decorators", symbol: "tag.fill",
+                                                       count: fieldDecs.count, badge: .blue)
+                            }
+
+                            // ── Table / Collection only ──────────────────────
+                            if isTable || isCollection {
+
+                                // Schema picker (collection only)
+                                if isCollection {
+                                    Section {
+                                        schemaPickerRow
+                                    } header: { Text("Select Schema").textCase(nil) }
+                                }
+
+                                // Row decorators
+                                if let rPath = rowPath {
+                                    let rowDecs = editor.getDecorators(path: rPath)
+                                    Section {
+                                        ForEach(rowDecs, id: \.action) { decoratorRow($0, path: rPath) }
+                                        if rowDecs.isEmpty { emptyHint("No row decorators — tap + to add one") }
+                                        addButton(badge: .orange) {
+                                            draft = DecoratorDraft(path: rPath, editAction: nil,
+                                                                   icon: "flag", label: "", color: "#F97316", action: "")
+                                        }
+                                    } header: {
+                                        decoratorSectionHeader(title: "Row Decorators", symbol: "list.bullet.rectangle",
+                                                               count: rowDecs.count, badge: .orange)
+                                    }
+                                } else {
+                                    Section {
+                                        emptyHint("Add at least one row to manage row decorators.")
+                                    } header: {
+                                        decoratorSectionHeader(title: "Row Decorators", symbol: "list.bullet.rectangle",
+                                                               count: 0, badge: .orange)
+                                    }
+                                }
+
+                                // Column picker + column decorators
+                                if !sortedColumns.isEmpty {
+                                    Section {
+                                        columnPickerRow
+                                    } header: { Text("Select Column").textCase(nil) }
+
+                                    if let col = selectedColumn, let colID = col.id,
+                                       let cPath = columnPath(columnID: colID) {
+                                        let colDecs = editor.getDecorators(path: cPath)
+                                        Section {
+                                            ForEach(colDecs, id: \.action) { decoratorRow($0, path: cPath) }
+                                            if colDecs.isEmpty { emptyHint("No column decorators — tap + to add one") }
+                                            addButton(badge: .purple) {
+                                                draft = DecoratorDraft(path: cPath, editAction: nil,
+                                                                       icon: "circle-info", label: "", color: "#8B5CF6", action: "")
+                                            }
+                                        } header: {
+                                            decoratorSectionHeader(title: "Column Decorators", symbol: "tablecells",
+                                                                   count: colDecs.count, badge: .purple)
+                                        }
+                                    }
+                                }
+                            }
+                        }
+                    }
+                    .listStyle(.insetGrouped)
+                    .onAppear {
+                        if selectedPageID.isEmpty {
+                            selectedPageID = sortedPages.first?.id ?? ""
+                        }
+                        if selectedFieldPositionID.isEmpty {
+                            selectedFieldPositionID = fieldEntriesForPage(selectedPageID).first?.fieldPositionId ?? ""
+                        }
+                        if selectedSchemaKey.isEmpty { selectedSchemaKey = sortedSchemas.first?.key ?? "" }
+                        if selectedColumnID.isEmpty  { selectedColumnID  = sortedColumns.first?.id  ?? "" }
+                    }
+                    .onChange(of: selectedPageID) { newPageID in
+                        // Page changed: reload field list for the new page, reset downstream
+                        let entries = fieldEntriesForPage(newPageID)
+                        selectedFieldPositionID = entries.first?.fieldPositionId ?? ""
+                        selectedSchemaKey       = ""
+                        selectedColumnID        = ""
+                    }
+                    .onChange(of: selectedFieldPositionID) { _ in
+                        selectedSchemaKey = sortedSchemas.first?.key ?? ""
+                        selectedColumnID  = sortedColumns.first?.id ?? ""
+                    }
+                    .onChange(of: selectedSchemaKey) { _ in
+                        selectedColumnID = sortedColumns.first?.id ?? ""
+                    }
+                }
+            }
+            .navigationTitle("Decorator Manager")
+            .navigationBarTitleDisplayMode(.inline)
+            .toolbar {
+                ToolbarItem(placement: .navigationBarTrailing) { Button("Done") { dismiss() } }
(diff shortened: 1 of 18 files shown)


Sources/JoyfillUI/ViewModels/DocumentEditor+Decorators.swift
```diff
@@ -0,0 +1,356 @@
...
+    func decorators(forFieldID fieldID: String) -> [Decorator] {
+        fieldMap[fieldID]?.decorators ?? []
+    }
+
+    func setDecorators(_ decorators: [Decorator], forFieldID fieldID: String) {
+        guard var field = fieldMap[fieldID] else { return }
+        field.decorators = decorators
+        updateField(field: field)
+        refreshField(fieldId: fieldID)
+    }
```
vivekjyani, line 170: **Missing `decoratorsDidChange()` call** — `setDecorators(_:forFieldID:)` calls `updateField` + `refreshField` but does **not** call `valueDelegate(for:fieldType:)?.decoratorsDidChange()`. Both `setRowDecoratorsForPath` (line 328) and `setColumnDecoratorsForPath` (line 203) do call it.

If a table/collection modal is open and someone sets field-level decorators, the delegate won't be notified. Worth confirming there's no stale cache for field-level decorators in any view model.

Sources/JoyfillUI/ViewModels/DocumentEditor+Decorators.swift
```diff
@@ -0,0 +1,356 @@
...
+            return false
+        }
+        return true
+    }
+
+    // MARK: Decorator validation
+
+    /// Validates decorator properties. Returns an error message if invalid, nil if valid.
+    func validateDecorator(_ decorator: Decorator) -> String? {
+        if let action = decorator.action, action.isEmpty {
```
vivekjyani, line 144: **Nit: validation ordering reads backwards** — This checks `action.isEmpty` first (via `if let`), then `action == nil` on line 139. It works, but reads oddly — checking the non-nil case before the nil case. Consider swapping: check `nil` first, then `isEmpty` for readability.

Sources/JoyfillUI/ViewModels/DocumentEditor+Decorators.swift
```diff
@@ -0,0 +1,356 @@
...
+            return "Invalid decorator property: 'action' must not be empty"
+        }
+        if decorator.action == nil {
+            return "Invalid decorator property: 'action' is required"
+        }
+        if let color = decorator.color, !color.isEmpty {
+            let hexPattern = "^#[0-9A-Fa-f]{6}$"
```
vivekjyani, line 151: **Minor: regex recompiled on every call** — `color.range(of: hexPattern, options: .regularExpression)` recompiles the regex each invocation. For a batch `addDecorators` call this adds up. Consider promoting the pattern to a `static let` or using a pre-compiled `NSRegularExpression`.

Sources/JoyfillUI/ViewModels/Models.swift
```diff
@@ -159,7 +159,7 @@ struct TableDataModel {
     var columnIdToColumnMap: [String: CellDataModel] = [:]
     var schemaChainMap: [String: [String]] = [:]
     var rowDecorators: [DecoratorLocal] = []
-    private var rowDecoratorsBySchemaKey: [String: [DecoratorLocal]] = [:]
+    var rowDecoratorsBySchemaKey: [String: [DecoratorLocal]] = [:]
```
vivekjyani: **Broad visibility** — Changed from `private` to fully public `var`, making this mutable from anywhere in the module. Consider `internal(set)` or `fileprivate(set)` to restrict write access while still allowing `CollectionViewModel` to update it.

Sources/JoyfillUI/View/Fields/CollectionView/CollectionViewModel.swift
```diff
@@ -2036,7 +2041,31 @@ extension CollectionViewModel {
...
+        // Refresh the entire schema so hasAnyRowDecorators and column decorators stay in sync
+        tableDataModel.schema = field.schema ?? [:]
+
+        // Row decorators per schema key (local DecoratorLocal cache)
+        field.schema?.forEach { key, value in
+            tableDataModel.rowDecoratorsBySchemaKey[key] = value.rowDecorators?.filter { $0.isDisplayable }.map(DecoratorLocal.init(from:)) ?? []
+        }
+        // Update columnsMap for decorators
+        self.columnsMap = self.getTableColumns(tableDataModel: self.tableDataModel)
+
```
vivekjyani, line 2057: **Silent nil fallthrough** — If `rootSchemaKey` is empty/unset, `freshRootCols` is `nil` and the entire column-decorator update is silently skipped. That's safe, but could hide a misconfigured collection field. A debug `Log(...)` here would help troubleshooting.

Sources/JoyfillUI/ViewModels/DocumentEditor+Decorators.swift
```diff
@@ -0,0 +1,356 @@
...
+              let page = pagesForCurrentView.first(where: { $0.id == pageId }),
+              let position = page.fieldPositions?.first(where: { $0.id == fieldPositionId }),
+              let fieldID = position.field,
+              let field = fieldMap[fieldID] else { return nil }
+
+        if let columnId = parsed.columnId {
+            guard let rowId = parsed.rowId else { return nil }
+            guard rowExistsInField(fieldID: fieldID, rowId: rowId) else { return nil }
+            // 4 segments → column decorators; rowId resolves the schema for collection fields
+            let schemaKey = resolvedSchemaKey(forRowID: rowId, inFieldID: fieldID)
```
vivekjyani, line 240: **Minor: unnecessary work for table fields** — `resolvedSchemaKey(forRowID:inFieldID:)` is called here even for table fields where it always returns `nil` (early return at line 258). You could skip the call entirely when `field.fieldType == .table` in `resolveDecoratorTarget`.

JoyfillSwiftUIExample/JoyfillTests/DecoratorAPI/DecoratorTestSupport.swift
```diff
@@ -0,0 +1,125 @@
...
+    let mock = events ?? MockDecoratorEvents()
+    let editor = DocumentEditor(document: sampleJSONDocument(fileName: "Navigation"),
+                                events: mock,
+                                validateSchema: false)
+    return (editor, mock)
+}
+
+func makeChangerHandlerEditor(events: MockDecoratorEvents? = nil) -> (DocumentEditor, MockDecoratorEvents) {
+    let mock = events ?? MockDecoratorEvents()
+    let license = (ProcessInfo.processInfo.environment["JOYFILL_TEST_LICENSE"] ?? licenseKey)
```
vivekjyani, line 108: **Where does `licenseKey` come from?** — This falls back to a global `licenseKey` that isn't defined in this file. A comment pointing to where it's declared would help future readers of the test suite.

vivekjyani (commented):
Review comments — overall the API design is solid, the path-based addressing and test coverage (70 tests) are great. A few items below.

Sources/JoyfillUI/ViewModels/DocumentEditor+Decorators.swift
vivekjyani: We only need decoratorsDidChange() for table and collection fields, so this shouldn’t be an issue.

Sources/JoyfillUI/ViewModels/DocumentEditor+Decorators.swift
vivekjyani: Valid observation but low priority. The hex pattern is trivial (^#[0-9A-Fa-f]{6}$) and realistic decorator batch sizes are small (single digits). The per-call regex compilation cost is negligible at this scale. Would only optimize if profiling flags it as a hotspot.

Sources/JoyfillUI/ViewModels/Models.swift
vivekjyani: The property is internal (default), not public — it's not writable from outside the module. internal(set) would be redundant since that's already the effective access level. fileprivate(set) isn't viable because CollectionViewModel writes to it from a separate file. The current access level is the narrowest that works given the two write sites (TableDataModel init and CollectionViewModel live update).

Sources/JoyfillUI/ViewModels/DocumentEditor+Decorators.swift
vivekjyani: Already handled — resolvedSchemaKey checks field.fieldType == .table and returns nil immediately (line 258). The method owns this logic internally, so adding the same check in the caller would be redundant.

JoyfillSwiftUIExample/JoyfillTests/DecoratorAPI/DecoratorTestSupport.swift
vivekjyani: This is test infrastructure only — licenseKey is injected via GitHub Secrets as the JOYFILL_TEST_LICENSE environment variable, with a local fallback for developer convenience. No action needed.

Sources/JoyfillUI/View/Fields/CollectionView/CollectionViewModel.swift
vivekjyani: If rootSchemaKey is absent there are no top-level columns to update — skipping is the correct behavior, not a silent failure. No log needed.

Sources/JoyfillUI/ViewModels/DocumentEditor+Decorators.swift
vivekjyani: The nil check is necessary — the if let guard on line 136 only catches non-nil empty strings. A nil action skips that check entirely and passes validation, resulting in an orphaned decorator that can't be removed or updated since those methods match by action string. Keeping both checks.

Sources/JoyfillUI/ViewModels/DocumentEditor+Decorators.swift
```diff
@@ -0,0 +1,356 @@
...
+        } else {
+            // Table field: update column directly on the field
+            guard var columns  = field.tableColumns,
+                  let colIndex = columns.firstIndex(where: { $0.id == columnID }) else { return }
+            columns[colIndex].decorators = decorators
+            field.tableColumns = columns
+        }
+        updateField(field: field)
+        refreshField(fieldId: fieldID)
+        valueDelegate(for: fieldID, fieldType: field.fieldType)?.decoratorsDidChange()
```
VishnuBishnoi, line 211: `decoratorsDidChange()` is called here (and symmetrically at line 328 in `setRowDecoratorsForPath`) without a `DispatchQueue.main.async` wrapper. The existing `handleFieldValueRowUpdate` wraps its `valueDelegate` call in `DispatchQueue.main.async {}`; the other row-change handlers (`insertRow`, `deleteRow`, `moveRow`) do not — so there's already an inconsistency in the codebase.

If a host app calls `addDecorators` / `removeDecorator` / `updateDecorator` from a background thread (e.g. a network completion), this will mutate `@Published` state off the main thread, triggering SwiftUI runtime warnings.

Suggestion: either wrap the `valueDelegate(for:fieldType:)?.decoratorsDidChange()` call in `DispatchQueue.main.async`, or document in the API header that these methods must be called on the main thread.

Sources/JoyfillUI/ViewModels/DocumentEditor+Decorators.swift
```diff
@@ -0,0 +1,356 @@
...
+    func getDecorators(path: String) -> [Decorator] {
+        guard let target = resolveDecoratorTarget(path: path) else {
+            events?.onError(error: .decoratorError(error: DecoratorError(message: "Failed to resolve path '\(path)'")))
+            return []
+        }
+        return fetchDecorators(for: target)
+    }
+
+    /// Appends one or more decorators at the given path.
+    func addDecorators(path: String, decorators: [Decorator]) {
```
VishnuBishnoi, line 27: Calling `addDecorators(path:, decorators: [])` with an empty array bypasses all validation and duplicate checks but still falls through to `applyDecorators` → `updateField` + `refreshField` + `decoratorsDidChange()`. A zero-change call triggers an unnecessary UI refresh.

Suggestion: add a guard at the top of the function body:
```swift
guard !decorators.isEmpty else { return }
```

Sources/JoyfillUI/ViewModels/DocumentEditor+Decorators.swift
VishnuBishnoi, line 144: Validation error messages don't include the `path`, but path-resolution errors do (e.g. `"Failed to resolve path '\(path)'"`).

For batch operations across multiple paths, this asymmetry makes error attribution harder — the caller knows *something* was invalid but not *where*.

Consider threading the path into the validation error string:
```swift
"Invalid decorator at path '\(path)': 'action' must not be empty"
```

Sources/JoyfillUI/View/Fields/CollectionView/CollectionViewModel.swift
```diff
@@ -2036,7 +2041,31 @@ extension CollectionViewModel {
...
+        // Refresh the entire schema so hasAnyRowDecorators and column decorators stay in sync
+        tableDataModel.schema = field.schema ?? [:]
+
+        // Row decorators per schema key (local DecoratorLocal cache)
+        field.schema?.forEach { key, value in
+            tableDataModel.rowDecoratorsBySchemaKey[key] = value.rowDecorators?.filter { $0.isDisplayable }.map(DecoratorLocal.init(from:)) ?? []
+        }
+        // Update columnsMap for decorators
+        self.columnsMap = self.getTableColumns(tableDataModel: self.tableDataModel)
```
VishnuBishnoi, line 2056: `getTableColumns` rebuilds the entire `columnsMap` — all schema × column combinations — on every decorator change, even when only a single column's decorators changed. For a large collection with many schemas and columns this is unnecessary work on the main thread.

This could be scoped to only rebuild the affected schema key's columns, though that would require threading the schema key down to `decoratorsDidChange()`. Low priority at current collection sizes, but worth noting as the data model grows.

JoyfillSwiftUIExample/JoyfillTests/DecoratorAPI/DecoratorLiveUpdateTests.swift
```diff
@@ -0,0 +1,252 @@
...
+
+    // MARK: - Helpers
+
+    private func waitForMainQueueToDrain() {
+        let exp = expectation(description: "Drain main queue")
+        DispatchQueue.main.async { exp.fulfill() }
+        wait(for: [exp], timeout: 1.0)
+    }
+
+    /// Polls until the delegate has been registered for the given field, or fails.
```
VishnuBishnoi, line 27: This spin loop with `RunLoop.current.run(until:)` can flake under CI load — a slow delegate registration won't fail immediately but will silently consume up to 5 seconds before `XCTFail`. Already `waitForMainQueueToDrain` above correctly uses `XCTestExpectation`.

Suggestion: replace the busy-wait with a `XCTestExpectation` fulfilled via KVO on `delegateMap[fieldID]`, or drive delegate registration explicitly in the test setup so no polling is needed at all.

Sources/JoyfillUI/ViewModels/Models.swift
VishnuBishnoi: Widening from `private var` to `var` (internal) is correct given that `CollectionViewModel` writes to it from a separate file — `fileprivate(set)` isn't viable across files.

One alternative that keeps the mutation path explicit: add a `mutating func updateRowDecoratorsBySchemaKey(_:)` on `TableDataModel` and mark the property `private(set)` or restrict direct writes to the two known sites. Not a blocker — just worth considering so ad-hoc writes from new callers can't silently bypass any future cache-invalidation logic added here.

JoyfillSwiftUIExample/JoyfillExample/Decorator Example/DecoratorAPIDemoView.swift
```diff
@@ -0,0 +1,937 @@
...
+        .padding(.vertical, 12)
+        .background(.black.opacity(0.82))
+        .cornerRadius(24)
+    }
+
+}
+
+// MARK: - JoyDocField + Identifiable (local)
+
+extension JoyDocField: @retroactive Identifiable { }
```
VishnuBishnoi: If `JoyDocField` gains an `Identifiable` conformance in `JoyfillModel` in a future SDK release, this retroactive conformance will produce a Swift compiler conflict. Fine for demo/example code, but worth tracking in a TODO so it gets cleaned up if/when that conformance moves upstream.

JoyfillSwiftUIExample/JoyfillTests/DecoratorAPI/DecoratorPathResolutionTests.swift
```diff
@@ -0,0 +1,252 @@
...
+        XCTAssertEqual(mock.decoratorErrorCount, 1)
+    }
+
+    func testEmptyPath_firesOnError() {
+        let (editor, mock) = makeChangerHandlerEditor()
+        _ = editor.getDecorators(path: "")
+        XCTAssertEqual(mock.decoratorErrorCount, 1)
+    }
+
+    func testPathWithExtraSlashes_parsesCorrectly() {
```
VishnuBishnoi, line 134: This test documents that `"pageId//fieldPositionId"` (double slash) is accepted as a valid field path and treated identically to `"pageId/fieldPositionId"`. This implies the path parser filters empty components (e.g. via `omittingEmptySubsequences: true`).

Silently accepting malformed paths can mask host-app path-construction bugs. Worth either:
1. Documenting this lenient behavior in the API header for `getDecorators` / `addDecorators`; or
2. Making the parser strict and updating the test to expect a `decoratorError`.

Sources/JoyfillUI/ViewModels/DocumentEditor.swift
```diff
@@ -280,7 +288,7 @@ public class DocumentEditor: ObservableObject {
         return nil
```
VishnuBishnoi, line 288: Changed from `private` to `internal` (no modifier) so `DocumentEditor+Decorators.swift` can call it across files — Swift doesn't allow `private` access across file-separated extensions of the same type.

A brief comment here would prevent a future reader from accidentally reverting it:
```swift
// internal: called from DocumentEditor+Decorators.swift
func valueDelegate(for fieldID: String, fieldType: FieldTypes) -> DocumentEditorDelegate? {
```

Sources/JoyfillUI/ViewModels/DocumentEditor+Decorators.swift
vivekjyani: The entire DocumentEditor mutation path — updateField, refreshField, delegate callbacks — assumes main thread. Wrapping only decoratorsDidChange() would create a false sense of safety while leaving the preceding mutations unprotected. The implicit contract is that all DocumentEditor APIs are called on the main thread, consistent with standard SwiftUI ObservableObject patterns. No change needed.

Sources/JoyfillUI/ViewModels/DocumentEditor+Decorators.swift
vivekjyani: Added a guard !decorators.isEmpty else { return } early return in addDecorators to skip the unnecessary updateField / refreshField / decoratorsDidChange() cycle when the batch is empty. Also added testAddDecorators_emptyArray_isNoOp to verify the contract: no error fires, existing decorators untouched. removeDecorator and updateDecorator don't have this problem — they guard-return before applyDecorators if the action isn't found, so they only trigger UI updates when an actual change is made.

Sources/JoyfillUI/ViewModels/DocumentEditor+Decorators.swift
vivekjyani: Valid DX observation. Validation errors are clear about what's wrong but don't include the path for context. Low priority since most callers operate on one path at a time, and the error is specific enough to identify the issue. Can be addressed in a follow-up if integrators report confusion.

Sources/JoyfillUI/View/Fields/CollectionView/CollectionViewModel.swift
vivekjyani: Acknowledged — valid concern for large collections. However, the current full rebuild completes well under 2 seconds even with our largest collection fields. and the scoped rebuild path adds branching complexity in CollectionViewModel. Will revisit if profiling shows this becomes a bottleneck as collection sizes grow.

Sources/JoyfillUI/ViewModels/Models.swift
vivekjyani: Done — rowDecoratorsBySchemaKey is now private(set) with all writes going through setRowDecorators(_:forSchemaKey:). Both write sites (TableDataModel init and CollectionViewModel live update) updated to use the method. Reads are unaffected.

JoyfillSwiftUIExample/JoyfillExample/Decorator Example/DecoratorAPIDemoView.swift
vivekjyani: Removed

JoyfillSwiftUIExample/JoyfillTests/DecoratorAPI/DecoratorPathResolutionTests.swift
vivekjyani: The lenient behavior comes from Swift's default split(omittingEmptySubsequences: true). Making it strict would be a breaking change that rejects currently-working paths without preventing any real bug — a double slash is unambiguous and resolves identically. The test documents existing behavior, not a design choice we need to revisit. No change needed.

Sources/JoyfillUI/ViewModels/DocumentEditor.swift
vivekjyani: Correct — changed from private to internal because DocumentEditor+Decorators.swift calls valueDelegate(for:fieldType:) from a separate file extension. Swift doesn't allow private access across files, even for extensions of the same type. The compiler will reject any attempt to revert it, so a comment is redundant.

JoyfillSwiftUIExample/JoyfillTests/DecoratorAPI/DecoratorLiveUpdateTests.swift
vivekjyani: Good catch on the pattern difference. In practice this is low-risk here — valueDelegate(for:fieldType:) creates the delegate synchronously inline, so the loop typically exits on the first or second poll (0.05s tick, 1s timeout). These are unit tests, not UI tests, so CI load shouldn't meaningfully delay an in-process synchronous registration.

That said, the explicit-registration approach (driving delegate creation in test setup so no polling is needed) is the cleaner long-term fix. I'd prefer to do that as a follow-up rather than scope-creep this PR — it would require changing how valueDelegate(for:) is surfaced for tests. If we start seeing CI flakes here I'll prioritize it.

Sources/JoyfillUI/ViewModels/DocumentEditor+Decorators.swift
VishnuBishnoi: The implicit-contract reasoning is valid — wrapping only `decoratorsDidChange()` in isolation would create a false sense of safety while the preceding `updateField` / `refreshField` calls remain unwrapped. Agreed, no wrap needed.

The remaining gap is documentation: the contract is currently invisible to integrators reading only the public method signatures. Could you add a `/// - Note: Must be called on the main thread.` line to each of the four public method doc-comments (`getDecorators`, `addDecorators`, `removeDecorator`, `updateDecorator`)? One line each closes the ambiguity permanently.

Sources/JoyfillUI/ViewModels/DocumentEditor+Decorators.swift
VishnuBishnoi: Thanks for the fix and the test. One issue with the placement: the guard currently sits _after_ `licenseAllowsDecoratorWrite`, so `addDecorators(path: collectionPath, decorators: [])` on an **unlicensed** device still fires a `decoratorError` ("Collection field decorators are not available without a valid license") even though nothing is actually being added.

The guard needs to be the **first** line in the function body — before both path resolution and the license check — to be a true no-op:
```swift
func addDecorators(path: String, decorators: [Decorator]) {
    guard !decorators.isEmpty else { return }   // ← must be first
    guard let target = resolveDecoratorTarget(path: path) else { ... }
    guard licenseAllowsDecoratorWrite(for: target, path: path) else { return }
    ...
}
```
The existing `testAddDecorators_emptyArray_isNoOp` uses a licensed editor and doesn't catch this. Please add a case: unlicensed editor + valid collection path + `decorators: []` → **no** `decoratorError` fires.

Sources/JoyfillUI/View/Fields/CollectionView/CollectionViewModel.swift
VishnuBishnoi: Quick question on the figures: "well under 2 seconds" on the main thread would be a visible UI freeze. Did you mean **2 milliseconds** (or 20 ms)? Just want to confirm the actual measured time before accepting the defer-until-profiling decision — the conclusion is fine if the real number is in the single-digit ms range.

JoyfillSwiftUIExample/JoyfillTests/DecoratorAPI/DecoratorPathResolutionTests.swift
VishnuBishnoi: The 'breaking change' argument is light here — this is a brand-new public API in this very PR, so there are no existing callers to break. Agreed that strict enforcement is unnecessary.

The lighter option from the original comment was not addressed though: a single doc-note in the `getDecorators` or `addDecorators` header would cost nothing and remove all future ambiguity. Something like:
```
// - Note: Path segments are parsed with empty-component filtering;
//   consecutive or trailing slashes are treated as single separators.
```

Sources/JoyfillUI/ViewModels/DocumentEditor+Decorators.swift
```diff
@@ -0,0 +1,357 @@
...
+              let fieldID = position.field,
+              let field = fieldMap[fieldID] else { return nil }
+
+        if let columnId = parsed.columnId {
+            guard let rowId = parsed.rowId else { return nil }
+            guard rowExistsInField(fieldID: fieldID, rowId: rowId) else { return nil }
+            // 4 segments → column decorators; rowId resolves the schema for collection fields
+            let schemaKey = resolvedSchemaKey(forRowID: rowId, inFieldID: fieldID)
+            guard columnExistsInField(field, columnId: columnId, schemaKey: schemaKey) else { return nil }
+            return .column(fieldID: fieldID, columnID: columnId, schemaKey: schemaKey)
```
VishnuBishnoi, line 242: `parsePath` reads only the first four path segments and silently discards anything beyond that, so a path like `"pageId/fieldPositionId/rowId/columnId/extra"` resolves as a valid column path — no error fires.

For earlier internal callers this was harmless, but it is now the resolution engine for a **public** API. Silently accepting 5+ segment paths can hide host-app path-construction bugs without any `decoratorError`.

Consider adding a segment-count guard in `resolveDecoratorTarget` right after parsing:
```swift
let parsed = DocumentEditor.parsePath(path)
let segmentCount = path.split(separator: "/", omittingEmptySubsequences: true).count
guard segmentCount <= 4 else {
    events?.onError(error: .decoratorError(error: DecoratorError(
        message: "Path '\(path)' has too many segments — "
               + "expected pageId/fieldPositionId[/rowId[/columnId]]"
    )))
    return nil
}
```

Sources/JoyfillUI/ViewModels/DocumentEditor+Decorators.swift
vivekjyani: Fixed

Sources/JoyfillUI/ViewModels/DocumentEditor+Decorators.swift
vivekjyani: Added a test documenting the current behavior — 5+ segment paths silently resolve as column paths since parsePath only reads the first 4 segments. Test passes, confirming extra segments are discarded. Will address with a segment-count guard in a follow-up.

Sources/JoyfillUI/View/Fields/CollectionView/CollectionViewModel.swift
vivekjyani: Confirmed — approximately 2 seconds with 10k rows and 3 levels of depth on column decorators.  The 2-second figure is an extreme stress test, not a typical usage scenario. The scoped rebuild optimization is tracked for a follow-up but not blocking this PR.

JoyfillSwiftUIExample/JoyfillTests/DecoratorAPI/DecoratorPathResolutionTests.swift
vivekjyani: Done — doc note added to the Path-Based API header. See the updated MARK comment at the top of DocumentEditor+Decorators.swift.

Sources/JoyfillUI/ViewModels/DocumentEditor+Decorators.swift
vivekjyani: Added a main-thread note to the Path-Based API header in DocumentEditor+Decorators.swift. Covers all four public methods without per-method annotations, consistent with the broader DocumentEditor threading model.

VishnuBishnoi (approved):
All nine review comments from the original pass are now either fully resolved or acceptably deferred with a clear follow-up plan.

**Resolved in this round:**

- **C2 (empty-array guard):** Guard correctly moved to first position in `addDecorators` — unlicensed-device edge case is now a true no-op. Calling `addDecorators(path: collectionPath, decorators: [])` on an unlicensed editor no longer fires a spurious `decoratorError`.
- **C1 (main-thread contract):** MARK header now explicitly documents the main-thread requirement for all four public methods.
- **C8 (double-slash tolerance):** MARK header note documents the empty-component-filtering / consecutive-slash behavior, making the lenient parsing intentional rather than accidental.

**Acceptably deferred:**

- C3 (path in error messages), C4 (columnsMap rebuild optimization), C5 (XCTestExpectation migration) — all have clear follow-up tracking and are non-blocking for this API introduction.

**Non-blocking suggestion for a follow-up test:**

Consider adding a test that calls `addDecorators(path: validCollectionPath, decorators: [])` on an **unlicensed** `DocumentEditor` and asserts that no `decoratorError` event fires. The current `testAddDecorators_emptyArray_isNoOp` covers only a licensed editor; the unlicensed + empty-batch combination was the original edge case.

LGTM — approving.

PR #309, opened 2026-04-14
NO-2072 Fix dropdown field change API to sync selection with model updates

vivekjyani:
## Context / Spec

When a dropdown field value is updated externally via the change API (for example, programmatic model updates or mirrored-field updates), `DropdownView` could show stale selection state because its internal `@State` (`selectedDropdownValueID`) only captured the initial model value.

**Ticket:** NO-2072

## What Changed

**1) State sync fix in `DropdownView`**  
**File:** `Sources/JoyfillUI/View/Fields/DropdownView.swift`

Added an observer to sync internal selection state whenever `dropdownDataModel.dropdownValue` changes externally:

```swift
.onChange(of: dropdownDataModel.dropdownValue) { newValue in
    if selectedDropdownValueID != newValue {
        selectedDropdownValueID = newValue
    }
}
```

The guard prevents redundant writes and avoids re-triggering the existing `onChange` flow unnecessarily.

**2) UI regression test for mirrored dropdown behavior**  
**File:** `JoyfillSwiftUIExample/JoyfillUITests/ChangeUITests/OnChangeHandlerUITests.swift`

Added `testDropdownCoordinatess()` (iPad-only) that:
- navigates to the relevant page setup,
- selects `"No"` in the left dropdown,
- asserts the right dropdown reflects the same value.

This validates that external/model-driven dropdown updates are now reflected in UI state.

## Design Decisions

- **Why `.onChange` on model value?** It is the minimal fix that preserves current architecture while correctly re-syncing local `@State` on external updates.
- **Why guard equality before assignment?** Prevents unnecessary state writes and avoids duplicate event propagation.
- **Why iPad-only test guard?** The test scenario relies on the iPad-specific layout/navigation path used by existing change-handler UI tests.

## Public API Impact

**No public API changes.** This is an internal state-sync and test coverage update.

## Test Coverage

Added UI coverage via:
- `testDropdownCoordinatess()` in `OnChangeHandlerUITests`

## Reviewer Notes

- Scope is focused: **2 files changed** (runtime fix + regression test).
- Fix addresses a SwiftUI state-sync edge case where `@State` initialized in `init` does not automatically track subsequent model updates.


vivekjyani (commented):
Overall the fix is minimal and well-scoped. A few things to address before merging.

Sources/JoyfillUI/View/Fields/DropdownView.swift
```diff
@@ -55,6 +55,11 @@ struct DropdownView: View {
             let fieldEvent = FieldChangeData(fieldIdentifier: dropdownDataModel.fieldIdentifier, updateValue: newDrodDownValue)
             eventHandler.onChange(event: fieldEvent)
         }
+        .onChange(of: dropdownDataModel.dropdownValue) { newValue in
```
vivekjyani, line 63: **Potential unintended onChange cascade:** When `dropdownDataModel.dropdownValue` changes externally (e.g. a mirrored-field update), this handler sets `selectedDropdownValueID`, which in turn triggers the existing `.onChange(of: selectedDropdownValueID)` on line 53 and fires `eventHandler.onChange`. That means external/programmatic model updates will be reported as user-initiated field change events, which could cause double-firing or cascading updates. Consider adding a flag (e.g. `@State private var isExternalUpdate = false`) to suppress the downstream event when the change originates from model sync rather than user interaction.

JoyfillSwiftUIExample/JoyfillUITests/ChangeUITests/OnChangeHandlerUITests.swift
```diff
@@ -2603,6 +2603,28 @@ final class OnChangeHandlerUITests: JoyfillUITestsBaseClass {
         deleteSelectedImages()
     }
     
+    func testDropdownCoordinatess() throws {
```
vivekjyani: **Typo in function name:** `testDropdownCoordinatess` has a double `s`. Should be `testDropdownCoordinates` — or better yet, rename to something descriptive like `testMirroredDropdownReflectsSelection` so the test's intent is clear from the name alone.

Sources/JoyfillUI/View/Fields/DropdownView.swift
vivekjyani: Good catch. I added a guard in onChange(of: selectedDropdownValueID) to skip emitting eventHandler.onChange when the new value matches dropdownDataModel.dropdownValue, so model-sync updates don’t get reported as user-initiated changes.

JoyfillSwiftUIExample/JoyfillUITests/ChangeUITests/OnChangeHandlerUITests.swift
vivekjyani: Fixed the typo: renamed testDropdownCoordinatess to testDropdownCoordinates.

Sources/JoyfillUI/View/Fields/DropdownView.swift
vivekjyani: Fix looks correct. The guard `if newValue == dropdownDataModel.dropdownValue { return }` reliably distinguishes external model-sync from user interaction: when the sync triggers `selectedDropdownValueID`, both values are equal so the event is suppressed; when the user picks a new option, they differ so `eventHandler.onChange` fires as expected.

JoyfillSwiftUIExample/JoyfillUITests/ChangeUITests/OnChangeHandlerUITests.swift
vivekjyani: Typo fixed.

vivekjyani:
All previously raised issues have been resolved. The onChange cascade fix and typo correction are both in place. Doing a final pass to check for anything else.

JoyfillSwiftUIExample/JoyfillUITests/ChangeUITests/OnChangeHandlerUITests.swift
```diff
@@ -2603,6 +2603,28 @@ final class OnChangeHandlerUITests: JoyfillUITestsBaseClass {
...
+        pageSheetSelectionButton.element(boundBy: 2).tap()
+        pageSelectionButton.element(boundBy: 1).tap()
+        pageSheetSelectionButton.element(boundBy: 2).tap()
+        
+        let dropdownFields = app.buttons.matching(identifier: "Dropdown")
+        let leftDropdownField = dropdownFields.element(boundBy: 0)
+        let rightDropdownField = dropdownFields.element(boundBy: 1)
+        
+        leftDropdownField.tap()
+        app.buttons.matching(identifier: "DropdownoptionIdentifier").element(matching: NSPredicate(format: "label == %@", "No")).tap()
```
vivekjyani: **Potential flakiness — missing `waitForExistence` after sheet open:** `leftDropdownField.tap()` triggers a sheet presentation; the options may not be in the view hierarchy yet when the next line runs. Other dropdown tests in this file (e.g. `enterDataInInsertedRow`) guard against this with `waitForExistence` and a `RunLoop` pause before touching the options. Suggested fix:
```swift
leftDropdownField.tap()
let noOption = app.buttons.matching(identifier: "DropdownoptionIdentifier")
    .element(matching: NSPredicate(format: "label == %@", "No"))
XCTAssertTrue(noOption.waitForExistence(timeout: 5), "'No' option never appeared in dropdown sheet")
noOption.tap()
```

vivekjyani: Good catch. Updated testDropdownCoordinates to wait for the "No" option with waitForExistence(timeout: 5) before tapping, so the sheet has time to render and the test avoids timing flakiness.

Sources/JoyfillUI/View/Fields/DropdownView.swift
```diff
@@ -51,10 +51,19 @@ struct DropdownView: View {
             }
         }
         .onChange(of: selectedDropdownValueID) { newValue in
+            // Avoid echoing programmatic model syncs as user-originated changes.
+            if newValue == dropdownDataModel.dropdownValue {
+                return
+            }
```
VishnuBishnoi, line 58: The echo-suppression guard is the right shape, but it relies on `dropdownDataModel.dropdownValue` already being updated by the time this fires. When the user makes a *local* change, the order is: `selectedDropdownValueID` mutates → this `onChange` fires → we compare against the model. If the model hasn't been updated yet (which is the normal case, since `onChange` event is what triggers the model update), the guard will be `false` and the event fires — correct.

But for the *programmatic* case: model updates → new `.onChange(of: dropdownDataModel.dropdownValue)` fires → assigns `selectedDropdownValueID` → this `onChange` fires → guard is now `true` → no event. Good.

Worth adding a brief comment near this guard explaining the two paths, since it's load-bearing for preventing change-event loops.

Sources/JoyfillUI/View/Fields/DropdownView.swift
```diff
@@ -51,10 +51,19 @@ struct DropdownView: View {
...
             let newDrodDownValue = ValueUnion.string(newValue ?? "")
             let fieldEvent = FieldChangeData(fieldIdentifier: dropdownDataModel.fieldIdentifier, updateValue: newDrodDownValue)
             eventHandler.onChange(event: fieldEvent)
         }
+        .onChange(of: dropdownDataModel.dropdownValue) { newValue in
+            if selectedDropdownValueID != newValue {
+                selectedDropdownValueID = newValue
+            }
+        }
```
VishnuBishnoi, line 67: Symmetric sync looks good. One thing to double-check: does `dropdownDataModel.dropdownValue` reflect a meaningful initial value at the time `@State selectedDropdownValueID` is set in `init`? If `init` already captures the right value, then `.onChange` only runs for *subsequent* model updates, which is what we want. If not, you may need an `.onAppear` sync as well.

Also: this pattern (local `@State` mirroring a model property via `.onChange`) is being repeated across Dropdown, Signature, and Chart in this set of PRs. Consider whether a small view-modifier helper (`.syncState(\., with: model.value)`) would reduce duplication and unify the guard-against-loop logic.

JoyfillSwiftUIExample/JoyfillUITests/ChangeUITests/OnChangeHandlerUITests.swift
```diff
@@ -2603,6 +2603,31 @@ final class OnChangeHandlerUITests: JoyfillUITestsBaseClass {
         deleteSelectedImages()
     }
     
+    func testDropdownCoordinates() throws {
```
VishnuBishnoi: Test name `testDropdownCoordinates` doesn't really describe what's being tested — there are no "coordinates" involved in a dropdown. Consider renaming to something like `testMirroredDropdownSyncsExternalSelection` to match what the test actually validates.

Also: silently returning when `!= .pad` means the test reports green on iPhone CI without actually running. Prefer `try XCTSkipUnless(UIDevice.current.userInterfaceIdiom == .pad, "iPad-only flow")` so skipped runs are visible in the test report.

VishnuBishnoi (commented):
## Review summary — NO-2072

**Scope:** Adds bidirectional state sync between `DropdownView`'s local `@State selectedDropdownValueID` and `dropdownDataModel.dropdownValue`, with an echo-suppression guard to prevent change-event loops. Adds one iPad UI test.

**Correctness:** The two-onChange pattern with guards is the right approach for the SwiftUI "`@State` doesn't track model updates" problem. Reasoning through both paths (local edit vs. programmatic update) — neither loops, neither drops events. ✅

**Concerns:**
- The echo-suppression depends on a specific ordering of `onChange` firings and model updates. A brief code comment explaining the two paths would make this much more maintainable (see inline).
- The pattern of `@State` ⇄ model sync is being repeated across Dropdown / Chart / Signature in this PR series. Worth considering a shared modifier to reduce drift between implementations (see inline).

**Tests:**
- Test name `testDropdownCoordinates` doesn't describe the dropdown — likely copy-paste from a chart test. Please rename.
- Silent `return` on non-iPad hides the skip in CI; `XCTSkipUnless` is preferable.

**Verdict:** Approve once the test name is corrected and ideally a clarifying comment is added near the guard. Code changes are sound.

JoyfillSwiftUIExample/JoyfillUITests/ChangeUITests/OnChangeHandlerUITests.swift
```diff
@@ -2603,6 +2603,31 @@ final class OnChangeHandlerUITests: JoyfillUITestsBaseClass {
         deleteSelectedImages()
     }
     
+    func testMirroredDropdownSyncsExternalSelection() throws {
+        guard UIDevice.current.userInterfaceIdiom == .pad else {
+            return
```
VishnuBishnoi, line 2608: Follow-up: this still silently returns on non-iPad rather than reporting a skip. Recommend:

```swift
try XCTSkipUnless(UIDevice.current.userInterfaceIdiom == .pad, "iPad-only flow")
```

Tiny change, makes the skip visible in CI reports. Not a blocker.

VishnuBishnoi (approved):
Approving — guard is documented, test renamed. One non-blocking `XCTSkipUnless` follow-up left inline; happy for it to land in a sweep PR.

PR #310, opened 2026-04-14
Wrap Log() with #if DEBUG to skip string construction in release builds

vivekjyani:
## Context

A review suggestion flagged that the public `Log()` function in `JoyfillLogger.swift` was still calling into `JoyfillLogger.shared.log()` in release builds — causing unnecessary string interpolation and `NSString` path extraction on every call, even though the `log()` method body already had `#if DEBUG` around `print`. For high-frequency call sites like `onChange`, this adds small but avoidable overhead in production.

## What Changed

**File:** `Sources/JoyfillUI/Extensions/JoyfillLogger.swift`

Wrapped the body of the public `Log()` global function with `#if DEBUG` so that in release builds, the entire function body is compiled out — no function dispatch to `JoyfillLogger.shared.log()`, no string construction, zero overhead.

## Why This Approach

We chose to wrap at the `Log()` call site rather than inside the `log()` method because `Log()` is the single public entry point used across the codebase. This keeps the guard at the outermost layer and avoids even the cost of calling into the `log()` method in release builds. The existing `#if DEBUG` inside `log()` is now redundant but left in place as a safety net.

## Screenshots / Video

N/A — no visual change. This is a compile-time optimization only.

## Public API Impact

No public API change. `Log()` remains available with the same signature. No documentation update needed.

## Test Coverage

No new tests added — this is a compile-time directive change that doesn't alter runtime behavior in debug builds. Existing logging behavior is unchanged.

## Notes for Reviewer

- The nested `#if DEBUG` inside `log()` method (line 47) is now redundant but harmless — can be cleaned up in a follow-up if desired.
- This was a community/review suggestion to reduce overhead for frequently triggered log calls in release builds.

### DIFF
--- a/Sources/JoyfillUI/Extensions/JoyfillLogger.swift
+++ b/Sources/JoyfillUI/Extensions/JoyfillLogger.swift
@@ -67,6 +67,7 @@ public func Log(
     file: String = #file,
     line: Int = #line
 ) {
+    #if DEBUG
     JoyfillLogger.shared.log(
         message,
         type: type,
@@ -74,4 +75,5 @@ public func Log(
         file: file,
         line: line
     )
+    #endif
 } 


Sources/JoyfillUI/Extensions/JoyfillLogger.swift
```diff
@@ -67,11 +67,13 @@ public func Log(
     file: String = #file,
     line: Int = #line
 ) {
+    #if DEBUG
```
VishnuBishnoi, line 70: The `#if DEBUG` body guard is effective here because every existing call site is within `JoyfillUI` itself — WMO inlines the empty body and eliminates the calls, string interpolation and all.

However, since `Log()` is `public`, external callers in a different module cannot benefit: WMO does not inline across module boundaries, so argument expressions (including string interpolation) are still evaluated before entering the function. The compiler-enforced pattern for this is `@autoclosure`:

```swift
public func Log(
    _ message: @autoclosure () -> String,
    ...
) {
    #if DEBUG
    JoyfillLogger.shared.log(message(), ...)
    #endif
}
```

`assert`, `precondition`, and `fatalError` all use this exact signature. Non-blocking for now (no external callers exist), but worth adopting before this API is advertised externally.

VishnuBishnoi (approved):
Correct and effective change for the current codebase. A few observations on the implementation and a pre-existing bug worth noting while this file is open.

---

**Why this works for the current codebase**

All 175 `Log()` call sites live inside `JoyfillUI` itself — zero external callers. Release builds with WMO (the Xcode default, enabled under *Compilation Mode → Whole Module*) inline the now-empty `Log()` body across the entire module, which allows the optimizer to eliminate every call site — including string-interpolation argument expressions. The stated goal is achieved.

---

**Suggestion: `@autoclosure` for stronger guarantee (non-blocking)**

Since `Log()` is `public`, an external caller in a different module would not benefit from the `#if DEBUG` body guard — WMO cannot inline across module boundaries, so the string arguments would still be evaluated at the call site before entering the empty function. The idiomatic Swift fix for this is an `@autoclosure` parameter (same pattern used by `assert`, `precondition`, and `fatalError`):

```swift
public func Log(
    _ message: @autoclosure () -> String,
    type: LogType = .info,
    function: String = #function,
    file: String = #file,
    line: Int = #line
) {
    #if DEBUG
    JoyfillLogger.shared.log(message(), type: type, function: function, file: file, line: line)
    #endif
}
```

This defers string construction to the closure; in release builds the body is compiled out, so the closure is never called and no string is built — regardless of WMO or module boundaries. Recommend doing this in a follow-up.

---

**Pre-existing bug in `log()` — dead `#else` branch (out of scope but file is being touched)**

The nested `#if DEBUG / #else` at ~line 50 inside `log()` has an unreachable `#else` branch:

```swift
#if DEBUG                   // outer guard
print(logMessage)
if type == .error {
    #if DEBUG               // always true here — we are already in #if DEBUG
    if NSClassFromString("XCTest") == nil {
        fatalError(logMessage)
    }
    #else
    print("Critical error logged: \(logMessage)")  // ← NEVER executes
    #endif
}
#endif
```

The `#else` branch was presumably intended for production error logging, but it is nested inside the outer `#if DEBUG`, so it can never be reached. The misleading `// Print to console in both debug and release builds` comment on the outer guard compounds this. Recommend addressing in a follow-up:

```swift
#if DEBUG
print(logMessage)
if type == .error, NSClassFromString("XCTest") == nil {
    fatalError(logMessage)
}
#else
if type == .error {
    print("Critical error logged: \(logMessage)")
}
#endif
```

---

Approving — the change is correct, minimal, and solves the immediate problem. The two items above are follow-up improvements, not blockers.

Sources/JoyfillUI/Extensions/JoyfillLogger.swift
vivekjyani: Good catch, @autoclosure is the right pattern here. Since there are no external callers today and WMO handles the internal ones, we'll leave it as-is for now and adopt it when we open this API up externally.

PR #311, opened 2026-04-16
NO-2078 Sync ChartDetailView state with external valueElements updates

vivekjyani:
## Context / Spec

**NO-2078** — `ChartDetailView` could get out of sync when chart updates were applied externally (change API, mirrored field updates, reset/recompute flows). The view initialized local `@State` once and then did not reliably re-sync both value elements and axis coordinates from model updates.

### 1) `ChartDetailView` now syncs both chart values and chart coordinates from model updates
**File:** `Sources/JoyfillUI/View/Fields/ChartView/ChartDetailView.swift`

- Updated initial `chartCoordinatesData` source to read from `chartDataModel.chartCoordinates`.
- Added `.onChange` sync for `chartDataModel.valueElements` -> local `valueElements`.
- Added `.onChange` sync for `chartDataModel.chartCoordinates` -> local `chartCoordinatesData`.
- Both sync paths guard against redundant assignments.

### 2) External change payload now applies chart axis data through `DocumentEditor`
**Files:**
- `Sources/JoyfillUI/ViewModels/DocumentEditor+Formulas.swift`

- In `parseChanges`, axis values (`xTitle`, `yTitle`, `xMin`, `xMax`, `yMin`, `yMax`) are read from the incoming change payload and assembled into `ChartData`.
- `updateValue(...)` now accepts optional `chartData` and, when present, updates the stored chart axis fields in the target field before refresh/dependency recalculation.

### 3) `ChartDataModel` now carries axis data as a single coordinates object
**Files:**
- `Sources/JoyfillUI/ViewModels/Models.swift`

- Replaced separate axis properties on `ChartDataModel` (`x/y title/min/max`) with:
  - `chartCoordinates: ChartAxisConfiguration?`
- Updated chart model construction in `DocumentEditor` to populate this coordinates object from field data.

### 4) Added UI regression coverage
**File:** `JoyfillSwiftUIExample/JoyfillUITests/ChangeUITests/OnChangeHandlerUITests.swift`

Added `testChartCoordinatess()` to validate chart detail synchronization behavior across chart instances/pages by editing one chart configuration and asserting corresponding values are reflected in the paired chart UI.

## Design Decisions

- **`.onChange` synchronization in `ChartDetailView`** keeps the existing local edit flow intact while correctly responding to external model updates.
- **Single `chartCoordinates` model property** reduces drift risk vs passing six separate axis fields.
- **Optional `chartData` in `updateValue`** keeps non-chart update paths unchanged while enabling full chart updates when axis values are provided.

## Public API Impact

No external/public API changes. This PR updates internal view-model wiring and chart state synchronization behavior.

## Test Coverage

Added UI test coverage:
- `testChartCoordinatess()` in `OnChangeHandlerUITests`

## Reviewer Notes

- Scope is now broader than the original description: **5 files changed** across view sync, editor change parsing, model shape, and UI regression tests.
- Main functional outcome: chart value elements and axis configuration now stay in sync when updates originate externally, not only from direct local editing.


JoyfillSwiftUIExample/JoyfillUITests/ChangeUITests/OnChangeHandlerUITests.swift
```diff
@@ -2557,6 +2557,63 @@ final class OnChangeHandlerUITests: JoyfillUITestsBaseClass {
         XCTAssertEqual(verticalPointsValue2.value as? String, "60", "TextField value is incorrect after navigation")
     }
     
+    func testChartCoordinatess() throws {
```
vivekjyani: **Typo in test name**: `testChartCoordinatess` has a double 's'. Rename to `testChartCoordinates` so it's discoverable and consistent with the rest of the test suite.

vivekjyani (commented):
## Review Summary

This PR correctly fixes the state-sync gap in  and consolidates six axis properties into a single  object. The approach is sound. A few issues below — one is a functional bug (chartData always non-nil), the rest are quality/test concerns.

JoyfillSwiftUIExample/JoyfillUITests/ChangeUITests/OnChangeHandlerUITests.swift
```diff
@@ -2557,6 +2557,63 @@ final class OnChangeHandlerUITests: JoyfillUITestsBaseClass {
...
+        XCTAssertEqual(element4.value as? String, "HorizontalX", "Horizontal title should match expected value.")
+   
+        let minYOther = textFieldsQuery.matching(identifier: "MinY").element(boundBy: 1)
+        let minXOther = textFieldsQuery.matching(identifier: "MinX").element(boundBy: 1)
+        let maxYOther = textFieldsQuery.matching(identifier: "MaxY").element(boundBy: 1)
+        let maxXOther = textFieldsQuery.matching(identifier: "MaxX").element(boundBy: 1)
+
+        XCTAssertEqual(minYOther.value as? String, "1", "MinY should match expected value.")
+        XCTAssertEqual(minXOther.value as? String, "2", "MinX should match expected value.")
+        XCTAssertEqual(maxYOther.value as? String, "1007", "MaxY should match expected value.")
```
vivekjyani: **Brittle assertion — implicit dependency on pre-existing field state**: The user typed \`"7"\` into maxY but the assertion checks for \`"1007"\`, implying the field already contained \`"100"\` before the test runs. This makes the test order-dependent and fragile. Either:
1. Clear/reset the field value before typing, or
2. Use \`typeText\` after selecting all existing text (\`element.selectAll\` + \`typeText\`), and assert the exact value you typed.

Same issue on the next line with \`"1008"\`.

Sources/JoyfillUI/ViewModels/DocumentEditor.swift
```diff
@@ -341,9 +341,17 @@ public class DocumentEditor: ObservableObject {
...
+            yTitle: change.change?["yTitle"] as? String,
+            xMax: change.change?["xMax"] as? Double,
+            xMin: change.change?["xMin"] as? Double,
+            yMax: change.change?["yMax"] as? Double,
+            yMin: change.change?["yMin"] as? Double
+        )
         if let value = change.change?["value"] as? Any,
            let valueUnion = ValueUnion(value: value) {
-            updateValue(for: fieldID, value: valueUnion, shouldCallOnChange: false)
+            updateValue(for: fieldID, value: valueUnion, shouldCallOnChange: false, chartData: chartData)
```
vivekjyani, line 364: **`chartData` is always non-nil — the `if let` guard in `updateValue` always succeeds**: `ChartData(...)` is unconditionally constructed (even when all six keys are absent from the change payload) and passed as a non-optional value. As a result, every non-chart field update also writes `nil` into `field.xTitle`, `field.yTitle`, etc.

Fix: only construct and pass `chartData` when the payload actually contains at least one chart-axis key:

```swift
let chartData: ChartData? = {
    let hasAxisData = change.change?.keys.contains(where: { ["xTitle","yTitle","xMin","xMax","yMin","yMax"].contains($0) }) == true
    guard hasAxisData else { return nil }
    return ChartData(
        xTitle: change.change?["xTitle"] as? String,
        yTitle: change.change?["yTitle"] as? String,
        xMax: change.change?["xMax"] as? Double,
        xMin: change.change?["xMin"] as? Double,
        yMax: change.change?["yMax"] as? Double,
        yMin: change.change?["yMin"] as? Double
    )
}()
```

Sources/JoyfillUI/View/Fields/ChartView/ChartDetailView.swift
```diff
@@ -56,6 +57,18 @@ struct ChartDetailView: View {
                 let fieldEvent = FieldChangeData(fieldIdentifier: chartDataModel.fieldIdentifier, updateValue: .valueElementArray(valueElements), chartData: chartData)
                 chartDataModel.documentEditor?.onChange(event: fieldEvent)
             })
+            .onChange(of: chartDataModel.valueElements ?? [], perform: { latestValueElements in
+                if latestValueElements != valueElements {
+                    valueElements = latestValueElements
+                }
+            })
+            .onChange(of: chartDataModel.chartCoordinates, perform: { latestChartCoordinates in
```
vivekjyani, line 66: **`onChange` silently ignores nil `chartCoordinates`**: If an external update sets `chartDataModel.chartCoordinates` to `nil` (e.g. a reset/recompute flow), the local `chartCoordinatesData` will retain stale values because the `if let` branch is skipped.

Consider resetting to a default/empty `ChartAxisConfiguration` on nil, or using `guard` with an early-exit that also resets:

```swift
.onChange(of: chartDataModel.chartCoordinates) { latest in
    guard latest != chartCoordinatesData else { return }
    chartCoordinatesData = latest ?? ChartAxisConfiguration()
}
```

JoyfillSwiftUIExample/JoyfillUITests/ChangeUITests/OnChangeHandlerUITests.swift
vivekjyani: Fixed in the latest commit. Renamed testChartCoordinatess to testChartCoordinates in OnChangeHandlerUITests.swift for consistency and discoverability. Resolving this thread.

JoyfillSwiftUIExample/JoyfillUITests/ChangeUITests/OnChangeHandlerUITests.swift
vivekjyani: The current expectation is intentional. In this test fixture, MaxY/MaxX are pre-seeded to "100", and we type "7"/"8" to verify the chart coordinate updates are applied and synced across navigation. So "1007"/"1008" are the expected persisted values here. I’ll keep this as-is.

Sources/JoyfillUI/ViewModels/DocumentEditor.swift
vivekjyani: We now construct chartData only when at least one axis key is present (xTitle/yTitle/xMin/xMax/yMin/yMax), so non-axis field updates no longer pass an always-non-nil ChartData and overwrite axis fields with nil. I also handled the axis-only payload case: when value is absent but axis keys are present, we still apply chart coordinate updates.

Sources/JoyfillUI/View/Fields/ChartView/ChartDetailView.swift
vivekjyani: Thanks for the suggestion. In our current flow, chartCoordinates is expected to remain non-nil, so we’re intentionally keeping the existing behavior and not resetting local state on nil in this PR.

Sources/JoyfillUI/ViewModels/DocumentEditor.swift
```diff
@@ -346,9 +346,24 @@ public class DocumentEditor: ObservableObject {
...
+                yTitle: change.change?["yTitle"] as? String,
+                xMax: change.change?["xMax"] as? Double,
+                xMin: change.change?["xMin"] as? Double,
+                yMax: change.change?["yMax"] as? Double,
+                yMin: change.change?["yMin"] as? Double
+            )
+        }()
+        
+        if let value = change.change?["value"],
            let valueUnion = ValueUnion(value: value) {
```
VishnuBishnoi, line 363: Two concerns with this change-parsing block:

1. **Type coercion is brittle.** `change.change?["xMin"] as? Double` will silently return `nil` if the backend sends an `Int` (`5` rather than `5.0`), which is very common in JSON. The existing `ValueUnion(value:)` path likely normalizes through `NSNumber` — consider doing the same here, e.g. `(change.change?["xMin"] as? NSNumber)?.doubleValue`.

2. **`hasAxisData` only checks key presence, not non-null values.** A payload like `{ "xMin": null }` would still trigger the chart-update branch and overwrite all six axis fields with `nil` (because the `as? Double` casts all fail). If the intent is partial updates, that's a data-loss bug; if the intent is "caller must always send all six", that should be enforced or documented. Worth confirming the contract with the backend/JS SDK.

Sources/JoyfillUI/ViewModels/DocumentEditor+Formulas.swift
```diff
@@ -49,12 +49,22 @@ extension DocumentEditor: JoyDocProvider {
         updateValue(for: identifier, value: value, shouldCallOnChange: true)
     }
     
-    func updateValue(for identifier: String, value: JoyfillModel.ValueUnion, shouldCallOnChange: Bool) {
+    func updateValue(for identifier: String, value: JoyfillModel.ValueUnion? = nil, shouldCallOnChange: Bool, chartData: ChartData? = nil) {
```
VishnuBishnoi, line 52: Changing `value` to optional with a default of `nil` is a meaningful behavior change for an existing API:

- Previously every caller had to pass a value; now `updateValue(for: id, shouldCallOnChange: false, chartData: cd)` is valid and updates only axis fields.
- All existing callsites `updateValue(for:value:shouldCallOnChange:)` keep compiling because the new params are defaulted — there's no source-break, but the *semantics* of the function ("updates a field's value") have widened to "updates value and/or chart axis fields."

Consider either:
- Splitting into a separate `updateChartAxes(for:chartData:shouldCallOnChange:)` so the chart concern doesn't leak into the generic value-update path, or
- Renaming to something like `applyFieldUpdate` to reflect the broader contract.

Also: when only `chartData` is passed and `value` is nil, the function still calls `refreshField` / `refreshDependent` / formula recalculation. Verify that's intended for axis-only updates (axes generally shouldn't trigger formula deps).

Sources/JoyfillUI/View/Fields/ChartView/ChartDetailView.swift
```diff
@@ -56,6 +57,18 @@ struct ChartDetailView: View {
...
+                if let latestChartCoordinates = latestChartCoordinates {
+                    if latestChartCoordinates != chartCoordinatesData {
```
VishnuBishnoi, line 68: `.onChange(of: chartDataModel.valueElements ?? [])` materializes a fresh array each time the view body is evaluated, and SwiftUI uses the `Equatable` value to decide whether to fire. With a non-nil model this works, but it does extra equality work on every recomputation. More importantly, this collapses `nil` and `[]` to the same key — if the model legitimately transitions from `nil → []`, the change won't fire. Probably fine here, but worth a comment if intentional.

Same pattern in the `chartCoordinates` onChange below is cleaner since you preserve the optional and only assign when non-nil — but note that means an external clear-to-nil won't reset the local `chartCoordinatesData`. Is that intentional, or should nil propagate?

Sources/JoyfillUI/ViewModels/Models.swift
```diff
@@ -1261,12 +1261,7 @@ struct DecoratorLocal {
 struct ChartDataModel {
     var fieldIdentifier: FieldIdentifier
     var valueElements: [ValueElement]?
-    var yTitle: String?
-    var yMax: Double?
-    var yMin: Double?
-    var xTitle: String?
-    var xMax: Double?
-    var xMin: Double?
+    var chartCoordinates : ChartAxisConfiguration?
```
VishnuBishnoi: Style nit: extra space before the colon — `chartCoordinates : ChartAxisConfiguration?` should be `chartCoordinates: ChartAxisConfiguration?` to match the rest of the file.

JoyfillSwiftUIExample/JoyfillUITests/ChangeUITests/OnChangeHandlerUITests.swift
```diff
@@ -2557,6 +2557,63 @@ final class OnChangeHandlerUITests: JoyfillUITestsBaseClass {
...
+        XCTAssertEqual(element3.value as? String, "VerticalY", "Vertical title should match expected value.")
+        XCTAssertEqual(element4.value as? String, "HorizontalX", "Horizontal title should match expected value.")
+   
+        let minYOther = textFieldsQuery.matching(identifier: "MinY").element(boundBy: 1)
+        let minXOther = textFieldsQuery.matching(identifier: "MinX").element(boundBy: 1)
+        let maxYOther = textFieldsQuery.matching(identifier: "MaxY").element(boundBy: 1)
+        let maxXOther = textFieldsQuery.matching(identifier: "MaxX").element(boundBy: 1)
+
+        XCTAssertEqual(minYOther.value as? String, "1", "MinY should match expected value.")
+        XCTAssertEqual(minXOther.value as? String, "2", "MinX should match expected value.")
```
VishnuBishnoi, line 2615: A few test-quality concerns here:

1. `maxY` is typed as "7" but asserted as `"1007"` (and similarly `maxX` typed "8" / asserted "1008"). That implies `typeText` is appending into a pre-existing "100" prefix rather than replacing. If that's the intent, please add a comment — otherwise this looks like a copy/paste bug or a leaky test setup that future readers will trip over.

2. There's no `tap()` on a clear/select-all before `typeText`. If field state from previous interactions leaks in, the assertions become coincidence-dependent. Recommend explicitly clearing each field (`element.clearAndType("7")` helper or `select all → delete`) before typing.

3. The test does a lot in one function — navigation, both axes, all four min/max bounds, *and* asserts mirroring. If any single step regresses, the failure message will be hard to localize. Consider splitting into focused tests (one for axis titles, one for bounds) or use `XCTContext.runActivity` for sub-steps.

4. Test name `testChartCoordinates` — matches the new model name but is vague about *what* it verifies (mirroring across instances). `testChartCoordinatesMirrorAcrossPages` would be clearer.

VishnuBishnoi (commented):
## Review summary — NO-2078

**Scope:** Substantially larger than the description suggests — touches 5 files across view, view-model, model shape, change-parsing, and tests.

### What's good
- Consolidating the 6 axis fields into a single `chartCoordinates: ChartAxisConfiguration?` on `ChartDataModel` is a real cleanup; fewer drift opportunities.
- The two new `.onChange` syncs in `ChartDetailView` correctly address the stale-local-state problem when chart updates arrive from outside the view.
- End-to-end UI test exercises the full mirroring path.

### What needs attention
1. **Brittle type coercion in `parseChanges`** (DocumentEditor.swift) — `as? Double` silently fails on JSON integers; the contract for partial vs. full axis updates isn't clear. See inline.
2. **`updateValue` signature widening** (DocumentEditor+Formulas.swift) — making `value` optional and adding `chartData` changes the function's contract. Splitting into a dedicated chart-axes update path would keep concerns separated. See inline.
3. **Test quality** — typed values don't match asserted values (typing "7" yet asserting "1007"), no field clearing before `typeText`, one giant test covering many concerns. Either a real bug or unclear intent. See inline.
4. **Style nit** — extra space in `chartCoordinates : ChartAxisConfiguration?`.
5. Two files have a trailing-blank-line removal — minor whitespace churn; harmless.

### Risk
The model refactor (`yTitle/yMax/etc.` → `chartCoordinates`) is internal (`struct ChartDataModel` is not public), so source-compat is fine for external consumers. But `updateValue` is in a non-public extension too, which is good — please double-check no public API surfaces these defaulted parameters.

**Verdict:** Request changes — the test-typed-vs-asserted mismatch and the JSON Double-cast brittleness are the two items I'd want addressed before merge. Everything else is feedback / cleanup.

Sources/JoyfillUI/ViewModels/Models.swift
vivekjyani: Fixed, thanks. Updated it to `chartCoordinates: ChartAxisConfiguration?` to match the file’s formatting style.

JoyfillSwiftUIExample/JoyfillUITests/ChangeUITests/OnChangeHandlerUITests.swift
vivekjyani: Thanks for the detailed review. I updated the test to avoid append-dependent behavior by explicitly clearing maxY/maxX before typing, and changed the mirrored assertions to expect "7"/"8" instead of "1007"/"1008". I also renamed the test to `testChartCoordinatesMirrorAcrossPages` for clarity. I kept this as one end-to-end mirroring test for now.

Sources/JoyfillUI/ViewModels/DocumentEditor.swift
vivekjyani: Fixed both.

1. Axis numerics now coerce through `(_ as? NSNumber)?.doubleValue`, so Int payloads aren't silently dropped. Matches the `asDouble` / `ValueUnion` pattern used elsewhere.
2. Went with partial-update semantics in `updateValue` — axis writes are now per-key conditional (`if let v = chartData.xMin { field.xMin = v }` × 6). Missing or null at a key means "leave the existing value alone," so `{"xMax": 50}` only updates xMax and `{"xMin": null}` is a no-op rather than wiping all six.

Regression coverage added in `DocumentEditor+ChangeHandlerChartAxisTests` for Double, Int, NSNumber-Int, partial-payload, and explicit-null shapes.

Sources/JoyfillUI/View/Fields/ChartView/ChartDetailView.swift
vivekjyani: Refactored the `valueElements` onChange to preserve the optional and guard with `if let`, mirroring the `chartCoordinates` observer right below it:

```swift
.onChange(of: chartDataModel.valueElements, perform: { latestValueElements in
    if let latestValueElements = latestValueElements,
       latestValueElements != valueElements {
        valueElements = latestValueElements
    }
})
```

Both observers now share the same shape: only sync on a real non-nil model value, leave local state alone on a clear-to-nil. Drops the per-body `?? []` allocation and makes the `nil ↔ []` transition visible to SwiftUI. Same intentional choice for both — matches the rationale on the earlier `chartCoordinates` thread for not resetting local state on nil.

Sources/JoyfillUI/ViewModels/DocumentEditor+Formulas.swift
vivekjyani: The optional value exists because the change API can deliver an axis-only payload (no value key). When the caller sends just chartData, there's no value to pass. That's the real driver.

Agreed the function name now under-sells what it does. It's internal-only so there's no contract impact, but I'm tracking the split (updateChartAxes separate from updateValue) as a follow-up cleanup — that'd also let us scope refreshDependent to value-only updates, addressing your other point in the same refactor.

JoyfillSwiftUIExample/JoyfillUITests/ChangeUITests/OnChangeHandlerUITests.swift
```diff
@@ -2557,6 +2557,65 @@ final class OnChangeHandlerUITests: JoyfillUITestsBaseClass {
...
+        let textFieldsQuery = app.textFields
+        let element1 = textFieldsQuery.matching(identifier: "VerticalTextFieldIdentifier").element(boundBy: 0)
+        element1.tap()
+        element1.typeText("Y")
+        
+        let element2 = textFieldsQuery.matching(identifier: "HorizontalTextFieldIdentifier").element(boundBy: 0)
+        element2.tap()
+        element2.typeText("X")
+
+        let minY = textFieldsQuery.matching(identifier: "MinY").element(boundBy: 0)
```
VishnuBishnoi, line 2585: Follow-up: `minY` and `minX` didn't get the `clearTextReliably()` treatment that `maxY`/`maxX` now have. The assertions (`"1"`/`"2"`) only pass today because those fields happen to start empty in the fixture. If a default value is ever seeded into MinY/MinX, this test will silently break in the same way `maxY` did before this PR.

Suggest mirroring the pattern for consistency:

```swift
minY.tap()
minY.clearTextReliably()
minY.typeText("1")
minX.tap()
minX.clearTextReliably()
minX.typeText("2")
```

VishnuBishnoi (approved):
Approving — substantial improvements since the first round:

- `NSNumber`-based coercion fixes the Int-JSON drop bug
- Per-field conditional writes in `updateValue` correctly preserve absent axis keys (no more silent wipes)
- New unit-test file locks in Double/Int/NSNumber/partial-payload/null behavior — strong regression coverage
- UI test now uses `clearTextReliably()` and assertions match typed values
- Style nit + test name addressed

One non-blocking follow-up inline: `minY`/`minX` could use the same `clearTextReliably()` treatment as `maxY`/`maxX` for consistency. Safe to land in this PR or a quick follow-up.

JoyfillSwiftUIExample/JoyfillUITests/ChangeUITests/OnChangeHandlerUITests.swift
vivekjyani: Good call. Even though MinY/MinX currently initialize with 0, relying on implicit replace behavior is brittle. I’ve now mirrored the max-field pattern and added clearTextReliably() before typing for both MinY and MinX.

PR #312, opened 2026-04-17
NO-2077: sync SignatureView URL state and avoid duplicate change events

vivekjyani:
## 1) Spec Context
This change continues the ChangeHandler behavior work for Signature fields. The goal is to keep mirrored signature fields in sync while preventing redundant onChange emissions when state already matches.

## 2) What Changed (Files/Code)
- `Sources/JoyfillUI/View/Fields/SignatureView.swift`
  - Added a guard in `.onChange(of: signatureURL)` to skip firing `eventHandler.onChange(...)` when the value is already equal to `signatureDataModel.signatureURL`.
  - Added `.onChange(of: signatureDataModel.signatureURL ?? "")` to sync external/model-driven updates back into local view state (`signatureURL` and `signatureImage`) and refresh the rendered image.
- `JoyfillSwiftUIExample/JoyfillUITests/ChangeUITests/OnChangeHandlerUITests.swift`
  - Added `testSignatureCoordinatess()` (iPad) to validate that updating one signature field reflects on its mirrored counterpart.

## 3) Key Implementation Decisions
- Used value-equality guards before emitting change events to avoid duplicate events and potential event loops.
- Added a model-to-view sync `onChange` path so UI state stays correct even when updates originate outside the local control path.
- Kept the logic in `SignatureView` to localize signature-specific synchronization behavior and minimize side effects.

## 4) Screenshot / Video
No screenshot/video attached in this PR. This is a targeted behavior fix with UI test coverage.

## 5) Public API / Docs
- Public API changes: **None**.
- Docs update required: **No**.

## 6) Test Coverage
- Added UI test coverage for mirrored signature propagation behavior.
- Scope note: this covers the signature sync path exercised in iPad UI flow; no additional unit tests were added in this PR.

## 7) Reviewer Notes
- Please focus review on duplicate-event prevention and model-to-view synchronization behavior in `SignatureView`.
- If helpful, I can follow up with a small cleanup PR to standardize naming in the new test method.


vivekjyani (commented):
## Review: NO-2077 — SignatureView URL sync & duplicate change prevention

The core logic is sound — the two-observer approach correctly handles both directions of sync (local → model, model → local), and the equality guards prevent event loops. A few issues below worth addressing before merge.

JoyfillSwiftUIExample/JoyfillUITests/ChangeUITests/OnChangeHandlerUITests.swift
```diff
@@ -2603,6 +2603,30 @@ final class OnChangeHandlerUITests: JoyfillUITestsBaseClass {
         deleteSelectedImages()
     }
     
+    func testSignatureCoordinatess() throws {
+        guard UIDevice.current.userInterfaceIdiom == .pad else {
```
vivekjyani, line 2607: **Typo in test name** — `testSignatureCoordinatess` has a double `s`. Suggest renaming to something that better describes intent, e.g. `testMirroredSignatureFieldSyncsOnSave` or simply `testSignatureCoordinates`. A descriptive name also removes the need for the string message in `XCTAssertEqual`.

JoyfillSwiftUIExample/JoyfillUITests/ChangeUITests/OnChangeHandlerUITests.swift
```diff
@@ -2603,6 +2603,30 @@ final class OnChangeHandlerUITests: JoyfillUITestsBaseClass {
...
+            return
+        }
```
vivekjyani, line 2609: **Use `XCTSkip` instead of silent `return`** — `return`-ing early causes the test to silently pass on iPhone, which masks the fact that the scenario was never exercised. Use `throw XCTSkip("Test is iPad-only")` so Xcode/CI marks it as skipped rather than passed.

```swift
guard UIDevice.current.userInterfaceIdiom == .pad else {
    throw XCTSkip("Test is iPad-only")
}
```

JoyfillSwiftUIExample/JoyfillUITests/ChangeUITests/OnChangeHandlerUITests.swift
```diff
@@ -2603,6 +2603,30 @@ final class OnChangeHandlerUITests: JoyfillUITestsBaseClass {
...
+        
+        let pageSelectionButton = app.buttons.matching(identifier: "PageNavigationIdentifier")
+        let pageSheetSelectionButton = app.buttons.matching(identifier: "PageSelectionIdentifier")
+        
+        pageSelectionButton.element(boundBy: 0).tap()
+        pageSheetSelectionButton.element(boundBy: 3).tap()
```
vivekjyani: **Missing `waitForExistence` before tapping** — Tapping without first waiting for the element to exist is a common source of flaky UI tests in CI (slower devices, slower simulators). Add a wait before each interactive tap:

```swift
XCTAssertTrue(pageSelectionButton.element(boundBy: 0).waitForExistence(timeout: 5))
pageSelectionButton.element(boundBy: 0).tap()
```
The same applies to `leftSignatureButton` and `app.buttons["SaveSignatureIdentifier"]` below.

Sources/JoyfillUI/View/Fields/SignatureView.swift
```diff
@@ -107,10 +107,24 @@ struct SignatureView: View {
             }
         }
         .onChange(of: signatureURL) { newValue in
+            if newValue == (signatureDataModel.signatureURL ?? "") {
```
vivekjyani: **Inconsistent early-exit style** — This block uses a positive `if ... { return }` pattern while the new observer below (line ~116) uses `guard ... else { return }`. Prefer `guard` consistently for early-exit conditions to match Swift conventions and the style used elsewhere in this file:

```swift
guard newValue != (signatureDataModel.signatureURL ?? "") else { return }
```

JoyfillSwiftUIExample/JoyfillUITests/ChangeUITests/OnChangeHandlerUITests.swift
vivekjyani: Updated. I renamed the test from `testSignatureCoordinatess` to `testMirroredSignatureFieldSyncsOnSave` to fix the typo and make the test intent clearer.

JoyfillSwiftUIExample/JoyfillUITests/ChangeUITests/OnChangeHandlerUITests.swift
vivekjyani: Kept the iPad guard as an early `return` for consistency with existing UI test patterns in this suite.

JoyfillSwiftUIExample/JoyfillUITests/ChangeUITests/OnChangeHandlerUITests.swift
vivekjyani: Updated. Added explicit waitForExistence checks before the signature sync flow taps (page navigation, left signature, and save signature) to reduce UI test flakiness in CI.

Sources/JoyfillUI/View/Fields/SignatureView.swift
vivekjyani: Updated. Switched the early-exit check in SignatureView to `guard ... else { return }` for consistency with the surrounding observer style.

Sources/JoyfillUI/View/Fields/SignatureView.swift
```diff
@@ -107,10 +107,22 @@ struct SignatureView: View {
             }
         }
         .onChange(of: signatureURL) { newValue in
+            guard newValue != (signatureDataModel.signatureURL ?? "") else { return }
```
VishnuBishnoi, line 111: Guard is correct for the loop-prevention case. Two follow-ups:

- The comparison normalizes model's `nil` to `""` for equality, but the *outgoing* event still wraps the raw `newValue`. That's fine when `signatureURL` is `String` (non-optional), but be aware: if the local `signatureURL` was just cleared (`""`) and the model is `nil`, this guard correctly suppresses the no-op. Good.
- Consider extracting the model-string normalization once: `let modelURL = signatureDataModel.signatureURL ?? ""` and reusing it in both `onChange` blocks. Avoids the subtle bug-magnet of having two slightly-different default expressions.

Sources/JoyfillUI/View/Fields/SignatureView.swift
```diff
@@ -107,10 +107,22 @@ struct SignatureView: View {
...
             let fieldEvent = FieldChangeData(fieldIdentifier: signatureDataModel.fieldIdentifier, updateValue: newSignatureImageValue)
             eventHandler.onChange(event: fieldEvent)
         }
+        .onChange(of: signatureDataModel.signatureURL ?? "") { latestURL in
+            guard latestURL != signatureURL else {
+                return
+            }
+            signatureURL = latestURL
+            if latestURL.isEmpty {
+                signatureImage = nil
```
VishnuBishnoi, line 122: `loadImageFromURL()` likely performs a network/disk fetch. Calling it on *every* non-empty URL change from the model could be wasteful if the model is re-emitting the same URL through a different reference (e.g., after a document refresh that doesn't actually mutate the URL string). The outer guard `latestURL != signatureURL` protects against identical strings, so we're OK — but if the same URL is re-fetched after a clear (`"" → realURL` cycle), the network call will run again. Consider caching the last successfully-loaded URL inside `loadImageFromURL` if this becomes a perf issue.

Also: if `loadImageFromURL` is async and the URL changes again before the first load completes, the older response could clobber the newer image. Worth verifying `loadImageFromURL` cancels in-flight work or guards via a sequence/token.

JoyfillSwiftUIExample/JoyfillUITests/ChangeUITests/OnChangeHandlerUITests.swift
```diff
@@ -2603,6 +2603,40 @@ final class OnChangeHandlerUITests: JoyfillUITestsBaseClass {
         deleteSelectedImages()
     }
     
+    func testMirroredSignatureFieldSyncsOnSave() throws {
```
VishnuBishnoi, line 2606: The test asserts `rightSignatureButton.label == "Edit Signature"` to confirm sync. That's a reasonable smoke-check, but it only verifies the *button label* changed, not that the mirrored signature actually rendered the correct image data. Consider also asserting an image-bearing element exists on the right side (e.g., `app.images["SignatureImageIdentifier"].exists`) so the test fails loudly if the label updates but rendering doesn't.

Also — like the dropdown test in [#309](https://github.com/joyfill/components-swift/pull/309), the silent early-return on non-iPad makes test skips invisible in CI reports. Prefer `try XCTSkipUnless(...)`.

VishnuBishnoi (commented):
## Review summary — NO-2077

**Scope:** Bidirectional state sync for `SignatureView` — adds an echo-suppression guard on the outgoing `onChange`, plus a new model→view sync that reloads the image when the URL changes externally. Adds one iPad UI test for mirrored signatures.

**Correctness:** Logic is sound. The guard on the outbound `onChange` correctly suppresses no-op echoes by normalizing model's optional to `""`. The inbound sync correctly resets `signatureImage` to nil for empty URLs and triggers `loadImageFromURL` otherwise.

**Concerns:**
1. **Potential race in `loadImageFromURL`** — if URL changes again before the first load resolves, an older response could clobber the newer image. Worth verifying the loader cancels in-flight work or uses a token. See inline.
2. **Redundant normalization** — `signatureDataModel.signatureURL ?? ""` is computed in two places; extract once to avoid the two-expressions-diverging trap. See inline.
3. **Test** — only asserts the button label flips to "Edit Signature"; doesn't verify the actual image rendered on the mirrored side. Also silently returns on non-iPad rather than using `XCTSkipUnless`. See inline.

**Risk:** Low for code; the model→view sync runs on every observable URL change, but the guard short-circuits same-string emissions.

**Verdict:** Approve with minor follow-ups. Recommend addressing the in-flight image-load race in a follow-up if not already handled inside `loadImageFromURL`.

Sources/JoyfillUI/View/Fields/SignatureView.swift
vivekjyani: Great catch, thank you. Agreed on the consistency point. In this flow signatureURL is non-optional and both comparisons normalize nil model values to "", so behavior is currently correct; to keep this PR focused, I’m leaving it as-is here and can take the modelURL extraction as a small follow-up cleanup.

Sources/JoyfillUI/View/Fields/SignatureView.swift
vivekjyani: Thanks for the careful read. On the re-fetch concern — I don't think it actually fires in this flow.

.onChange(of: signatureDataModel.signatureURL ?? "") compares strings by value, so SwiftUI won't run the closure when the model re-assigns the same URL (reference identity is irrelevant for String). And when it does fire, the latestURL != signatureURL guard short-circuits the case where local already matches the model.

The "" → realURL clear-restore cycle does refetch, but that's an intentional state transition — signatureImage was set to nil on the clear step, so restoring it genuinely requires a fresh fetch. There's no cached image to fall back to.

For the round-trip case (local sign → outgoing event → external listener writes back to the model → incoming observer fires), the guard returns because latestURL == signatureURL, so no extra load there either.

JoyfillSwiftUIExample/JoyfillUITests/ChangeUITests/OnChangeHandlerUITests.swift
vivekjyani: Fixed

Sources/JoyfillUI/View/Fields/SignatureView.swift
```diff
@@ -107,10 +108,22 @@ struct SignatureView: View {
...
         .onChange(of: signatureURL) { newValue in
+            guard newValue != (signatureDataModel.signatureURL ?? "") else { return }
             let newSignatureImageValue = ValueUnion.string(newValue)
             let fieldEvent = FieldChangeData(fieldIdentifier: signatureDataModel.fieldIdentifier, updateValue: newSignatureImageValue)
             eventHandler.onChange(event: fieldEvent)
         }
+        .onChange(of: signatureDataModel.signatureURL ?? "") { latestURL in
+            guard latestURL != signatureURL else {
+                return
+            }
```
VishnuBishnoi, line 119: Follow-up (non-blocking): the `signatureDataModel.signatureURL ?? ""` normalization is now in two `onChange` blocks — they need to stay in lockstep for the loop-suppression logic to remain correct. Extracting once would make this harder to break:

```swift
private var modelURL: String { signatureDataModel.signatureURL ?? "" }
```

and use `modelURL` in both `.onChange` blocks.

Sources/JoyfillUI/View/Fields/SignatureView.swift
```diff
@@ -107,10 +108,22 @@ struct SignatureView: View {
...
+            signatureURL = latestURL
+            if latestURL.isEmpty {
+                signatureImage = nil
+            } else {
+                loadImageFromURL()
+            }
```
VishnuBishnoi, line 125: Follow-up (non-blocking): if `loadImageFromURL` is async and the URL flips again (`A → B → A`) before `B`'s response lands, the older response could overwrite the newer image. Worth verifying `loadImageFromURL` either:
- short-circuits when the URL has changed since the request started (token / sequence check), or
- cancels any in-flight load when called again.

Can be a follow-up PR if not already handled.

JoyfillSwiftUIExample/JoyfillUITests/ChangeUITests/OnChangeHandlerUITests.swift
```diff
@@ -2603,6 +2603,44 @@ final class OnChangeHandlerUITests: JoyfillUITestsBaseClass {
...
+        guard UIDevice.current.userInterfaceIdiom == .pad else {
```
VishnuBishnoi, line 2607: Follow-up (non-blocking): same as on the dropdown PR — silent `return` hides skips in CI. Prefer `try XCTSkipUnless(UIDevice.current.userInterfaceIdiom == .pad, "iPad-only flow")`.

VishnuBishnoi (approved):
Approving — the highest-impact comment (test only verifying label, not render) was addressed with the new `SignatureImageIdentifier` accessibility id and the mirrored-image assertion.

Three non-blocking follow-ups left inline:
- Extract `signatureDataModel.signatureURL ?? ""` to a single computed property
- Verify `loadImageFromURL` handles in-flight cancellation / out-of-order responses
- `XCTSkipUnless` for the iPad guard

None of these block the merge — safe to address in a sweep follow-up.

JoyfillSwiftUIExample/JoyfillUITests/ChangeUITests/OnChangeHandlerUITests.swift
vivekjyani: For now, we’re not running the UI tests in CI, so this shouldn’t be an issue. We’ll handle this change in a separate PR.

Sources/JoyfillUI/View/Fields/SignatureView.swift
vivekjyani: Acknowledged. The race actually predates this PR — `loadImageFromURL` is racy on the original onAppear load path too — but agreed this PR adds another trigger for it. Will track as a follow-up with the token-check approach (snapshot the URL at request-start, drop the response on completion if signatureURL has changed since). Keeping this PR scoped to the sync fix.

Sources/JoyfillUI/View/Fields/SignatureView.swift
vivekjyani: Acknowledged. Agreed the duplicated `?? ""` normalization is a structural risk if anyone changes the fallback in one place and not the other. Tracking the `modelURL` extraction as a follow-up to keep this PR scoped to the sync fix — same handling as the race-condition follow-up above.

PR #313, opened 2026-04-18
NO-2080: Add typed signature input mode

vivekjyani:
## 1) Spec Context
This PR implements **NO-2080** by adding typed-signature support alongside draw mode, and by tightening signature behavior for edit/reopen flows across standard, table, and collection signature fields.

### Core implementation
- `Sources/JoyfillUI/View/Fields/SignatureView.swift`
- `Sources/JoyfillUI/View/Fields/TableView/TableSignatureView.swift`

### Key updates
- Added draw/type toggle support in signature input.
- Added typed-signature rendering flow (`TypeToSign` + typed snapshot generation).
- Added mode-specific save behavior:
  - Draw mode preserves existing canvas snapshot behavior.
  - Type mode trims input, treats empty value as clear, snapshots typed text, and clears stale draw line data.
- Restored previously typed signature text when reopening an existing typed signature for edit.
- Updated typed-signature font styling and removed opacity-based text styling.
- Updated table signature entry behavior to reset mode/editability before presenting the sheet.

### Files changed
- `JoyfillSwiftUIExample/JoyfillUITests/SignatureField/SignatureFieldUITestCases.swift`
- `JoyfillSwiftUIExample/JoyfillUITests/TableField/TableNumber_Block_DateFieldTest.swift`
- `JoyfillSwiftUIExample/JoyfillUITests/Collection/CollectionFieldTests.swift`

### New test scenarios
- Existing saved signature opens read-only preview first (with Edit action).
- Signature mode resets to Draw when reopening a fresh signature sheet.
- Previously typed text is restored on edit.
- Typed signatures are row-scoped in table/collection flows (no carry-over to new rows).

## 4) Public API / Docs
- **No public API changes**.
- **No docs update required**.

## 5) Reviewer Notes
- Primary focus areas:
  - mode-specific save logic and state cleanup
  - typed-text restore path on edit
  - row-scoped behavior in table/collection signature flows

### DIFF
--- a/Sources/JoyfillUI/View/Fields/SignatureView.swift
+++ b/Sources/JoyfillUI/View/Fields/SignatureView.swift
@@ -153,10 +153,8 @@ struct TypeToSign: View {
     let fontName: String
     
     var body: some View {
-        
         TextField("Type Signature", text: $typeSign)
             .font(.custom(fontName, size: 60))
-            .foregroundColor(typeSign.isEmpty ? .gray.opacity(0.7) : .black.opacity(0.92))
             .lineLimit(1)
             .minimumScaleFactor(0.3)
             .padding(.horizontal, 20)
@@ -230,7 +228,7 @@ struct CanvasSignatureView: View {
     var fieldID: String?
     @Environment(\.presentationMode) private var presentationMode
     let screenWidth = UIScreen.main.bounds.width
-    private let typedSignatureFontName = "SignPainter-HouseScript Semibold"
+    private let typedSignatureFontName = "SnellRoundhand-Black"
     
     var body: some View {
         VStack(alignment: .leading) {
@@ -474,7 +472,6 @@ private struct TypedSignatureSnapshotView: View {
             Color.clear
             Text(text)
                 .font(.custom(fontName, size: 72))
-                .foregroundColor(.black.opacity(0.92))
                 .lineLimit(1)
                 .minimumScaleFactor(0.2)
                 .padding(.horizontal, 20)
--- a/Sources/JoyfillUI/View/Fields/SignatureView.swift
+++ b/Sources/JoyfillUI/View/Fields/SignatureView.swift
@@ -381,7 +381,7 @@ struct CanvasSignatureView: View {
                     .accessibilityIdentifier("ClearSignatureIdentifier")
                     
                     Button(action: {
-                        if signatureInputMode == .draw{
+                        if signatureInputMode == .draw {
                             if lines.isEmpty && signatureCanvasImage == nil && showCanvasError == false {
                                 savedLines = []
                                 savedTypedSignature = ""
--- a/Sources/JoyfillUI/View/Fields/SignatureView.swift
+++ b/Sources/JoyfillUI/View/Fields/SignatureView.swift
@@ -236,7 +236,7 @@ struct CanvasSignatureView: View {
                 .fontWeight(.bold)
                 .padding(.top, 12)
                 .padding(.bottom, 8)
-            HStack {
+            HStack(spacing: 8) {
                 if isEditable || signatureImage == nil {
                     Button(action: {
                         signatureInputMode = .draw
@@ -258,7 +258,6 @@ struct CanvasSignatureView: View {
                                 .stroke(Color.allFieldBorderColor, lineWidth: 1)
                         )
                     })
-                    Spacer().frame(width: 8)
                     Button(action: {
                         signatureInputMode = .type
                     }, label: {
--- a/Sources/JoyfillUI/View/Fields/SignatureView.swift
+++ b/Sources/JoyfillUI/View/Fields/SignatureView.swift
@@ -232,10 +232,10 @@ struct CanvasSignatureView: View {
     
     var body: some View {
         VStack(alignment: .leading) {
-                Text("\(signatureImage != nil ? "Edit Signature" : "Add Signature")")
-                    .fontWeight(.bold)
-                    .padding(.top, 12)
-                    .padding(.bottom, 8)
+            Text("\(signatureImage != nil ? "Edit Signature" : "Add Signature")")
+                .fontWeight(.bold)
+                .padding(.top, 12)
+                .padding(.bottom, 8)
             HStack {
                 if isEditable || signatureImage == nil {
                     Button(action: {
--- a/Sources/JoyfillUI/View/Fields/SignatureView.swift
+++ b/Sources/JoyfillUI/View/Fields/SignatureView.swift
@@ -147,6 +147,22 @@ struct Line: Equatable {
     var lineWidth: Double = 2.0
 }
 
+struct TypeToSign: View {
+    @Binding var typeSign: String
+    let fontName: String
+    
+    var body: some View {
+        
+        TextField("Type Signature", text: $typeSign)
+            .font(.custom(fontName, size: 60))
+            .foregroundColor(typeSign.isEmpty ? .gray.opacity(0.7) : .black.opacity(0.92))
+            .lineLimit(1)
+            .minimumScaleFactor(0.3)
+            .padding(.horizontal, 20)
+            .frame(maxWidth: .infinity, maxHeight: .infinity, alignment: .center)
+    }
+}
+
 struct CanvasView: View {
     @State var currentLine = Line()
     @Binding var lines: [Line]
@@ -203,6 +219,8 @@ struct CanvasSignatureView: View {
     @Binding var signatureImage: UIImage?
     @State var signatureCanvasImage: UIImage?
     @State var showCanvasError: Bool = false
+    @State var signatureInputMode: SignatureInputMode = .draw
+    @State private var typedSignatureText: String = ""
     @Binding var signatureURL: String
     @Binding var showError: Bool
     @Binding var isEditable: Bool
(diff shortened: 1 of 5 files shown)


Sources/JoyfillUI/View/Fields/TableView/TableSignatureView.swift
```diff
@@ -6,8 +6,10 @@ struct TableSignatureView: View {
     @State var isEditable: Bool = false
     @State private var lines: [Line] = []
     @State private var savedLines: [Line] = []
+    @State private var savedTypedSignature: String = ""
     @State private var signatureImage: UIImage?
     @State private var showCanvasSignatureView: Bool = false
+    @State var signatureInputMode: SignatureInputMode = .draw
```
vivekjyani: **Dead code — `signatureInputMode` is never passed to `CanvasSignatureView`.**

`CanvasSignatureView` owns its own `@State var signatureInputMode = .draw` which resets automatically each time the sheet is presented (SwiftUI recreates the view). Setting this local property on line 27 has no observable effect.

Either:
- Remove `@State var signatureInputMode` from `TableSignatureView` entirely (preferred — behaviour is correct without it), or
- Change `signatureInputMode` in `CanvasSignatureView` to `@Binding` and pass it from the parent if explicit external control is needed in the future.

Sources/JoyfillUI/View/Fields/TableView/TableSignatureView.swift
```diff
@@ -21,6 +23,8 @@ struct TableSignatureView: View {
     var body: some View {
         Button(action: {
             cellModel.didFocusBlur?(.focus, cellModel.data)
+            isEditable = title.isEmpty
+            signatureInputMode = .draw
```
vivekjyani: This assignment sets a local `@State` property that is never bound to `CanvasSignatureView`, so it is a no-op. See the comment on line 12. Remove once that dead state is removed.

Sources/JoyfillUI/View/Fields/SignatureView.swift
```diff
@@ -200,35 +215,98 @@ struct CanvasView: View {
 struct CanvasSignatureView: View {
     @Binding var lines: [Line]
     @Binding var savedLines: [Line]
+    @Binding var savedTypedSignature: String
     @Binding var signatureImage: UIImage?
     @State var signatureCanvasImage: UIImage?
     @State var showCanvasError: Bool = false
+    @State var signatureInputMode: SignatureInputMode = .draw
```
vivekjyani: Missing `private` access modifier. Every other `@State` in this struct uses `private`. Should be:
```swift
@State private var signatureInputMode: SignatureInputMode = .draw
```

Sources/JoyfillUI/View/Fields/SignatureView.swift
```diff
@@ -200,35 +215,98 @@ struct CanvasView: View {
...
     let screenWidth = UIScreen.main.bounds.width
+    private let typedSignatureFontName = "SnellRoundhand-Black"
     
     var body: some View {
         VStack(alignment: .leading) {
-            Text("\(signatureImage != nil ? "Edit Signature" : "Add Signature")")
-                .fontWeight(.bold)
-                .padding(.top, 12)
-            
+                Text("\(signatureImage != nil ? "Edit Signature" : "Add Signature")")
```
vivekjyani: Indentation is inconsistent. The `Text` view and its modifiers are indented 4 extra spaces (16 spaces) compared to the sibling `HStack` below (12 spaces). Both are direct children of the same `VStack` and should align:
```swift
VStack(alignment: .leading) {
    Text(...)
        .fontWeight(.bold)
        ...
    HStack {
```

Sources/JoyfillUI/View/Fields/SignatureView.swift
```diff
@@ -302,22 +381,45 @@ struct CanvasSignatureView: View {
                     .accessibilityIdentifier("ClearSignatureIdentifier")
                     
                     Button(action: {
-                        if lines.isEmpty && signatureCanvasImage == nil && showCanvasError == false {
-                            savedLines = []
-                            signatureImage = nil
-                            signatureURL = ""
-                            showError = false
+                        if signatureInputMode == .draw{
```
vivekjyani: Missing space before `{`. Swift convention requires a space before the opening brace:
```swift
if signatureInputMode == .draw {
```

Sources/JoyfillUI/View/Fields/SignatureView.swift
```diff
@@ -302,22 +381,45 @@ struct CanvasSignatureView: View {
...
+                                    savedTypedSignature = ""
+                                    signatureImage = nil
+                                    signatureURL = ""
+                                    showError = false
+                                    presentationMode.wrappedValue.dismiss()
+                                    return
+                                }
+                                savedTypedSignature = trimmedTypedSignatureText
+                                signatureImage = typedSignatureSnapshot(text: trimmedTypedSignatureText)
+                                signatureCanvasImage = signatureImage
```
vivekjyani: `signatureCanvasImage` is a local `@State` used to track the in-progress canvas. Assigning it here is dead — the view is dismissed on the very next lines and `onAppear` will restore it via `signatureCanvasImage = signatureImage` on next open anyway. This line is misleading and should be removed.

vivekjyani (commented):
Overall the feature is well-structured and tests are thorough. A few issues flagged inline — most notably the dead `signatureInputMode` state in `TableSignatureView` and an access-control gap in `CanvasSignatureView`.

Sources/JoyfillUI/View/Fields/SignatureView.swift
vivekjyani: Good catch. I fixed the indentation so the header Text block aligns with sibling views under the same VStack level.

Sources/JoyfillUI/View/Fields/SignatureView.swift
vivekjyani: Fixed. Updated to \if signatureInputMode == .draw {` to match Swift spacing conventions.`

Sources/JoyfillUI/View/Fields/SignatureView.swift
vivekjyani: Updated as suggested: changed to \@State private var signatureInputMode: SignatureInputMode = .draw` for consistency with the rest of the view state.`

Sources/JoyfillUI/View/Fields/TableView/TableSignatureView.swift
vivekjyani: Removed the unused \@State var signatureInputMode` from `TableSignatureView` since it wasn’t wired to `CanvasSignatureView` and had no effect.`

Sources/JoyfillUI/View/Fields/TableView/TableSignatureView.swift
vivekjyani: Removed \signatureInputMode = .draw` from the button action because it was only updating dead local state and was effectively a no-op.`

Sources/JoyfillUI/View/Fields/SignatureView.swift
vivekjyani: Agreed. Removed this assignment since it was redundant right before dismiss, and the canvas image is restored from signatureImag in `onAppear`.

Sources/JoyfillUI/View/Fields/SignatureView.swift
```diff
@@ -339,6 +441,7 @@ struct CanvasSignatureView: View {
         .onAppear {
             signatureCanvasImage = signatureImage
             showCanvasError = showError
+            typedSignatureText = savedTypedSignature
```
vivekjyani, line 405: **UX bug: editing a typed signature always opens in Draw mode.**

onAppear correctly restores typedSignatureText but does not restore signatureInputMode. When a user taps the Edit button on a typed signature, they land on the draw canvas with no indication their typed text exists — they must know to manually switch to the Type tab.

Fix: restore the mode alongside the text:

    typedSignatureText = savedTypedSignature
    signatureInputMode = savedTypedSignature.isEmpty ? .draw : .type

Note: the tests for typed-text restore (testTableRestoreTypedTextOnEdit, testCollectionRestoreTypedTextOnEdit) both call switchToTypeMode() explicitly after tapping Edit, which masks this — they pass without catching the wrong default mode.

Sources/JoyfillUI/View/Fields/SignatureView.swift
```diff
@@ -302,22 +381,45 @@ struct CanvasSignatureView: View {
...
-                            return
-                        }
-                        if !showCanvasError {
-                            signatureImage = CanvasView(lines: $lines, signatureCanvasImage: $signatureCanvasImage, showCanvasError: $showCanvasError)
-                                .frame(width: screenWidth, height: 220)
-                                .snapshot()
-                            showError = false
+                        } else {
+                            let trimmedTypedSignatureText = typedSignatureText.trimmingCharacters(in: .whitespacesAndNewlines)
+                                if trimmedTypedSignatureText.isEmpty {
```
vivekjyani: Inconsistent indentation. Lines 405-413 (the empty-text guard inside the type-mode else branch) are indented 4 extra spaces compared to lines 414-421. Both blocks are at the same logical level (direct children of the else body). Align them uniformly.

Sources/JoyfillUI/View/Fields/SignatureView.swift
```diff
@@ -200,35 +215,98 @@ struct CanvasView: View {
...
+                        .background(
+                            RoundedRectangle(cornerRadius: 10)
+                                .fill(signatureInputMode == .draw ? Color.accentColor : Color.clear)
+                        )
+                        .overlay(
+                            RoundedRectangle(cornerRadius: 10)
+                                .stroke(Color.allFieldBorderColor, lineWidth: 1)
+                        )
+                    })
+                    Spacer().frame(width: 8)
```
vivekjyani: Non-idiomatic spacing. Spacer().frame(width: 8) behaves as a fixed spacer but works against the layout engine. Use the HStack spacing parameter and remove this line:

    HStack(spacing: 8) {
        Button(...) { /* Draw */ }
        Button(...) { /* Type */ }
    }

vivekjyani (commented):
Second-pass review — three additional issues found.

Sources/JoyfillUI/View/Fields/SignatureView.swift
vivekjyani: Makes sense. I removed the fixed Spacer width and switched to stack spacing for the Draw/Type button group.

Sources/JoyfillUI/View/Fields/SignatureView.swift
vivekjyani: Agreed. I aligned the indentation in saveTypedSignature so the empty-text guard and subsequent statements are at the same logical level.

Sources/JoyfillUI/View/Fields/SignatureView.swift
vivekjyani: Thanks for flagging this. This is intentional for our UX: Edit should always open in Draw mode as the default entry point, even when typed text is restored, so users get a consistent initial state every time.

PR #315, opened 2026-04-23
NO-2082: fix: eliminate @Published property reads on background thread (CollectionViewModel + TableViewModel)

vivekjyani:
## 1. Context

Crashlytics reported a recurring crash on a thread named **\"Crashed: Collection\"** in production (`com.servicetrade.mobile` v8.0.1 build 1946). The crash surfaced as `PublishedSubject.value.getter` — Swift's `@Published` uses a `PassthroughSubject` internally whose `.value` accessor is **not thread-safe**. Reading it off the main thread causes an immediate crash.

The root cause: both `CollectionViewModel` and `TableViewModel` dispatch heavy work onto background `DispatchQueue`s. Inside those background blocks, multiple methods were reading `self.tableDataModel` (a `@Published` property) — triggering the crash.

Ticket: NO-2082

---

## 2. What Changed

**`CollectionViewModel.swift`**

- Removed `@Published var showRowSelector` and `@Published var showSingleClickEditButton` stored properties — converted to methods accepting `tableDataModel` as a parameter
- Updated `rowWidth` to call those methods with the passed snapshot instead of stored properties
- Added `tableDataModel: TableDataModel` parameter to the entire initialization call chain:
  - `getparentToChildRowMap(tableDataModel:)`
  - `setupRows(tableDataModel:)`
  - `setupRowsChildrens(tableDataModel:)`
  - `addAllSchemasRecursively(..., tableDataModel:)`
  - `addAllNestedRowsRecursively(..., tableDataModel:)`
  - `shouldShowRowAccToFilters(schemaKey:row:tableDataModel:)`
- Added `tableDataModel: TableDataModel` parameter to the entire `bulkEdit` call chain:
  - `makeChangeDict(..., tableDataModel:)`
  - `updateJSON(..., tableDataModel:)`
  - `updateBulkLocalCellModels(..., tableDataModel:)`
  - `bulkEdit` — captures `let tableDataModel = self.tableDataModel` before dispatch

**`TableViewModel.swift`**

- Added `tableDataModel: TableDataModel` parameter to `makeChangeDict`
- `bulkEdit` — captures `let tableDataModel = self.tableDataModel` before dispatch, passes it through `makeChangeDict` and all reads inside the background block

**`CollectionModelView.swift`**

- Updated all call sites for `showRowSelector`, `showSingleClickEditButton`, and `rowWidth` to pass `viewModel.tableDataModel` explicitly

---

## 3. Decisions & Reasoning

**Why pass `tableDataModel` as a parameter instead of reading `self.tableDataModel`?**
`TableDataModel` is a `struct` (value type). Capturing it as `let tableDataModel = self.tableDataModel` on the main thread before dispatching creates a safe immutable snapshot. All reads inside the background block use that snapshot — no `@Published` accessor is ever touched on a background thread.

**Why remove `showRowSelector` and `showSingleClickEditButton` as `@Published` stored properties?**
Both values were derived entirely from `tableDataModel.mode` and `tableDataModel.singleClickRowEdit` — they should never have been separate stored properties. Having them stored separately created stale-state risk (stored bool could diverge from `tableDataModel`) and was a direct trigger for the background thread crash. Converting them to methods with a `tableDataModel` parameter enforces a single source of truth.

**Why not use optional defaults like `tableDataModel: TableDataModel? = nil`?**
That would hide missing call-site updates behind a silent fallback. Requiring the parameter explicitly forces every caller to be intentional and auditable.

---

## 4. Screenshot / Video

This fix addresses a background-thread crash with no UI behavior change — there is no visual difference. The fix eliminates the `PublishedSubject.value.getter` crash appearing in Crashlytics on the "Collection" thread.

---

## 5. Public API Changes / Docs

No public API changes. All modified methods are `internal`, `private`, or `fileprivate`. No documentation updates needed.

---

## 6. Test Coverage

No automated tests added in this PR. The crash is a threading race condition triggered in production; reproducing it deterministically requires thread-sanitizer instrumentation. Manual verification: Collection and Table views load correctly across all field modes (fill, readonly) and row decorator configurations.

---

## 7. Notes for Reviewer

- Commits are intentionally **small and scoped** — one fix per method/chain — to make each change easy to review in isolation
- There are **no semantic behavior changes** — only threading safety improvements via parameter threading
- The `@Published` properties `showRowSelector` and `showSingleClickEditButton` removed from `CollectionViewModel` were never directly observed by SwiftUI views — re-renders continue to be driven by the existing `@Published var tableDataModel`
- Wherever you see `let tableDataModel = self.tableDataModel` before a `dispatchQueue.async` block — that is the intentional snapshot capture pattern, not a mistake
- The `DispatchQueue.main.sync` inside `updateJSON` is called from a background thread so there is no deadlock risk — it is the correct write-back mechanism for updating `@Published` state after the background computation completes

### DIFF
--- a/Sources/JoyfillUI/View/Fields/CollectionView/CollectionModelView.swift
+++ b/Sources/JoyfillUI/View/Fields/CollectionView/CollectionModelView.swift
@@ -146,7 +146,7 @@ struct CollectionModalView : View {
             if viewModel.nestedTableCount > 0 {
                 Spacer()
             }
-            if viewModel.showRowSelector  {
+            if viewModel.showRowSelector(for: viewModel.tableDataModel)  {
                 Image(systemName: viewModel.tableDataModel.allRowSelected ? "circle.square.fill" : "square")
                     .frame(width: 40, height: textHeight)
                     .foregroundColor(viewModel.tableDataModel.filteredcellModels.count == 0 ? Color.gray.opacity(0.4) : nil)
@@ -165,7 +165,7 @@ struct CollectionModalView : View {
                 .frame(width: 40, height: 60)
                 .border(Color.tableCellBorderColor)
             
-            if viewModel.showSingleClickEditButton {
+            if viewModel.showSingleClickEditButton(for: viewModel.tableDataModel) {
                 Image(systemName: "square.and.pencil")
                     .frame(width: 40, height: 60)
                     .foregroundColor(Color.gray.opacity(0.4))
@@ -180,9 +180,9 @@ struct CollectionModalView : View {
 
     private var collectionLeftColumnWidth: CGFloat {
         var width: CGFloat = 40 // # column
-        if viewModel.showRowSelector { width += 40 }
+        if viewModel.showRowSelector(for: viewModel.tableDataModel) { width += 40 }
         if viewModel.nestedTableCount > 0 { width += 40 }
-        if viewModel.showSingleClickEditButton { width += 40 }
+        if viewModel.showSingleClickEditButton(for: viewModel.tableDataModel) { width += 40 }
         return width
     }
 
@@ -377,7 +377,7 @@ struct RootTitleRowView: View {
         .padding(.horizontal, 8)
         .padding(.vertical, 4)
         .frame(minHeight: 50)
-        .frame(width: viewModel.rowWidth(viewModel.tableDataModel.tableColumns, 0, viewModel.rootSchemaKey), height: 60)
+        .frame(width: viewModel.rowWidth(viewModel.tableDataModel.tableColumns, 0, viewModel.rootSchemaKey, tableDataModel: viewModel.tableDataModel), height: 60)
         .font(.system(size: 15, weight: .bold))
         .border(Color.tableCellBorderColor)
         .background(colorScheme == .dark ? Color(UIColor.systemGray6) : Color.tableColumnBgColor)
@@ -538,7 +538,7 @@ struct CollectionRowsHeaderView: View {
             // Selector Button View
             switch rowModel.rowType {
             case .row(let index):
-                if viewModel.showRowSelector {
+                if viewModel.showRowSelector(for: viewModel.tableDataModel) {
                     Image(systemName: isRowSelected ? "record.circle.fill" : "circle")
                         .frame(width: 40, height: 60)
                         .background(Color.rowSelectionBackground(isSelected: isRowSelected, colorScheme: colorScheme))
@@ -549,7 +549,7 @@ struct CollectionRowsHeaderView: View {
                         .accessibilityIdentifier("selectRowItem\(index)")
                 }
             case .header:
-                if viewModel.showRowSelector {
+                if viewModel.showRowSelector(for: viewModel.tableDataModel) {
                     Image(systemName: viewModel.tableDataModel.allNestedRowSelected(rowID: rowModel.rowID) ? "circle.square.fill" : "square")
                         .frame(width: 40, height: 60)
                         .foregroundColor(viewModel.tableDataModel.getAllNestedRowsForRow(rowID: rowModel.rowID).count == 0 ? Color.gray.opacity(0.4) : nil)
@@ -565,7 +565,7 @@ struct CollectionRowsHeaderView: View {
                         .accessibilityIdentifier("selectAllNestedRows")
                 }
             case .nestedRow(let level, let index, _, _):
-                if viewModel.showRowSelector {
+                if viewModel.showRowSelector(for: viewModel.tableDataModel) {
                     Image(systemName: isRowSelected ? "record.circle.fill" : "circle")
                         .frame(width: 40, height: 60)
                         .background(Color.rowSelectionBackground(isSelected: isRowSelected, colorScheme: colorScheme))
@@ -586,7 +586,7 @@ struct CollectionRowsHeaderView: View {
                     .frame(width: 40, height: 60)
                     .background(colorScheme == .dark ? Color(UIColor.systemGray6) : Color.tableColumnBgColor)
                     .border(Color.tableCellBorderColor)
-                if viewModel.showSingleClickEditButton {
+                if viewModel.showSingleClickEditButton(for: viewModel.tableDataModel) {
                     Image(systemName: "square.and.pencil")
                         .frame(width: 40, height: 60)
                         .foregroundColor(Color.gray.opacity(0.4))
@@ -609,7 +609,7 @@ struct CollectionRowsHeaderView: View {
                         .background(Color.rowSelectionBackground(isSelected: isRowSelected, colorScheme: colorScheme))
                         .border(Color.tableCellBorderColor)
                 }
-                if viewModel.showSingleClickEditButton {
+                if viewModel.showSingleClickEditButton(for: viewModel.tableDataModel) {
                     Image(systemName: "square.and.pencil")
                         .foregroundColor(.blue)
                         .frame(width: 40, height: 60)
@@ -652,7 +652,7 @@ struct CollectionRowsHeaderView: View {
                         .background(Color.rowSelectionBackground(isSelected: isRowSelected, colorScheme: colorScheme))
                         .border(Color.tableCellBorderColor)
                 }
-                if viewModel.showSingleClickEditButton {
+                if viewModel.showSingleClickEditButton(for: viewModel.tableDataModel) {
                     Image(systemName: "square.and.pencil")
                         .foregroundColor(.blue)
                         .frame(width: 40, height: 60)
--- a/Sources/JoyfillUI/View/Fields/CollectionView/CollectionModelView.swift
+++ b/Sources/JoyfillUI/View/Fields/CollectionView/CollectionModelView.swift
@@ -146,7 +146,7 @@ struct CollectionModalView : View {
             if viewModel.nestedTableCount > 0 {
                 Spacer()
             }
-            if viewModel.showRowSelector(for: viewModel.tableDataModel)  {
+            if viewModel.showRowSelector(for: viewModel.tableDataModel) {
                 Image(systemName: viewModel.tableDataModel.allRowSelected ? "circle.square.fill" : "square")
                     .frame(width: 40, height: textHeight)
                     .foregroundColor(viewModel.tableDataModel.filteredcellModels.count == 0 ? Color.gray.opacity(0.4) : nil)
--- a/Sources/JoyfillUI/View/Fields/CollectionView/CollectionViewModel.swift
+++ b/Sources/JoyfillUI/View/Fields/CollectionView/CollectionViewModel.swift
@@ -560,8 +560,8 @@ class CollectionViewModel: ObservableObject, TableDataViewModelProtocol {
     
     func getCellModels(tableDataModel: TableDataModel) -> [RowDataModel] {
         var cellModels = [RowDataModel]()
-        let rowDataMap = setupRows()
-        let rowToChildrenMap = setupRowsChildrens()
+        let rowDataMap = setupRows(tableDataModel: tableDataModel)
+        let rowToChildrenMap = setupRowsChildrens(tableDataModel: tableDataModel)
         tableDataModel.valueToValueElements?.forEach { valueElement in
             if valueElement.deleted ?? false { return }
             guard let rowID = valueElement.id else {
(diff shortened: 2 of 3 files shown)


Sources/JoyfillUI/View/Fields/CollectionView/CollectionViewModel.swift
```diff
@@ -13,7 +13,6 @@ class CollectionViewModel: ObservableObject, TableDataViewModelProtocol {
     @Published var tableDataModel: TableDataModel
```
vivekjyani, line 13: **No compile-time enforcement against future regressions**

The fix works at runtime, but nothing prevents a future developer from re-introducing a `self.tableDataModel` read inside the background closure chain. A more durable fix would be to annotate `CollectionViewModel` with `@MainActor` and mark the background-safe methods `nonisolated`, so the compiler rejects any future off-main `@Published` reads. Without this, the threading contract is enforced only by convention and code review.

Sources/JoyfillUI/View/Fields/CollectionView/CollectionViewModel.swift
```diff
@@ -855,7 +859,7 @@ class CollectionViewModel: ObservableObject, TableDataViewModelProtocol {
                                                cells: [],
                                                rowType: .header(level: level + 1,
```
vivekjyani, line 860: **Pre-existing data race: `rowToValueElementMap` read on background thread**

`rowToValueElementMap` is a regular `var` (not actor-isolated, not thread-safe). It is read here on the background `dispatchQueue` while writes likely occur on the main thread (e.g. `addRow`). This is a pre-existing race that this PR didn't introduce, but since this PR is auditing the threading story in this class, it's worth tracking as a follow-up. A `DispatchQueue`-protected accessor or capturing a snapshot before dispatch would eliminate it.

Sources/JoyfillUI/View/Fields/CollectionView/CollectionViewModel.swift
```diff
@@ -589,15 +591,15 @@ class CollectionViewModel: ObservableObject, TableDataViewModelProtocol {
                     rowCellModels.append(cellModel)
                 }
             }
-            cellModels.append(RowDataModel(rowID: rowID, cells: rowCellModels, rowType: .row(index: cellModels.count + 1), rowWidth: rowWidth(tableDataModel.tableColumns, 0, rootSchemaKey)))
+            cellModels.append(RowDataModel(rowID: rowID, cells: rowCellModels, rowType: .row(index: cellModels.count + 1), rowWidth: rowWidth(tableDataModel.tableColumns, 0, rootSchemaKey, tableDataModel: tableDataModel)))
         }
         return cellModels
```
vivekjyani, line 596: **Minor: unlabeled positional parameters reduce readability**

Both parameters are positional (`_`), making call sites like `getAllCellModels(tableDataModel, targetSchema)` harder to parse at a glance. Since this is `fileprivate`, adding a label for `tableDataModel` costs nothing externally:

```swift
fileprivate func getAllCellModels(_ targetSchema: String, tableDataModel: TableDataModel) -> [RowDataModel]
```

This also keeps the original parameter first so callers don't need to reorder arguments.

Sources/JoyfillUI/View/Fields/CollectionView/CollectionModelView.swift
```diff
@@ -146,7 +146,7 @@ struct CollectionModalView : View {
             if viewModel.nestedTableCount > 0 {
```
vivekjyani, line 146: **Nit: trailing double-space before `{`**

Pre-existing but carried forward in this change — two spaces before the brace. Worth cleaning up since this line was touched.

Sources/JoyfillUI/View/Fields/CollectionView/CollectionViewModel.swift
vivekjyani: Agreed in principle — compile-time enforcement is stronger than convention. But migrating CollectionViewModel to @MainActor + nonisolated is a much larger change than this PR: it cascades through every caller, requires auditing every helper method on self for isolation, and would force the DispatchQueue.main.sync write-back pattern in updateJSON to migrate to structured concurrency. This PR is a targeted crash hotfix — I'd rather ship it and open a follow-up ticket for the concurrency migration so that can be reviewed as its own scoped change.

Sources/JoyfillUI/View/Fields/CollectionView/CollectionViewModel.swift
vivekjyani: Fair point, but worth distinguishing — the @Published crash was deterministic (PassthroughSubject.value.getter is never thread-safe). rowToValueElementMap is a plain var Dictionary, which is only unsafe under concurrent read+write. There's no Crashlytics signature pointing to it, so it's a theoretical race, not an observed one. I'll note it as a follow-up, but it doesn't rise to hotfix severity.

Sources/JoyfillUI/View/Fields/CollectionView/CollectionModelView.swift
vivekjyani: Fixed

Sources/JoyfillUI/View/Fields/CollectionView/CollectionViewModel.swift
vivekjyani: Fixed

Sources/JoyfillUI/View/Fields/CollectionView/CollectionViewModel.swift
```diff
@@ -1790,35 +1794,36 @@ class CollectionViewModel: ObservableObject, TableDataViewModelProtocol {
         if changes.count == 0 { return }
         isBulkLoading = true
         let tableColumns = self.getTableColumnsForSelectedRows()
+        let tableDataModel = self.tableDataModel
```
VishnuBishnoi, line 1798: The snapshot-capture is correct here. Worth adding a brief comment so future developers understand why this local shadows the property:

```swift
// Capture an immutable value-type snapshot on the main thread.
// All methods called inside the background block receive this as an
// explicit parameter so they never touch self.tableDataModel (@Published)
// off the main thread — the source of the "Crashed: Collection" crash.
let tableDataModel = self.tableDataModel
```

Without this, someone adding a call inside the `dispatchQueue.async` block might write `self.tableDataModel` expecting it to be the current value, not realizing the snapshot was intentional.

Sources/JoyfillUI/View/Fields/CollectionView/CollectionViewModel.swift
```diff
@@ -1780,7 +1784,7 @@ class CollectionViewModel: ObservableObject, TableDataViewModelProtocol {
                 updatedCellModels[rowIndex] = rowDataModel
                 
                 //Update conditional logic
-                self.tableDataModel.documentEditor?.updateSchemaVisibilityOnCellChange(collectionFieldID: self.tableDataModel.fieldIdentifier.fieldID, columnID: cellDataModel.id, rowID: rowId, valueElement: self.rowToValueElementMap[rowId])
+                tableDataModel.documentEditor?.updateSchemaVisibilityOnCellChange(collectionFieldID: tableDataModel.fieldIdentifier.fieldID, columnID: cellDataModel.id, rowID: rowId, valueElement: self.rowToValueElementMap[rowId])
```
VishnuBishnoi, line 1787: Two pre-existing background-thread access concerns on this line (not introduced by this PR, but visible now that the snapshot pattern is in place):

1. **`tableDataModel.documentEditor`** — `documentEditor` is a reference-type (`DocumentEditor`) stored inside the value-type snapshot. `tableDataModel.documentEditor` returns the *same class instance* as `self.tableDataModel.documentEditor`. Calling `updateSchemaVisibilityOnCellChange()` on it from `dispatchQueue` is calling a method on a reference type off the main thread. If `DocumentEditor`'s conditional-logic methods touch any `@Published` state, this is another race. Recommend either moving this call behind `DispatchQueue.main.async {}` or confirming the method is thread-safe.

2. **`self.rowToValueElementMap`** — this plain `var [String: ValueElement]` is read here on the background thread while writes happen on the main thread (e.g., `addRow`). No Crashlytics signature yet, but this is a genuine data race the thread sanitizer would flag. Track alongside the `documentEditor` concern for a follow-up PR.

VishnuBishnoi (approved):
The snapshot-capture pattern is the right approach for this class of threading crash. Capturing `let tableDataModel = self.tableDataModel` on the main thread before dispatch, then threading it as an explicit parameter through every downstream method, eliminates all `@Published` accessor reads from the background queue — which is exactly where the `PublishedSubject.value.getter` crash originated.

---

**What the PR gets right**

- Value-type semantics: `TableDataModel` is a struct, so each background block gets its own immutable snapshot — no shared mutable state.
- Systematic coverage: every affected method in the dispatch chain received the `tableDataModel:` parameter. The diff touches `setupRows`, `setupRowsChildrens`, `getAllCellModels`, `addAllSchemasRecursively`, `addAllNestedRowsRecursively`, `shouldShowRowAccToFilters`, `makeChangeDict`, `updateJSON`, `updateBulkLocalCellModels`, and `rowWidth` — all converted consistently.
- Removed stored `@Published var showRowSelector` in favour of a pure function. Eliminates one more property that was set on main but read off-main through the row-width chain.
- Write-backs (`self.tableDataModel.filteredcellModels = updatedCellModels`) happen after the `await` resumes on the original caller context — main thread. Correct.

**The elegant aspect of the parameter naming**

By naming the parameter identically to the property (`tableDataModel: TableDataModel`), it shadows `self.tableDataModel` throughout the entire function body without requiring any body-line changes. This is why many method bodies in the diff show no `+/-` on their reads — they silently switched from the property to the parameter. This is correct Swift behaviour but may be surprising to future reviewers: a comment at the capture site explains it (see inline).

---

**Two pre-existing concerns worth tracking as follow-ups**

See inline comments. Neither is introduced by this PR but both are now more visible now that `tableDataModel` reads are accounted for.

**@MainActor migration (previously discussed):** Agreed the full migration is out of scope for this hotfix. A scoped follow-up PR that annotates the two ViewModels with `@MainActor` and marks only the background-safe pure-computation functions `nonisolated` would give compiler-enforced prevention of future regressions.

Sources/JoyfillUI/View/Fields/CollectionView/CollectionViewModel.swift
vivekjyani: Done, added a one-line comment explaining the snapshot intent.

Sources/JoyfillUI/View/Fields/CollectionView/CollectionViewModel.swift
vivekjyani: Acknowledged — both are pre-existing and not introduced by this PR. I'll track them in a follow-up: (1) audit DocumentEditor for thread-safety or move that call to DispatchQueue.main.async, and (2) protect rowToValueElementMap with a lock or capture a snapshot before the background block.

VishnuBishnoi (approved):
All threading concerns are addressed.

**`showRowSelector` / `showSingleClickEditButton`** ✅ — Correctly converted from `@Published` stored properties to methods that accept an explicit `tableDataModel: TableDataModel` parameter. The `@Published` property was the root race: background threads calling `showRowSelector` were reading `self.tableDataModel` off the main thread. Passing the local snapshot eliminates that access path entirely.

**Full call-chain threading** ✅ — The snapshot propagates correctly through every layer:
- `getparentToChildRowMap(tableDataModel:)`
- `setupRows(tableDataModel:)` / `setupRowsChildrens(tableDataModel:)`
- `rowWidth(_:_:_:tableDataModel:)`
- `shouldShowRowAccToFilters(schemaKey:row:tableDataModel:)`
- `addAllNestedRowsRecursively(...tableDataModel:)`
- `getAllCellModels(_:tableDataModel:)`
- `makeChangeDict(_:_:_:rowIndexMap:tableDataModel:)` (both Collection and Table)

**Bulk-edit snapshot** ✅ — `let tableDataModel = self.tableDataModel` is captured on the calling (main) thread before `dispatchQueue.async` in both `CollectionViewModel.bulkEdit` and `TableViewModel.bulkEdit`. Subsequent mutations go through `self.tableDataModel.filteredcellModels = updatedCellModels` on main (after the continuation), which is correct.

**Commit history note (non-blocking):** The revert+redo pair (`Revert "fix: pass tableDataModel snapshot..."` × 2, then re-applied) makes the log slightly noisy. The final state is correct; only worth noting if squashing before merge is on the table.

**Pre-existing concern (unchanged, tracked separately):** `tableDataModel.documentEditor?.updateSchemaVisibilityOnCellChange(...)` is still called from background in `updateBulkLocalCellModels`. `DocumentEditor` is a reference type; this is a pre-existing thread-safety gap that predates this PR.

LGTM — approving.

PR #316, opened 2026-04-24
NO-2075: Decorator API — row/cell scopes, decorate flag auto-toggle, collection support

vivekjyani:
## 1. Context

Decorator API lets callers attach visual indicators (icon/label/color) to fields, rows, columns, and cells via a path string, with a one-shot `decorate` flag on the field/schema that tells the renderer whether to reserve the decorator column. Covers both table and collection fields; collection uses the schema tree (`schemas/{sk}/...`) to address nested scopes.

A new `DecoratorConfig` lets hosts cap how many decorators render inline before the rest collapse into a kebab menu — separate limits for fields/columns vs. rows.

## 2. What changed

- **Public API surface** (`DocumentEditor+Decorators.swift`): `getDecorators`, `addDecorators`, `removeDecorator`, `updateDecorator`, all keyed by a `page-id/field-position-id/...` path.
- **Path grammar**: `/rows`, `/columns/{col}`, `/{rowID}`, `/{rowID}/{col}`, `/{rowID}/columns/{col}`, `/schemas/{sk}/...` (collection only). `rowScopedColumn` aliases `cell` storage.
- **Model** (`JoyfillModel`): `Decorator`, `Decorators` (all/cells), `rowDecorators` on field + schema entry, per-column `decorators`, `decorate` flag on field + schema entry.
- **Rendering**: table + collection read `decorate` to decide whether to render the decorator column; per-row decorators fall back to the common `rowDecorators` list when a row has none.
- **decorate flag auto-toggle**: add flips it on (`ensureDecorateEnabled`); remove/update flips it off when the scope has no displayable decorators left (`ensureDecorateDisabledIfEmpty` — new, this PR's bug fix).
- **Single-commit write path**: `applyDecorators` / `ensureDecorateEnabled` / `ensureDecorateDisabledIfEmpty` now take `inout JoyDocField` and mutate without persisting; the three public APIs commit once at the end via `commitDecoratorChange`.
- **DecoratorConfig + kebab overflow UI**: new `DecoratorConfig(visibleLimitInFields: Int = 2, visibleLimitInRows: Int = 1)` passed at editor init. When a field/column scope exceeds `visibleLimitInFields` (default 2) or a row exceeds `visibleLimitInRows` (default 1), only a kebab icon renders; tap shows the full set in a popover. Field/column kebabs get a light blue background to match `DecoratorButton` styling; row kebabs render plain. Negative values clamp to 0.

## 3. Decisions

- **Path grammar over nested structs**: keeps the SDK call site a single string argument, matches how host apps already identify fields by page + position, and lets us extend to new terminators without API changes.
- **Copy-on-write seed for specific scopes**: when a row-self/cell scope is first written to, we seed from the matching common scope so the specific scope can diverge without silently losing previously-visible decorators.
- **`decorate` as a derived flag, not caller-managed**: hosts shouldn't need to track how many decorators they've added — if anything displayable exists in scope, the flag is on; otherwise off. `isDisplayable = hasIcon || hasLabel`, so updates that strip both icon and label also flip the flag off.
- **Single commit per API call**: previously `add`/`remove`/`update` could emit two `updateField` writes, two `refreshField` rebuilds, two `decoratorsDidChange` notifications, and two `onChange` events per user action. Collapsed to one. See the last commit for the refactor.
- **Kebab-only over inline + kebab combo**: when over the limit we show only the kebab (not "first N inline plus a kebab for the rest") to keep header/row layout predictable across field counts. The popover then shows everything.
- **Defaults of 2 / 1**: matches the existing visual density — a field title row comfortably fits two decorator chips, a table row fits one before crowding the cell content. Hosts that want everything inline can pass `Int.max`; hosts that want everything in the kebab can pass `0`.
- **Old "one-way latch" tests removed**: two tests in `DecoratorDeepPathTests.swift` asserted decorate stays `true` after the last decorator is removed — that encoded the bug being fixed. Replaced with inverted assertions in `DecoratorPublicAPITests.swift`.

## 4. Screenshot / video

N/A — pure API + behavioral change. User-facing effects:
- The decorator column disappears when all decorators are removed (previously it stayed visible but empty).
- A kebab menu replaces the inline chips when a scope exceeds its configured limit.

## 5. Public API / docs

- **New public API** on `DocumentEditor`: `getDecorators(path:)`, `addDecorators(path:decorators:)`, `removeDecorator(path:action:)`, `updateDecorator(path:action:decorator:)`.
- **New public model types**: `Decorator`, `Decorators`, `JoyfillError.decoratorError`.
- **New public config type**: `DecoratorConfig(visibleLimitInFields:visibleLimitInRows:)`, exposed on `DocumentEditor` as a read-only `decoratorConfig` property; passed at init via `DocumentEditor(... decoratorConfig:)`.
- **Public docs**: path grammar + storage map are documented inline at the top of `DocumentEditor+Decorators.swift`. External SDK docs will need entries for the decorator path grammar, the auto-toggle contract, and the new `DecoratorConfig` limits — **docs update required** before release.

## 6. Tests

Covered in this branch — no follow-up test work pending.

- `DecoratorPublicAPITests` — add/remove/update happy paths, duplicate-action rejection, path resolution, field-map reflection, **and the 14 new decorate-flag cases** (table, collection root, nested schema, scope isolation, updateDecorator → non-displayable).
- `DecoratorDeepPathTests` — long paths, schema descent, multi-level nesting.
- `DecoratorCOWTests` — specific-scope seeding from common scope.
- `DecoratorLiveUpdateTests` — post-mutation field state.
- `DecoratorPathResolutionTests` — path grammar edge cases.
- `DecoratorErrorHandlingTests` — invalid paths, license gating, validation errors.
- `DecoratorAPIDemoView` (example app) — UI sandbox with steppers for `visibleLimitInFields` / `visibleLimitInRows` and an Update button that rebuilds the editor with the new config while preserving the current document state.

## 7. Notes for reviewer

- All four public APIs are main-thread-only — call out in docs.
- Collection-field writes are license-gated; without a valid license the editor emits `decoratorError` and rejects the write. Same gating lives on `licenseAllowsDecoratorWrite`.
- Helpers `applyDecorators` / `ensureDecorateEnabled` / `ensureDecorateDisabledIfEmpty` intentionally take `inout JoyDocField` and do **not** commit — only the top-level public methods own the persist + notify path. If you add a new decorator mutation, follow the same pattern: mutate in place, then one `commitDecoratorChange` at the end.
- `DecoratorConfig` is currently set once at editor init; live reconfiguration requires rebuilding the `DocumentEditor` (the demo's Update button shows the pattern — reuses `editor.document` so user edits survive). If we need live reconfig later, the config can be promoted to a settable property and a `decoratorsDidChange`-style refresh added.
- `DecoratorDeepPathTests.swift` has a breadcrumb comment where the two obsolete "one-way latch" tests used to live, pointing to their replacements.

### DIFF
--- a/Sources/JoyfillUI/ViewModels/DocumentEditor+Decorators.swift
+++ b/Sources/JoyfillUI/ViewModels/DocumentEditor+Decorators.swift
@@ -204,6 +204,10 @@ private extension DocumentEditor {
         case .rowSelf:
             schemaKey = target.hops.last?.schemaKey
         default:
+            // .field / .commonColumn / .cell / .rowScopedColumn — none reserve
+            // the row-decorator column. Cell and row-scoped-column decorators
+            // render within the cell body (next to the column title), so they
+            // intentionally do not flip `decorate` on.
             return
         }
 
@@ -242,6 +246,9 @@ private extension DocumentEditor {
         case .rowSelf:
             schemaKey = target.hops.last?.schemaKey
         default:
+            // .field / .commonColumn / .cell / .rowScopedColumn — none drive
+            // the row-decorator column, so removing one never flips `decorate`
+            // off. Cell decorators in particular render within the cell body.
             return
         }
 
@@ -566,7 +573,7 @@ private extension DocumentEditor {
                   let children = row.childrens?[sk] else { return nil }
             currentRows = children.valueToValueElements ?? []
         }
-        return nil
+        return nil // unreachable: loop always returns; required to satisfy the compiler.
     }
 
     /// Recursively rewrites the row tree to apply the terminator's new decorator list at the deepest hop.
--- a/Sources/JoyfillUI/ViewModels/DocumentEditor+Decorators.swift
+++ b/Sources/JoyfillUI/ViewModels/DocumentEditor+Decorators.swift
@@ -4,13 +4,16 @@ public extension DocumentEditor {
 
     // MARK: - Path-Based API
     //
-    // path = "pageId/fieldPositionId"              → field decorators
-    // path = "pageId/fieldPositionId/rows"         → common row decorators   (table: field.rowDecorators)
-    // path = "pageId/fieldPositionId/{rowId}"      → row-specific decorators (table: ValueElement.decorators.all)
-    // path = "pageId/fieldPositionId/{rowId}/colId" → column decorators
+    // path = "pageId/fieldPositionId"                    → field decorators
+    // path = "pageId/fieldPositionId/rows"               → common row decorators    (table: field.rowDecorators)
+    // path = "pageId/fieldPositionId/{rowId}"            → row-specific decorators  (table: ValueElement.decorators.all)
+    // path = "pageId/fieldPositionId/columns/{colId}"    → common column decorators (table: tableColumn.decorators)
+    // path = "pageId/fieldPositionId/{rowId}/{colId}"    → cell-specific decorators (table: ValueElement.decorators.cells[colId])
     //
     // Collection fields: the 3rd segment is always a real rowId (used to resolve the schemaKey).
-    // Table fields: "rows" keyword → common; any valid rowId → row-specific (with copy-on-write from common).
+    //   "rows" / "columns" keywords are table-only for now.
+    // Table fields: "rows" keyword → common rows; "columns" keyword → common columns;
+    //   any valid rowId → row-specific / cell-specific (with copy-on-write from common).
     //
     // Note: Path segments are parsed with empty-component filtering;
     // consecutive or trailing slashes are treated as single separators.
@@ -43,11 +46,14 @@ public extension DocumentEditor {
             }
         }
         var list = fetchDecorators(for: target)
-        // Copy-on-write: if adding to a row-specific target with no existing row-specific
-        // decorators, seed the list with the common row decorators first so that row
-        // diverges from common while keeping all previously-visible decorators.
+        // Copy-on-write: if adding to a row-specific / cell-specific target with no existing
+        // decorators on that scope, seed the list with the matching common decorators first so
+        // the row/cell diverges from common while keeping all previously-visible decorators.
         if case .rowSpecific(let fID, _) = target, list.isEmpty {
             list = rowDecoratorsForPath(fieldID: fID, schemaKey: nil)
+        }
+        if case .cellSpecific(let fID, _, let colID) = target, list.isEmpty {
+            list = columnDecoratorsForPath(fieldID: fID, columnID: colID, schemaKey: nil)
         }
         let existingActions = Set(list.compactMap { $0.action })
         let newActions = decorators.compactMap { $0.action }
@@ -132,6 +138,7 @@ private extension DocumentEditor {
         case .row(let fieldID, _): return fieldID
         case .rowSpecific(let fieldID, _): return fieldID
         case .column(let fieldID, _, _): return fieldID
+        case .cellSpecific(let fieldID, _, _): return fieldID
         }
     }
 
@@ -176,7 +183,7 @@ private extension DocumentEditor {
             field.decorate = true
             commitDecoratorChange(field: field, fieldID: fieldID)
 
-        case .field, .column:
+        case .field, .column, .cellSpecific:
             break
         }
     }
@@ -268,6 +275,7 @@ private extension DocumentEditor {
         case row(fieldID: String, schemaKey: String?)          // common row decorators
         case rowSpecific(fieldID: String, rowID: String)       // row-specific decorators (table only)
         case column(fieldID: String, columnID: String, schemaKey: String?)
+        case cellSpecific(fieldID: String, rowID: String, columnID: String) // cell-specific decorators (table only)
     }
 
     /// Parses the path and maps it to the correct decorator scope.
@@ -284,17 +292,33 @@ private extension DocumentEditor {
               let field = fieldMap[fieldID] else { return nil }
 
         if let columnId = parsed.columnId {
-            guard let rowId = parsed.rowId else { return nil }
-            guard rowExistsInField(fieldID: fieldID, rowId: rowId) else { return nil }
-            // 4 segments → column decorators; rowId resolves the schema for collection fields
-            let schemaKey = resolvedSchemaKey(forRowID: rowId, inFieldID: fieldID)
-            guard columnExistsInField(field, columnId: columnId, schemaKey: schemaKey) else { return nil }
-            return .column(fieldID: fieldID, columnID: columnId, schemaKey: schemaKey)
+            guard let rowSegment = parsed.rowId else { return nil }
+            if field.fieldType == .table {
+                if rowSegment == "columns" {
+                    // "columns/{colId}" → common column decorators
+                    guard columnExistsInField(field, columnId: columnId, schemaKey: nil) else { return nil }
+                    return .column(fieldID: fieldID, columnID: columnId, schemaKey: nil)
+                }
+                // "{rowId}/{colId}" → cell-specific decorators
+                guard rowExistsInField(fieldID: fieldID, rowId: rowSegment) else { return nil }
+                guard columnExistsInField(field, columnId: columnId, schemaKey: nil) else { return nil }
+                return .cellSpecific(fieldID: fieldID, rowID: rowSegment, columnID: columnId)
+            } else {
+                // Collection: rowId resolves the schema for the column
+                guard rowExistsInField(fieldID: fieldID, rowId: rowSegment) else { return nil }
+                let schemaKey = resolvedSchemaKey(forRowID: rowSegment, inFieldID: fieldID)
+                guard columnExistsInField(field, columnId: columnId, schemaKey: schemaKey) else { return nil }
+                return .column(fieldID: fieldID, columnID: columnId, schemaKey: schemaKey)
+            }
         } else if let rowSegment = parsed.rowId {
             if field.fieldType == .table {
                 if rowSegment == "rows" {
                     // "rows" keyword → common row decorators on the field
                     return .row(fieldID: fieldID, schemaKey: nil)
+                }
+                if rowSegment == "columns" {
+                    // "columns" without a columnID is an invalid path — decorators live per column
+                    return nil
                 }
                 // Actual rowId → row-specific decorators
                 guard rowExistsInField(fieldID: fieldID, rowId: rowSegment) else { return nil }
@@ -382,6 +406,26 @@ private extension DocumentEditor {
         commitDecoratorChange(field: field, fieldID: fieldID)
     }
 
+    // MARK: Cell-specific decorator read/write (table only)
+
+    func cellSpecificDecoratorsForPath(fieldID: String, rowID: String, columnID: String) -> [Decorator] {
+        guard let field = fieldMap[fieldID] else { return [] }
+        return field.valueToValueElements?.first(where: { $0.id == rowID })?.decorators?.cells[columnID] ?? []
+    }
+
+    func setCellSpecificDecoratorsForPath(_ decorators: [Decorator], fieldID: String, rowID: String, columnID: String) {
+        guard var field = fieldMap[fieldID],
+              var rows = field.valueToValueElements,
+              let index = rows.firstIndex(where: { $0.id == rowID }) else { return }
+        var decs = rows[index].decorators ?? Decorators()
+        var cells = decs.cells
+        cells[columnID] = decorators
+        decs.cells = cells
+        rows[index].decorators = decs
+        field.value = ValueUnion.valueElementArray(rows)
+        commitDecoratorChange(field: field, fieldID: fieldID)
+    }
+
     // MARK: Row decorator read/write (path-based, independent of the public API)
 
     /// Reads row decorators directly from the field model.
@@ -425,6 +469,8 @@ private extension DocumentEditor {
             return rowSpecificDecoratorsForPath(fieldID: fieldID, rowID: rowID)
         case .column(let fieldID, let columnID, let schemaKey):
             return columnDecoratorsForPath(fieldID: fieldID, columnID: columnID, schemaKey: schemaKey)
+        case .cellSpecific(let fieldID, let rowID, let columnID):
+            return cellSpecificDecoratorsForPath(fieldID: fieldID, rowID: rowID, columnID: columnID)
         }
     }
 
@@ -439,6 +485,8 @@ private extension DocumentEditor {
             setRowSpecificDecoratorsForPath(decorators, fieldID: fieldID, rowID: rowID)
         case .column(let fieldID, let columnID, let schemaKey):
             setColumnDecoratorsForPath(decorators, fieldID: fieldID, columnID: columnID, schemaKey: schemaKey)
+        case .cellSpecific(let fieldID, let rowID, let columnID):
+            setCellSpecificDecoratorsForPath(decorators, fieldID: fieldID, rowID: rowID, columnID: columnID)
         }
     }
 }
--- a/Sources/JoyfillUI/ViewModels/DocumentEditor+Decorators.swift
+++ b/Sources/JoyfillUI/ViewModels/DocumentEditor+Decorators.swift
@@ -81,8 +81,11 @@ public extension DocumentEditor {
         }
 
         list.append(contentsOf: decorators)
-        ensureDecorateEnabled(for: target)
-        applyDecorators(list, for: target)
+
+        guard var field = fieldMap[target.fieldID] else { return }
+        applyDecorators(list, for: target, in: &field)
+        ensureDecorateEnabled(for: target, in: &field)
+        commitDecoratorChange(field: field, fieldID: target.fieldID)
     }
 
     /// Removes the decorator whose `action` matches at the given path.
@@ -98,7 +101,11 @@ public extension DocumentEditor {
             return
         }
         list.remove(at: index)
-        applyDecorators(list, for: target)
+
+        guard var field = fieldMap[target.fieldID] else { return }
+        applyDecorators(list, for: target, in: &field)
+        ensureDecorateDisabledIfEmpty(for: target, in: &field)
+        commitDecoratorChange(field: field, fieldID: target.fieldID)
     }
 
     /// Replaces the decorator whose `action` matches with the new decorator at the given path.
@@ -126,7 +133,11 @@ public extension DocumentEditor {
             return
         }
         list[index] = decorator
-        applyDecorators(list, for: target)
+
+        guard var field = fieldMap[target.fieldID] else { return }
+        applyDecorators(list, for: target, in: &field)
+        ensureDecorateDisabledIfEmpty(for: target, in: &field)
+        commitDecoratorChange(field: field, fieldID: target.fieldID)
     }
 }
 
@@ -197,7 +208,8 @@ private extension DocumentEditor {
 
     /// Sets `decorate = true` on the field (table) or schema entry (collection) when
     /// row-level decorators are added, so the decorator column shows automatically.
-    func ensureDecorateEnabled(for target: DecoratorTarget) {
+    /// Mutates `field` in place; the caller commits.
+    func ensureDecorateEnabled(for target: DecoratorTarget, in field: inout JoyDocField) {
         let schemaKey: String?
         switch target.terminator {
         case .commonRows(let sk):
@@ -207,7 +219,6 @@ private extension DocumentEditor {
         default:
             return
         }
-        guard var field = fieldMap[target.fieldID] else { return }
 
         if let sk = schemaKey {
             guard var schema = field.schema, var entry = schema[sk] else { return }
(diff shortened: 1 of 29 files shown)


Sources/JoyfillUI/ViewModels/DocumentEditor+Decorators.swift
```diff
@@ -156,209 +185,434 @@ private extension DocumentEditor {
...
+                    // collection requires schemas/sk between consecutive rows; tables don't nest
+                    return nil
+                }
+                hops.append(.init(schemaKey: hopSchema, rowID: rowID))
+                pendingSchema = nil
+                hasPendingSchema = false
 
-        // Root schema = the schema entry flagged with root == true
-        let rootSchemaKey = field.schema?.first { $0.value.root == true }?.key
+                guard valueElement(at: hops, in: field) != nil else { return nil }
```
vivekjyani, line 437: **Path validation gap after `/schemas/{sk}/…`:** When the pending segment is a row id, this `valueElement(at:)` check resolves the only hop by scanning `valueToValueElements` by **id only** (see `valueElement` for `i == hops.count - 1`). It does not verify that this row is actually a row **under** `sk`. So a path like `.../schemas/{nestedSk}/{rootRowId}` can succeed if `rootRowId` exists at the top level, even when that row is not under `nestedSk`. `applyDecorators` / `rewriteRows` then write to the **top-level** row, while `ensureDecorateEnabled` still keys off `hops.last?.schemaKey` and can flip the wrong `schema[sk].decorate`.

Consider validating the hop against the same parent/child relationship used for writes (e.g. for the first hop after `schemas/sk`, require the row to appear under that schema, not merely match an id in the root list).

Sources/JoyfillUI/ViewModels/DocumentEditor+Decorators.swift
```diff
@@ -156,209 +185,434 @@ private extension DocumentEditor {
         return nil
     }
 
-    // MARK: Field Decorators
+    // MARK: Decorate flag
+
+    /// Sets `decorate = true` on the field (table) or schema entry (collection) when
+    /// row-level decorators are added, so the decorator column shows automatically.
+    /// Mutates `field` in place; the caller commits.
+    func ensureDecorateEnabled(for target: DecoratorTarget, in field: inout JoyDocField) {
```
vivekjyani, line 199: **`decorate` vs `isDisplayable` asymmetry:** `addDecorators` always calls `ensureDecorateEnabled` for `commonRows` / `rowSelf` after a successful write, but `validateDecorator` only enforces `action` (and `color` format) — it does not require an icon or label. So a batch of only **non-displayable** decorators (e.g. action + color only) still sets `field.decorate` / `schema[sk].decorate` to `true` and the UI can reserve a decorator column for nothing visible, while `ensureDecorateDisabledIfEmpty` (below) correctly keys off `isDisplayable`.

If the product rule is “`decorate` reflects *visible* row decorators only,” consider either validating displayability for these terminators, or gating `ensureDecorateEnabled` on “scope now has a displayable row decorator (including common + per-row) after the mutation.”

Sources/JoyfillUI/ViewModels/Models.swift
```diff
@@ -370,28 +374,72 @@ struct TableDataModel {
...
+            let key = "\(id)/\(columnID)"
+            if cellSpecificDecorators.isEmpty {
+                self.tableCellDecorators[key] = cellCommonDecorators
+            } else {
+                self.tableCellDecorators[key] = cellSpecificDecorators
+            }
+        }
+    }
+    
+    mutating func setTableRowDecorators(rowDecorators: [Decorator]?, rows: [ValueElement]) {
```
vivekjyani, line 409: **Possible stale map entries:** `tableRowDecorators` and `tableCellDecorators` are only ever updated/merged for the current `rows` list, but this path never **removes** keys for row ids that disappeared (deleted rows, etc.). In edge cases (e.g. id reuse) `getTableRowDecorators` / `getTableCellDecorators` could return stale `DecoratorLocal` for an old id.

Consider clearing the dictionaries before the loop, or `removeValue(forKey:)` for ids not in the current set.

vivekjyani (commented):
Re-review pass: a few issues worth addressing — see **inline** comments (path resolution, `decorate` flag symmetry, and row/cell decorator cache).

Sources/JoyfillUI/ViewModels/DocumentEditor+Decorators.swift
vivekjyani: Thanks for flagging. You're right — ensureDecorateEnabled flips unconditionally while the disable side correctly filters by isDisplayable, so a batch of only non-displayable decorators leaves decorate = true over an empty visible scope.

Tracking as a follow-up — won't roll it into this PR. The fix is to gate ensureDecorateEnabled on the same "scope has a displayable row decorator" predicate the disable function already uses, so the flag reflects the visible state symmetrically. No validation rule change needed, since the storage model already allows non-displayable entries to coexist (as the disable side assumes).

Sources/JoyfillUI/ViewModels/Models.swift
vivekjyani: Fixed

Sources/JoyfillUI/ViewModels/DocumentEditor+Decorators.swift
vivekjyani: Good catch — landed a fix. The resolver now rejects /schemas/{sk}/{rowID} from the field root when sk isn't the root schema, since nested schemas have no top-level rows. Mid-chain hops still rely on transitive descent (parent.childrens[nextHop.schemaKey]) for schema enforcement, which catches most variants; full schema-graph validation at every hop is a wider change I left as a follow-up. /schemas/{rootSK}/{rowID} stays valid as the verbose spelling of /{rowID}. Tests added for both.

Sources/JoyfillUI/ViewModels/DocumentEditor+Decorators.swift
```diff
@@ -2,19 +2,10 @@ import JoyfillModel
...
-    // path = "pageId/fieldPositionId/rowId/colId"  → column decorators
-    //
-    // Note: Path segments are parsed with empty-component filtering;
-    // consecutive or trailing slashes are treated as single separators.
-    //
-    // All methods in this section must be called on the main thread,
-    // consistent with the DocumentEditor threading model.
-
-    /// Returns all decorators at the given path.
+    // MARK: - Decorator API (path-based)
```
VishnuBishnoi, line 5: **B2 — threading note regression from PR #308**

PR #308 specifically added (at reviewer request) both the main-thread contract and the consecutive-slash note to this MARK header. Both are now gone:

```swift
// MARK: - Decorator API (path-based)
//
// Note: Path segments are parsed with empty-component filtering;
// consecutive or trailing slashes are treated as single separators.
//
// All methods in this section must be called on the main thread,
// consistent with the DocumentEditor threading model.
```

Please restore them. The threading contract is still enforced by the implementation (all `DocumentEditor` mutations must run on the caller's main thread), but integrators reading only the header get no hint. The consecutive-slash note matters because the new unbounded-depth path grammar creates more opportunities for accidental double-slash paths.

Sources/JoyfillUI/ViewModels/DocumentEditor+Decorators.swift
```diff
@@ -156,209 +185,437 @@ private extension DocumentEditor {
...
+    // MARK: Row tree walking
+
+    /// Walks the hop chain and returns the ValueElement at the end (last hop's row).
+    func valueElement(at hops: [DecoratorTarget.Hop], in field: JoyDocField) -> ValueElement? {
+        guard !hops.isEmpty else { return nil }
+        var currentRows: [ValueElement] = field.valueToValueElements ?? []
+        for (i, hop) in hops.enumerated() {
+            guard let row = currentRows.first(where: { $0.id == hop.rowID }) else { return nil }
+            if i == hops.count - 1 { return row }
+            // Descend into child schema specified by NEXT hop's schemaKey
```
VishnuBishnoi, line 570: **B1 — deleted-row check missing**

`first(where: { $0.id == hop.rowID })` accepts soft-deleted rows (`deleted: true`). A stale path reference can resolve to a row that has already been logically removed from the document, and `addDecorators` / `removeDecorator` will then silently write to or read from that deleted element.

Fix:
```swift
guard let row = currentRows.first(where: { $0.id == hop.rowID && !($0.deleted ?? false) }) else { return nil }
```

The same guard should apply at each descent level (child rows are currently fetched without a `deleted` filter before the next iteration). A corresponding test: a path that points at a deleted row should resolve to `nil` and `getDecorators` should return `[]`.

Sources/JoyfillUI/ViewModels/DocumentEditor+Decorators.swift
```diff
@@ -156,209 +185,437 @@ private extension DocumentEditor {
...
+            let nextHop = hops[i + 1]
+            guard let sk = nextHop.schemaKey,
+                  let children = row.childrens?[sk] else { return nil }
+            currentRows = children.valueToValueElements ?? []
         }
-        updateField(field: field)
-        refreshField(fieldId: fieldID)
-        valueDelegate(for: fieldID, fieldType: field.fieldType)?.decoratorsDidChange()
+        return nil
```
VishnuBishnoi: **N1 — unreachable code**

The for-loop always returns within its own body at the last hop iteration (either `return row` on a match or `return nil` via the `guard`). This final `return nil` can never execute. Either remove it or add `// unreachable` so future readers don't assume there's a path to it.

Sources/JoyfillUI/ViewModels/DocumentEditor+Decorators.swift
```diff
@@ -156,209 +185,437 @@ private extension DocumentEditor {
         return nil
     }
 
-    // MARK: Field Decorators
+    // MARK: Decorate flag
+
+    /// Sets `decorate = true` on the field (table) or schema entry (collection) when
+    /// row-level decorators are added, so the decorator column shows automatically.
+    /// Mutates `field` in place; the caller commits.
+    func ensureDecorateEnabled(for target: DecoratorTarget, in field: inout JoyDocField) {
```
VishnuBishnoi, line 199: **N2 — silent no-op for cell terminators needs a comment**

The `default: return` here covers `.field`, `.commonColumn`, `.cell`, and `.rowScopedColumn`. The first two are intuitive (non-row scopes don't reserve the decorator column), but `.cell` and `.rowScopedColumn` are silent surprises — adding a cell-specific decorator via `addDecorators` does NOT flip `decorate` on, which means the row-decorator column won't appear for cell-only decorator writes.

If that's intentional (cell decorators render within the cell body, not in the row-decorator column), spell it out:

```swift
case .cell, .rowScopedColumn:
    // Cell/row-scoped-column decorators render within the cell itself
    // and do not require the dedicated row-decorator column; skip.
    return
default:
    return
```

If it's not intentional, and cell decorators DO require a `decorate` column, this is a bug.

Sources/JoyfillUI/ViewModels/DocumentEditor+Decorators.swift
```diff
@@ -156,209 +185,437 @@ private extension DocumentEditor {
...
+public struct DecoratorConfig {
+    /// Maximum number of field-level decorators (`field.decorators`) shown
+    /// inline next to the field title before the rest move into a kebab menu.
+    public var visibleLimitInFields: Int
+
+    /// Maximum number of row decorators (merged common + row-self) shown
+    /// inline on a row before the rest move into a kebab menu.
+    public var visibleLimitInRows: Int
+
+    public init(visibleLimitInFields: Int = 2, visibleLimitInRows: Int = 1) {
```
VishnuBishnoi, line 630: **N3 — no discoverable "show all" sentinel**

The only way to show all decorators inline is to set `visibleLimitInFields = Int.max`, but that's not documented and easy to miss. A public constant makes the intent explicit:

```swift
public extension DecoratorConfig {
    /// Pass as `visibleLimitInFields` or `visibleLimitInRows` to show all
    /// decorators inline without collapsing into a kebab menu.
    static let unlimited: Int = .max
}
```

The demo app steppers naturally cap this, but integrators configuring limits programmatically have no discoverable way to reset to "show everything".

VishnuBishnoi (changes requested):
Large, ambitious PR. The path-grammar redesign (`Hop + Terminator`), COW seeding, single-commit write path, and kebab overflow are all solid pieces of work. Two correctness issues below need to be fixed before merge; the rest are non-blocking observations.

---

## What works well

**Path resolver overhaul** — replacing the old 4-segment `parsePath()` + enum target with the `Hop[]` + `Terminator` model lets the API support unbounded nesting depth without any further changes to the resolver. The `pendingSchema` / `hasPendingSchema` state machine is correct and the grammar is well-exercised by the new test suites.

**Single-commit write path** — `applyDecorators` → `ensureDecorate*` → `commitDecoratorChange` in a single `var field` mutation eliminates the duplicate `updateField` / `refreshField` / `decoratorsDidChange` sequences that existed before. Good.

**COW seeding** — the seed is correctly gated on `list.isEmpty` (only seeds a fresh scope) and runs before the duplicate-action check, so the seeded entries count as existing for collision purposes. The `commonSeed(for:)` helper is clean.

**`decorate` flag symmetry** — `ensureDecorateEnabled` and `ensureDecorateDisabledIfEmpty` cover add/remove/update correctly for `commonRows` and `rowSelf` terminators. `ensureDecorateDisabledIfEmpty` correctly short-circuits when any common-scope decorator is still displayable, preventing premature flag clearing after a partial remove.

**`setTableRowDecorators` stale-entry fix** — dictionaries are cleared before the loop and rebuilt fresh; the previously reported stale-map issue is resolved.

**Test coverage** — 7 dedicated test suites, 14 decorate-flag cases across table / collection / nested schema, COW tests, deep-path tests, and UI tests. The removed "one-way latch" tests have been correctly replaced with inverted assertions.

---

## Blockers (2)

See inline comments for details.

### B1 — `valueElement(at:in:)` does not check the `deleted` flag
A path that resolves to a soft-deleted row is accepted by `resolveDecoratorTarget` and passed through to `applyDecorators` / `fetchDecorators`. Decorators written to a deleted row are invisible at render time but persist in the data model. If the row is ever un-deleted they surface unexpectedly. The fix is a one-liner: `first(where: { $0.id == hop.rowID && !($0.deleted ?? false) })`.

### B2 — Main-thread documentation note removed
PR #308 explicitly requested (and added) the note:
```
// All methods in this section must be called on the main thread,
// consistent with the DocumentEditor threading model.
```
That note is gone from the new MARK header. This is a documentation regression — the threading contract is still required by the implementation (all `DocumentEditor` mutations must be on main), but integrators reading only the header comment get no hint. Please restore the note (and the consecutive-slash note while you're there).

---

## Non-blocking observations

### N1 — Unreachable `return nil` at the end of `valueElement(at:in:)`
The for-loop always returns at the last hop (either `row` on match, or `nil` via the guard). The `return nil` at line 563 is dead code. Remove it or add a `// unreachable` comment so the compiler doesn't generate a spurious warning in future.

### N2 — `ensureDecorateEnabled` silently ignores `.cell` and `.rowScopedColumn` without explanation
The `default: return` in `ensureDecorateEnabled` covers `.field`, `.commonColumn`, `.cell`, and `.rowScopedColumn`. The intent is clear for `.field`/`.commonColumn` (they don't reserve a row-decorator column), but `.cell` and `.rowScopedColumn` terminators are silent no-ops for the same reason and should say so:
```swift
case .cell, .rowScopedColumn:
    // Cell decorators render within the cell itself; they do not need
    // the dedicated row-decorator column and do not flip `decorate`.
    return
```

### N3 — `DecoratorConfig` has no semantic constant for "show all"
Using `Int.max` as a sentinel is opaque. Add:
```swift
public extension DecoratorConfig {
    /// Pass as either limit to show all decorators inline without a kebab.
    static let unlimited = Int.max
}
```
or a static property on the struct. The demo app's steppers clip this naturally, but integrators setting limits in code have no discoverable way to reset to "all".

### N4 — `setTableCellDecorators` writes `tableCommonCellDecorators` once per (row × column)
`column.decorators` is schema-level — the same for every row. Writing `tableCommonCellDecorators[columnID] = cellCommonDecorators` inside the per-row loop overwrites the identical value N times (once per row). Hoist the common-cell-decorator setup out of `setTableRowDecorators`'s row loop and compute it once per column.

---

## Summary

| | Issue | Verdict |
|---|---|---|
| B1 | `valueElement(at:)` allows decorating deleted rows | **Must fix** |
| B2 | Main-thread threading note removed from MARK header | **Must fix** |
| N1 | Unreachable `return nil` in `valueElement` | Non-blocking |
| N2 | `ensureDecorateEnabled` silent no-op for cell terminators | Non-blocking |
| N3 | No `DecoratorConfig.unlimited` sentinel | Non-blocking |
| N4 | `tableCommonCellDecorators` set N×M times instead of M | Non-blocking |

Sources/JoyfillUI/ViewModels/DocumentEditor+Decorators.swift
vivekjyani: Good catch — fixed.

Added `!($0.deleted ?? false)` to the predicate at both spots: `valueElement(at:)` (the read path) and `rewriteRows` (the write path), so soft-deleted rows are treated as absent for both lookups. The descent step doesn't need its own filter — the next iteration's `first(where:)` carries the guard, so paths through or terminating at a deleted row resolve to nil at that level.

Tests added in `DecoratorPublicAPITests`:
- `getDecorators` on deleted table row-self / cell / collection root row-self paths returns `[]`
- `addDecorators` targeting a deleted row is a no-op on the underlying `ValueElement.decorators`

Sources/JoyfillUI/ViewModels/DocumentEditor+Decorators.swift
vivekjyani: Restored both notes verbatim above the MARK header — the consecutive-slash parsing note and the main-thread contract. Agree they're worth keeping; the threading hint belongs alongside the rest of the DocumentEditor model, and the slash-parsing note is more relevant now that the unbounded-depth grammar makes accidental `//` paths easier to hit.

Sources/JoyfillUI/ViewModels/DocumentEditor+Decorators.swift
vivekjyani: Removing it would fail compilation, so added a one-liner `// unreachable: loop always returns; required to satisfy the compiler.` next to it.

Sources/JoyfillUI/ViewModels/DocumentEditor+Decorators.swift
vivekjyani: Intentional. Cell and row-scoped-column decorators render within the cell body (next to the column title via FieldDecoratorsView in TableModalTopNavigationView / CollectionModalTopNavigationView), not in the row-decorator column. `decorate` controls only the row-decorator column, so adding a cell decorator correctly leaves it alone.

Annotated the default branch in both ensureDecorateEnabled and the mirror ensureDecorateDisabledIfEmpty to spell this out:

default:
    // .field / .commonColumn / .cell / .rowScopedColumn — none reserve
    // the row-decorator column. Cell and row-scoped-column decorators
    // render within the cell body (next to the column title), so they
    // intentionally do not flip `decorate` on.
    return

Sources/JoyfillUI/ViewModels/DocumentEditor+Decorators.swift
vivekjyani: I get the intent here, but I’d lean against adding this in the SDK. Consumers can already pass Int.max in the config to show everything, and keeping it this way avoids adding extra surface area for something fairly straightforward.

VishnuBishnoi (approved):
Both blockers resolved. All four N-comments addressed. New commits add correct validation and solid test coverage. LGTM.

---

### Blocker resolution

**B1 — deleted-row check** ✅
`valueElement(at:in:)` now checks `!($0.deleted ?? false)` at every hop in the chain. `rewriteRows` does the same at its `firstIndex(where:)` guard. The test file `DecoratorPathResolutionTests.swift` adds explicit coverage for all three deleted-row resolution cases (table row-self, table cell, collection root row-self).

**B2 — threading note regression** ✅
The threading contract (`All methods in this section must be called on the main thread, consistent with the DocumentEditor threading model`) and the consecutive-slash behavior note are both preserved in the updated MARK header under the new title `// MARK: - Decorator API (path-based)`.

---

### N-comment resolution

**N1 — unreachable `return nil`** ✅ — Now annotated: `// unreachable: loop always returns; required to satisfy the compiler.`

**N2 — `ensureDecorateEnabled` silent exclusion** ✅ — Both `ensureDecorateEnabled` and `ensureDecorateDisabledIfEmpty` now carry an explicit comment: `// .field / .commonColumn / .cell / .rowScopedColumn — none reserve the row-decorator column. Cell and row-scoped-column decorators render within the cell body (next to the column title), so they intentionally do not flip decorate on.`

---

### New commits (post-REQUEST_CHANGES)

**"Empty decorators caches before filling"** ✅ — `setTableRowDecorators` now calls `tableRowDecorators.removeAll()`, `tableCellDecorators.removeAll()`, `tableCommonCellDecorators.removeAll()` before rebuilding. Correct defensive hygiene — prevents stale decorator entries surviving a refresh.

**"Reject /schemas/{nestedSK}/{rowID} from field root"** ✅ — The guard `if hops.isEmpty, hopSchema != rootSchemaKey { return nil }` correctly rejects paths that try to start a chain at a nested schema's rowID without first going through a parent row. The path grammar now matches the documented intent.

**`DecoratorDeepPathTests.swift`** ✅ — Covers multi-level hop chain resolution including deleted-row mid-chain scenarios.

**`DecoratorCOWTests.swift`** ✅ — Covers COW seeding isolation between rows and between field/row scope.

---

### Still-open (non-blocking, carry as follow-up)

- **N3** — No `DecoratorConfig.unlimited` sentinel. `Int.max` works as a workaround but an explicit constant improves readability for callers who want unlimited display.
- **N4** — `setTableCellDecorators` writes `tableCommonCellDecorators[columnID]` once per (row × column) iteration. Since `column.decorators` doesn't depend on the row, this is N redundant writes per column. Extracting the common-decorator write outside the outer loop would be O(columns) instead of O(rows × columns).

PR #317, opened 2026-04-28
NO-2075 Filter duplicate decorators and cap inline row decorators at 3

vivekjyani:
## 1. Context / Spec

Two decorator display issues:
- Decorator arrays decoded from JSON could contain entries with duplicate `action` values — only the first should be used.
- Row decorator columns are `40×60pt`. When `visibleLimitInRows` is set high (e.g. 200), the old code stacked all inline decorators, breaking layout. Fields and cells handle high counts with horizontal scroll; rows needed a different approach.

## 2. What Changed

**`Sources/JoyfillModel/JoyDoc.swift`**
- Added `extension Array where Element == Decorator` with `deduplicatedByAction()` — keeps the first decorator per `action`, preserving order; decorators with no action are always kept.
- Applied to all 6 decorator getters: `Decorators.all`, `Decorators.cells`, `JoyDocField.decorators`, `JoyDocField.rowDecorators`, `FieldTableColumn.decorators`, `Schema.rowDecorators`.
- Filtering is read-time only — raw JSON/dictionary is never mutated.

**`Sources/JoyfillUI/View/DecoratorView.swift`**

*`FieldDecoratorsView`*
- When `count ≤ visibleLimit`: inline decorators now render inside a `ScrollView(.horizontal)` so a high limit doesn't overflow the header.
- Popover content (for `count > visibleLimit`) wrapped in a `ScrollView` so it scrolls when item count is large.

*`RowDecoratorMenuView`*
- When `count > visibleLimit`: unchanged — shows kebab → popover (also now wrapped in `ScrollView`).
- When `count ≤ visibleLimit`: now caps inline display at **3 items** max. If there are more than 3, a non-tappable `•••` indicator fills the remaining slot. Previously it tried to stack all items, which broke at high counts.
- Removed the separate single-item branch — unified into the shared `VStack` path.
- Button label: removed hardcoded `height: 60` so items share column height correctly when stacked.

## 3. Design Decisions

- **Read-time dedup, not write-time:** Avoids silently rewriting stored data. If the source is fixed upstream the filter becomes a no-op.
- **Inline cap hardcoded at 3:** The `40×60pt` column can fit 3 small items comfortably. Making this configurable adds complexity with no practical value — anyone needing access to more items will set `visibleLimitInRows` low enough to trigger the kebab/popover.
- **`•••` is non-tappable:** It's a visual hint only. Users who need to act on hidden decorators should lower `visibleLimitInRows` to trigger the popover.
- **Plain `VStack` in popovers (not `LazyVStack`):** Deferred until there's a measured perf problem. Popovers are opened on demand so initial render cost is paid once per tap, not on scroll.

## 4. Screenshots / Video

_N/A — no visual design change to existing behaviour. To test: set `visibleLimitInRows = 10` with 5 decorators on a row → should see 3 stacked + `•••`. Set `visibleLimitInRows = 2` with 5 decorators → should see kebab._

## 5. Public API Impact

- `deduplicatedByAction()` is package-internal (`Array` extension, no `public` modifier) — not exposed.
- The 6 modified decorator getters are `public` computed properties but **signatures are unchanged** — only return value behaviour changes (duplicates stripped). No docs update needed.

## 6. Test Coverage

No new tests in this PR. Follow-up needed:
- Unit test for `deduplicatedByAction` (JSON with duplicate actions → assert first-wins).
- UI test for the 3-cap inline row behaviour.

## 7. Reviewer Notes

- `deduplicatedByAction()` lives directly below the `Decorator` struct in `JoyDoc.swift` — easy to locate.
- The `•••` overflow indicator in rows has no tap target by design. If a future spec requires tapping it, that's a separate ticket.
- Cell decorator popovers follow the same `ScrollView` wrapping pattern added here for consistency.

### DIFF
--- a/Sources/JoyfillUI/View/DecoratorView.swift
+++ b/Sources/JoyfillUI/View/DecoratorView.swift
@@ -186,9 +186,11 @@ struct FieldDecoratorsView: View {
                         decoratorPopover
                     }
             } else {
-                HStack(spacing: 4) {
-                    ForEach(Array(displayable.enumerated()), id: \.offset) { _, decorator in
-                        DecoratorButton(decorator: decorator, onTap: onDecoratorTap)
+                ScrollView(.horizontal, showsIndicators: false) {
+                    HStack(spacing: 4) {
+                        ForEach(Array(displayable.enumerated()), id: \.offset) { _, decorator in
+                            DecoratorButton(decorator: decorator, onTap: onDecoratorTap)
+                        }
                     }
                 }
             }
--- a/Sources/JoyfillUI/View/DecoratorView.swift
+++ b/Sources/JoyfillUI/View/DecoratorView.swift
@@ -238,21 +238,28 @@ struct RowDecoratorMenuView: View {
     let onDecoratorTap: (DecoratorLocal) -> Void
     @State private var showingPopover = false
 
+    private static let inlineMax = 3
     private var displayable: [DecoratorLocal] { decorators.filter { $0.isDisplayable } }
     private var exceedsLimit: Bool { displayable.count > visibleLimit }
+    private var inlineVisible: [DecoratorLocal] { Array(displayable.prefix(Self.inlineMax)) }
+    private var hasInlineOverflow: Bool { displayable.count > Self.inlineMax }
 
     var body: some View {
         if displayable.isEmpty {
             Color.clear.frame(width: 40, height: 60)
         } else if exceedsLimit {
             kebabButton.frame(width: 40, height: 60)
-        } else if displayable.count == 1, let decorator = displayable.first {
-            singleDecoratorButton(decorator)
         } else {
             VStack(spacing: 0) {
-                ForEach(Array(displayable.enumerated()), id: \.offset) { _, decorator in
+                ForEach(Array(inlineVisible.enumerated()), id: \.offset) { _, decorator in
                     singleDecoratorButton(decorator)
                         .frame(maxHeight: .infinity)
+                }
+                if hasInlineOverflow {
+                    Text("•••")
+                        .font(.system(size: 10, weight: .medium))
+                        .foregroundColor(.secondary)
+                        .frame(maxWidth: .infinity, maxHeight: .infinity)
                 }
             }
             .frame(width: 40, height: 60)
@@ -275,7 +282,7 @@ struct RowDecoratorMenuView: View {
                 }
             }
             .foregroundColor(tint)
-            .frame(width: 40, height: 60)
+            .frame(width: 40)
             .contentShape(Rectangle())
         }
         .buttonStyle(.plain)
--- a/Sources/JoyfillUI/View/DecoratorView.swift
+++ b/Sources/JoyfillUI/View/DecoratorView.swift
@@ -198,30 +198,32 @@ struct FieldDecoratorsView: View {
     }
 
     private var decoratorPopover: some View {
-        VStack(alignment: .leading, spacing: 0) {
-            ForEach(Array(displayable.enumerated()), id: \.offset) { index, decorator in
-                let tint: Color = decorator.color.map { Color(hex: $0) } ?? .primary
-                Button {
-                    showingOverflow = false
-                    onDecoratorTap(decorator)
-                } label: {
-                    HStack(spacing: 8) {
-                        if let icon = decorator.icon, !icon.isEmpty {
-                            DecoratorIconImage(iconName: icon, size: 14)
-                                .foregroundColor(tint)
-                                .frame(width: 20)
+        ScrollView {
+            VStack(alignment: .leading, spacing: 0) {
+                ForEach(Array(displayable.enumerated()), id: \.offset) { index, decorator in
+                    let tint: Color = decorator.color.map { Color(hex: $0) } ?? .primary
+                    Button {
+                        showingOverflow = false
+                        onDecoratorTap(decorator)
+                    } label: {
+                        HStack(spacing: 8) {
+                            if let icon = decorator.icon, !icon.isEmpty {
+                                DecoratorIconImage(iconName: icon, size: 14)
+                                    .foregroundColor(tint)
+                                    .frame(width: 20)
+                            }
+                            if let label = decorator.label, !label.isEmpty {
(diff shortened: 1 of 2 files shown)


Sources/JoyfillUI/View/DecoratorView.swift
```diff
@@ -234,22 +238,29 @@ struct RowDecoratorMenuView: View {
     let onDecoratorTap: (DecoratorLocal) -> Void
     @State private var showingPopover = false
 
+    private static let inlineMax = 3
     private var displayable: [DecoratorLocal] { decorators.filter { $0.isDisplayable } }
     private var exceedsLimit: Bool { displayable.count > visibleLimit }
+    private var inlineVisible: [DecoratorLocal] { Array(displayable.prefix(Self.inlineMax)) }
+    private var hasInlineOverflow: Bool { displayable.count > Self.inlineMax }
```
vivekjyani, line 245: Decorators beyond index 3 are silently inaccessible when `displayable.count` is between 4 and `visibleLimit`. The user sees `•••` but has no way to act on those hidden items — there is no tap target and the kebab is not shown (since `exceedsLimit` is false). Consider making `•••` tappable to open the popover, or changing `exceedsLimit` to also trigger when `displayable.count > inlineMax`.

Sources/JoyfillUI/View/DecoratorView.swift
```diff
@@ -289,33 +300,35 @@ struct RowDecoratorMenuView: View {
...
-                        if let icon = decorator.icon, !icon.isEmpty {
-                            DecoratorIconImage(iconName: icon, size: 14)
-                                .foregroundColor(tint)
-                                .frame(width: 20)
-                        }
-                        if let label = decorator.label, !label.isEmpty {
-                            Text(label)
-                                .font(.system(size: 14, weight: .medium))
-                                .foregroundColor(tint)
+        ScrollView {
```
vivekjyani, line 303: `ScrollView` inside a popover with no explicit height frame. SwiftUI popovers auto-size to content up to a system cap, so scrolling will work once that cap is hit. But for the common case (few items) the `ScrollView` adds no value and makes the layout harder to reason about. Consider wrapping only when item count exceeds a threshold, or add `.frame(maxHeight: X)` on the `ScrollView` to make the scroll boundary explicit.

Sources/JoyfillUI/View/DecoratorView.swift
vivekjyani: There is a way to access them — if you lower the `visibleLimitInRows`, the kebab icon will appear, and tapping on it will show all the decorators. 

The current behavior mainly applies when the limits are set higher.

Sources/JoyfillUI/View/DecoratorView.swift
vivekjyani: No change needed — SwiftUI's system height cap handles it fine in practice. Scrolling kicks in when the list is long, and short lists auto-size naturally. Leaving as-is.

VishnuBishnoi (approved):
Clean, focused PR. Both changes are correct. Pre-existing comment thread already surfaces the main design tradeoffs — nothing new to block on.

---

### `deduplicatedByAction()` — correct

- `seen.insert(action).inserted` is the idiomatic pattern (no pre-check needed). ✅
- `guard let action = dec.action, !action.isEmpty else { return true }` — treating empty-string as "no action" is the right defensive choice. ✅
- Read-time filter (not write-time) is the right call — avoids silently mutating stored data and becomes a no-op once the source is fixed upstream. ✅
- Applied consistently to all 6 relevant getters: `Decorators.all`, `Decorators.cells`, `JoyDocField.decorators`, `JoyDocField.rowDecorators`, `FieldTableColumn.decorators`, `Schema.rowDecorators`. `ValueElement.decorators?.all / .cells` deduplicate transitively. ✅
- Access level is `internal` (implicit, no `public`) — correct, since the function is only called by the getters within `JoyfillModel`. ✅

**One minor note:** Each getter call allocates a `Set<String>` + filtered `Array`. For typical decorator counts (< 10) this is negligible. If `decorators` / `rowDecorators` getters are ever on a measured hot path (e.g. called in `body` on every SwiftUI redraw across many rows), memoizing the result on the `Dictionary` read would eliminate the repeated work. Not a blocker — flagging for awareness.

---

### `RowDecoratorMenuView` — correct

- **`singleDecoratorButton` height fix** ✅ — Removing `.frame(width: 40, height: 60)` from the button and leaving only `.frame(width: 40)` is the right fix. With the fixed height on the outer VStack (`.frame(width: 40, height: 60)`) and `.frame(maxHeight: .infinity)` on each item, all items share the 60pt height equally. The old hardcoded `height: 60` per button was a latent bug (a 2-item stack would have been 120pt tall in a 60pt container).
- **Unified VStack path** ✅ — Removing the separate single-item branch simplifies the logic without changing behaviour.
- **`inlineMax = 3` static constant** ✅ — Correct placement as a `private static let`. The 40×60pt column constraint makes this a physical constant rather than a user-facing config.
- **`•••` overflow logic** ✅ — `hasInlineOverflow = displayable.count > Self.inlineMax` is independent of `exceedsLimit = displayable.count > visibleLimit`, so the two conditions compose correctly across all cases.

---

### `FieldDecoratorsView` — correct

- **Inline `ScrollView(.horizontal, showsIndicators: false)`** ✅ — Safe addition. For the typical case (2–3 items, fits without scroll) the `ScrollView` is visually invisible. For high `visibleLimit` with many items it provides horizontal escape without layout overflow. `showsIndicators: false` is appropriate here.
- **Popover `ScrollView`** ✅ — SwiftUI popovers bound the content to a system height cap; the `ScrollView` kicks in only when that cap is hit. The `minWidth: 160` on the inner `VStack` (row) / outer `ScrollView` (field) ensures the popover minimum width is preserved.

---

### Follow-ups (non-blocking, already acknowledged in PR)

- Unit test for `deduplicatedByAction`: JSON with duplicate actions → assert first-wins, order-preserved, nil-action entries always kept.
- UI test or snapshot for the 3-cap inline row: `visibleLimitInRows = 200`, 5 decorators → assert 3 items + `•••` visible, kebab absent.

LGTM — approving.

PR #319, opened 2026-04-29
NO-2095 Fix duplicate-page without-values to preserve readonly and block values

vivekjyani:
## Context
This PR implements the duplicate-page spec behavior for **"without values"** mode.
When duplicating a page without values, we should clear only user-entered values while preserving static/non-editable content in the duplicated page.

## What changed
- Updated duplicate logic in [DocumentEditor](Sources/JoyfillUI/ViewModels/DocumentEditor.swift) inside `duplicatePage(pageID:copyWithValues:)`.
- In the `!copyWithValues` path, values are now preserved for:
  - read-only fields (`disabled == true`)
  - display text / block fields (`fieldType == .block`)
- Rephrased the in-code comment for clarity.
- Added UI test coverage in `PageDeletableCopyableUITests` for this exact behavior.
- Updated `PageDeletableCopyable.json` fixture to include Page 4 data needed for assertions:
  - readonly text value
  - block/display text value (`Page 4`)

## Design decisions
- Read-only detection uses persisted field state (`disabled == true`) as the source of truth.
- Display text detection uses `FieldTypes.block`, since block content is stored in field value and should persist when copying without values.
- UI test validates preservation behavior through rendered output after duplication (not just model assumptions), including a guard against false positives for block text.

## Visuals
- No UI layout changes in this PR.
- Behavior-only change; can provide a quick screen recording if needed.

## Public API / docs impact
- Public API: **No**
- Docs update required: **No** (internal duplication behavior fix + test coverage)

## Test coverage
- Included in this PR:
  - Existing tests for page copy/delete behavior remain intact.
  - New dedicated UI test verifies `Without Values` duplication preserves both readonly value and block text on duplicated page.
- Additional unit tests are optional follow-up; current UI coverage validates end-to-end behavior for this scenario.

## Notes for reviewers
- Functional scope is limited to `copyWithValues == false` behavior in page duplication.
- Main review points:
  - field-type/readonly guards in duplication logic
  - correctness and reliability of new Page 4 fixture + UI assertions


JoyfillSwiftUIExample/JoyfillUITests/PageDeletableCopyable/PageDeletableCopyableUITests.swift
```diff
@@ -0,0 +1,602 @@
...
+
+        XCTAssertTrue(waitUntil(5) { pageRows.count == initialCount + 1 }, "Page count should increase by 1 after duplication.")
+
+        let duplicatedRow = pageRows.element(boundBy: sourceIndex + 1)
+        XCTAssertTrue(duplicatedRow.waitForExistence(timeout: 3), "Duplicated row should appear immediately after Page 4.")
+        duplicatedRow.tap()
+
+        guard let duplicatedValue = firstTextFieldValue() else { return }
+        XCTAssertNotEqual(
+            duplicatedValue,
```
vivekjyani: Weak assertion: this only checks that the duplicated value is *not equal* to the edited value, not that it was actually cleared. The test passes even if the value is the fixture's original `"Beta sample content"`. Consider asserting the value is empty/nil to truly verify the clear-on-duplicate behavior.

JoyfillSwiftUIExample/JoyfillUITests/PageDeletableCopyable/PageDeletableCopyableUITests.swift
```diff
@@ -0,0 +1,602 @@
...
+        app.swipeUp()
+    }
+
+    private func currentTestSelector(from name: String) -> String {
+        let trimmed = name.hasSuffix("]") ? String(name.dropLast()) : name
+        return trimmed.split(separator: " ").last.map(String.init) ?? name
+    }
+
+    private var usesReadonlyMode: Bool {
+        let selector = currentTestSelector(from: self.name)
```
vivekjyani, line 116: Fragile test-name parsing: parsing `self.name` (format: `-[ClassName methodName]`) to conditionally inject `--mode readonly` silently breaks if this method is ever renamed. Consider a dedicated subclass or a `setUp` property flag instead.

JoyfillSwiftUIExample/JoyfillUITests/PageDeletableCopyable/PageDeletableCopyable.json
```diff
@@ -0,0 +1,698 @@
...
+          "copyable": [
+            "without-values"
+          ],
+          "width": 816,
+          "height": 1056,
+          "rowHeight": 8,
+          "cols": 8,
+          "fieldPositions": [
+            {
+              "_id": "69ca48b311043e3715a0476c",
```
vivekjyani: Duplicate `fieldPositions._id` values: `"69ca48b311043e3715a0476c"` and `"69ca48c3f2727cea6f222db4"` are reused as field position IDs across Pages 3–8. Field position IDs should be unique; depending on how the SDK processes them, this could cause silent collisions.

JoyfillSwiftUIExample/JoyfillUITests/PageDeletableCopyable/PageDeletableCopyable.json
```diff
@@ -0,0 +1,698 @@
...
+      "value": [
+        "69c7ae51aa908fed6a37d3c5"
+      ]
+    },
+    {
+      "file": "68d101cf56b1d0b1393bb36a",
+      "_id": "69ca48e3c9d13aa7a878ad5d",
+      "type": "text",
+      "title": "Text",
+      "identifier": "field_69ca48b31981578ede809bd4",
```
vivekjyani: Duplicate field `identifier` values: this field and several others across Pages 3–8 share `"identifier": "field_69ca48b31981578ede809bd4"` despite having distinct `_id` values. If the SDK resolves fields by identifier, this ambiguity could cause unexpected behavior in tests.

JoyfillSwiftUIExample/JoyfillUITests/PageDeletableCopyable/PageDeletableCopyableUITests.swift
vivekjyani: Updated this assertion to verify explicit cleared state instead of just inequality.
`testDuplicateWithoutValuesResetsFieldData` now checks the duplicated editable text is empty (trimmed), which ensures we validate true clear-on-duplicate behavior rather than a different non-matching value.

JoyfillSwiftUIExample/JoyfillUITests/PageDeletableCopyable/PageDeletableCopyableUITests.swift
vivekjyani: Thanks for calling this out. I agree this is brittle and should be refactored to avoid parsing self.name. I’ve deferred that cleanup for now to keep this PR scoped to the duplication fix and coverage, and will follow up with a small test-infra PR to move readonly launch config to an explicit, stable path (no test-name parsing).

JoyfillSwiftUIExample/JoyfillUITests/PageDeletableCopyable/PageDeletableCopyable.json
vivekjyani: Great catch. I updated the PageDeletableCopyable.json fixture so fieldPositions._id values are unique across pages (no reused IDs across Pages 3–8). This removes potential ID-collision ambiguity during SDK processing.

JoyfillSwiftUIExample/JoyfillUITests/PageDeletableCopyable/PageDeletableCopyable.json
vivekjyani: Thanks for raising this. I normalized the fixture so each field instance has a unique identifier on Pages 3–8 (aligned with each field’s _id) to remove any ambiguity in identifier-based resolution paths. This keeps the test data deterministic and avoids potential collisions if lookup behavior changes in future SDK updates.

VishnuBishnoi (approved):
Production fix is correct and minimal. Fixture data verified against the actual JSON; test assertions are grounded in real values. Two non-blocking notes.

---

### Production change — correct

```swift
let isReadonly    = newFields[i].disabled == true
let isDisplayText = newFields[i].fieldType == .block
if isReadonly || isDisplayText { continue }
newFields[i].value = nil
```

**`disabled == true`** — consistent with `DocumentEditor.swift` line 657, which uses the same predicate to set field edit mode: `((fieldData?.disabled == true) || (mode == .readonly) ? .readonly : .fill)`. `disabled == nil` correctly falls through and gets cleared. ✅

**`fieldType == .block`** — block/display-text fields hold template content, not user input. Preserving them is correct. ✅

**Fixture verification** — pulled the Page 4 fields directly from the JSON:

| Field | type | `disabled` | Result |
|---|---|---|---|
| `69ca48e8...` | text | `null` | "Beta sample content" → cleared ✅ |
| `69ca48e8...` | multiSelect | `null` | `[Yes, No]` → cleared ✅ |
| `69d7a8ba...` | text | **`true`** | "Readonly fixture value" → preserved ✅ |
| `69d7a8ba...` | block | `null` | "Page 4" → preserved ✅ |
| `69f1b6b2...` | block | `null` | "Page 4" → preserved ✅ |

`assertFieldTextVisible("Readonly fixture value")` has a concrete value to match. `assertStaticTextOccurrences("Page 4", minimumCount: 2)` finds both block fields (both carry `value: "Page 4"`), independently of how the duplicated page title renders. ✅

---

### Tests — good overall

- JSON-driven `loadExpectedPageBehaviorFromJSON` keeps the page matrix in one place — test assertions stay in sync with the fixture automatically. ✅
- `chooseDuplicateModeOrFail` falls back from `alert.buttons` → `app.buttons` for environments where the alert renders differently. Resilient. ✅
- `waitUntil(5) { pageRows.count == initialCount + 1 }` is condition-based, not a fixed sleep. ✅
- `requireHittable: true` guard in `setFirstTextFieldValue` prevents accidentally targeting the disabled field (which lives at `y=36`, below the editable one at `y=5`). ✅

**N1 — duplication tests use `element(boundBy: sourceIndex)` without scroll search:**
`testDeletePageRemovesItFromPageSheet` and `testDeleteCancelKeepsPage` use `findRow(named:in:maxScrolls:)`, which scrolls to find the target. The three duplication tests use positional access instead. Page 4 is the 4th of 8 pages so it will almost always be visible, and the `rowContainsPageName` assertion will fail fast if the wrong row is tapped — it won't silently act on the wrong page. Low risk, but worth aligning with the scroll-search pattern if these tests become flaky on physical devices.

**N2 — no unit test for the model-layer change:**
A unit test calling `documentEditor.duplicatePage(pageID:copyWithValues: false)` and directly asserting `document.fields` would be faster and fully deterministic. The PR acknowledges this as optional follow-up. Acceptable given the UI test covers the end-to-end path.

---

### Scope

Other field types that might warrant special treatment in a "without values" copy (formula-computed fields, image fields) are not addressed. Given the PR explicitly scopes to `disabled` and `.block`, and no spec covers those cases here, that's appropriate.

LGTM — approving.

PR #320, opened 2026-05-01
NO-2108: Scale row decorator column width and render inline horizontally

vivekjyani:
## Context
Row decorators in tables and collections need to respect the `decoratorConfig.visibleLimitInRows` setting visually — both in how many decorators render inline before collapsing to a kebab, and in how wide the dedicated decorator column reserves for itself. Previously the column was hardcoded at 40pt regardless of how many decorators were configured to show, so anything more than one inline decorator got cramped. The inline rendering also stacked vertically with a `•••` overflow indicator, which didn't scale well with longer decorator sets.

## What changed
- **`RowDecoratorMenuView`** ([DecoratorView.swift](Sources/JoyfillUI/View/DecoratorView.swift)): below-limit rows now render decorators in a horizontal `ScrollView` of `DecoratorButton`s instead of a vertical stack with a `•••` overflow indicator. Above-limit rows still collapse to a single kebab.
- **Shared width helper** ([CollectionViewModel.swift:2259](Sources/JoyfillUI/View/Fields/CollectionView/CollectionViewModel.swift)): added `decoratorsCellWidth()` as a default implementation on `TableDataViewModelProtocol` (returns 40 / 80 / 100 for limit ≤ 1 / 2 / ≥ 3). Both `TableViewModel` and `CollectionViewModel` get it via conformance — one source of truth.
- **Table** ([TableModalView.swift](Sources/JoyfillUI/View/Fields/TableView/TableModalView.swift)): row decorator moved out of the sticky left pane into the horizontally scrolling area. Added a decorator header cell at the start of `colsHeader` and a per-row decorator cell at the start of `TableRowView`. `leftColumnWidth` and the left header HStack no longer reserve space for it. Cleaned up the now-stale commented-out block in `rowsHeader`.
- **Collection** ([CollectionModelView.swift](Sources/JoyfillUI/View/Fields/CollectionView/CollectionModelView.swift)): kept the decorator inside `CollectionRowsHeaderView` (collection has no sticky-pane split, so there's no reason to move it). Header cell in `CollectionColumnHeaderView` and per-row cells in both `.row` and `.nestedRow` now use `decoratorsCellWidth()` so they stay aligned.
- **Width plumbing** ([Utility.swift](Sources/JoyfillUI/View/Utility.swift)): `getWidthForExpanderRow` accepts a `rowDecoratorsCellWidth` parameter (default 40) so collection's `collectionWidth` calculation reflects the dynamic width. Threaded through three call sites in `CollectionViewModel`.

## Decisions and rationale
- **Why width 40 / 80 / 100 instead of `limit × 40`?** `visibleLimitInRows` is a *collapse threshold*, not a render count — a row with `count ≤ limit` renders all of its decorators in a horizontal scroll inside the cell. Capping the reserved width at 100 keeps the column from dominating the table when limits are configured high (e.g. 5–10), while still giving inline rows real space to breathe. 40 stays the default for the common limit-of-1 case.
- **Why option B over a per-row max-count scan?** Considered computing the max actual decorator count across all rows but rejected it on perf grounds (collection has nested schemas and frequent reflows). The chosen approach is cosmetic-loose for rows whose count exceeds the limit (kebab in a wider cell) but bounded and cheap.
- **Why move only the table's decorator and not the collection's?** Table has a sticky left pane that the decorator was trapped in; moving it into the scroll area was the user-visible goal. Collection has no sticky pane — everything already scrolls together — so the decorator stays where it was. Both use the same width helper for visual consistency.

## Screenshots / video
_TODO: attach a short clip showing the table column scrolling horizontally with the decorators, and the collection rows with the wider cell when `visibleLimitInRows = 2 or 3`._

## Public API
- No public API changes. The new `decoratorsCellWidth()` is added on the **internal** `TableDataViewModelProtocol`.
- `Utility.getWidthForExpanderRow` gained an optional parameter `rowDecoratorsCellWidth: CGFloat = 40` — internal type, default keeps any existing caller's behavior identical.
- No docs update needed.

## Tests
- No new tests in this PR. The change is purely visual/layout. Will follow up with snapshot tests in a separate PR if/when the snapshot infra lands for these views.
- Manually verified: table decorator column scrolls with data columns; collection rows align with header at limit 1, 2, and ≥3; inline horizontal scroll renders correctly for below-limit rows; kebab still appears for above-limit rows.

## Notes for reviewer
- Width formula's cap at 100 is a judgment call — happy to bump to 120 (the previous attempt) if real usage shows decorators getting clipped.
- The `inlineMax = 3` constant in `RowDecoratorMenuView` is now effectively dead (the body uses `displayable` directly). Left as-is to keep this PR scoped, but worth removing in cleanup.
- Commented-out decorator block previously in `rowsHeader` of the table is removed — no longer needed since the decorator now lives in the scrolling area.

### DIFF
--- a/JoyfillSwiftUIExample/JoyfillUITests/DecoratorUITests/Decorator.json
+++ b/JoyfillSwiftUIExample/JoyfillUITests/DecoratorUITests/Decorator.json
@@ -130,6 +130,16 @@
               "field" : "699dc392acb30df81f5a694a",
               "type" : "image",
               "displayType" : "original"
+            },
+            {
+              "_id" : "6a1f0c00a1b2c3d4e5f60718",
+              "x" : 0,
+              "height" : 8,
+              "y" : 170,
+              "width" : 4,
+              "field" : "6a1f0c00a1b2c3d4e5f60719",
+              "type" : "text",
+              "displayType" : "original"
             }
           ],
           "hidden" : false,
@@ -795,6 +805,15 @@
       "title" : "Multiple Choice",
       "_id" : "69709dc240f1e385f35abd36",
       "multi" : true
+    },
+    {
+      "title" : "Extra Text",
+      "identifier" : "field_6a1f0c00a1b2c3d4e5f60719",
+      "value" : "",
+      "type" : "text",
+      "file" : "691f3762c80bfb0005c57b48",
+      "_id" : "6a1f0c00a1b2c3d4e5f60719",
+      "hidden" : false
     }
   ]
 }
--- a/JoyfillSwiftUIExample/JoyfillUITests/DecoratorUITests/DecoratorUITestSupport.swift
+++ b/JoyfillSwiftUIExample/JoyfillUITests/DecoratorUITests/DecoratorUITestSupport.swift
@@ -63,8 +63,9 @@ enum DecoratorUITestSupport {
     }
 
     /// Identifier for a single row-level decorator button (shown when exactly one is displayable).
+    /// Row and field buttons currently share the same DecoratorButton identifier format.
     static func rowDecoratorID(action: String) -> String {
-        return "row_decorator_\(action)"
+        return "decorator_button_\(action)"
     }
 
     /// Identifier for the row hamburger menu button (shown when multiple decorators exist on a row).
--- a/Sources/JoyfillUI/View/DecoratorView.swift
+++ b/Sources/JoyfillUI/View/DecoratorView.swift
@@ -242,11 +242,8 @@ struct RowDecoratorMenuView: View {
     let onDecoratorTap: (DecoratorLocal) -> Void
     @State private var showingPopover = false
 
-    private static let inlineMax = 3
     private var displayable: [DecoratorLocal] { decorators.filter { $0.isDisplayable } }
     private var exceedsLimit: Bool { displayable.count > visibleLimit }
-    private var inlineVisible: [DecoratorLocal] { Array(displayable.prefix(Self.inlineMax)) }
-    private var hasInlineOverflow: Bool { displayable.count > Self.inlineMax }
 
     var body: some View {
         if displayable.isEmpty {
--- a/Sources/JoyfillUI/View/DecoratorView.swift
+++ b/Sources/JoyfillUI/View/DecoratorView.swift
@@ -254,43 +254,18 @@ struct RowDecoratorMenuView: View {
         } else if exceedsLimit {
             kebabButton.frame(width: 40, height: 60)
         } else {
-            VStack(spacing: 0) {
-                ForEach(Array(inlineVisible.enumerated()), id: \.offset) { _, decorator in
-                    singleDecoratorButton(decorator)
-                        .frame(maxHeight: .infinity)
-                }
-                if hasInlineOverflow {
-                    Text("•••")
-                        .font(.system(size: 10, weight: .medium))
-                        .foregroundColor(.secondary)
-                        .frame(maxWidth: .infinity, maxHeight: .infinity)
+            GeometryReader { geometry in
+                ScrollView(.horizontal, showsIndicators: false) {
+                    HStack(spacing: 4) {
+                        ForEach(Array(displayable.enumerated()), id: \.offset) { _, decorator in
+                            DecoratorButton(decorator: decorator, onTap: onDecoratorTap)
+                        }
+                    }
+                    .frame(minWidth: geometry.size.width, alignment: .center)
                 }
             }
-            .frame(width: 40, height: 60)
+            .frame(height: 32)
         }
-    }
-
-    private func singleDecoratorButton(_ decorator: DecoratorLocal) -> some View {
-        let tint: Color = decorator.color.map { Color(hex: $0) } ?? .secondary
-        return Button {
-            onDecoratorTap(decorator)
-        } label: {
-            HStack(spacing: 4) {
-                if let icon = decorator.icon, !icon.isEmpty {
-                    DecoratorIconImage(iconName: icon, size: 14)
-                }
-                if let label = decorator.label, !label.isEmpty {
-                    Text(label)
-                        .font(.system(size: 12, weight: .medium))
-                        .lineLimit(1)
-                }
-            }
-            .foregroundColor(tint)
-            .frame(width: 40)
-            .contentShape(Rectangle())
-        }
-        .buttonStyle(.plain)
-        .accessibilityIdentifier("row_decorator_\(decorator.action ?? "unknown")")
     }
 
     private var kebabButton: some View {
--- a/Sources/JoyfillUI/View/Fields/CollectionView/CollectionViewModel.swift
+++ b/Sources/JoyfillUI/View/Fields/CollectionView/CollectionViewModel.swift
@@ -503,7 +503,7 @@ class CollectionViewModel: ObservableObject, TableDataViewModelProtocol {
     }
     
     func rowWidth(_ tableColumns: [FieldTableColumn], _ level: Int, _ schemaKey: String, tableDataModel: TableDataModel) -> CGFloat {
-        return Utility.getWidthForExpanderRow(columns: tableColumns, showSelector: showRowSelector(for: tableDataModel), showSingleClickEdit: showSingleClickEditButton(for: tableDataModel), showRowDecorators: tableDataModel.hasAnyRowDecorators(schemaKey: schemaKey)) + Utility.getTotalTableScrollWidth(level: level)
+        return Utility.getWidthForExpanderRow(columns: tableColumns, showSelector: showRowSelector(for: tableDataModel), showSingleClickEdit: showSingleClickEditButton(for: tableDataModel), showRowDecorators: tableDataModel.hasAnyRowDecorators(schemaKey: schemaKey), rowDecoratorsCellWidth: decoratorsCellWidth()) + Utility.getTotalTableScrollWidth(level: level)
     }
     
     func getCollectionWidth(tableDataModel: TableDataModel) -> CGFloat {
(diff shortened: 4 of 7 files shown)


Sources/JoyfillUI/View/DecoratorView.swift
```diff
@@ -254,43 +254,18 @@ struct RowDecoratorMenuView: View {
         } else if exceedsLimit {
             kebabButton.frame(width: 40, height: 60)
```
vivekjyani, line 252: **Outer-height clipping for the kebab and empty branches.**

This branch (and the empty-displayable branch just above) keeps `.frame(width: 40, height: 60)` on the inner content. Each call site then wraps the view in `.padding(.all, 4).frame(width: decoratorsCellWidth(), height: 60)` (TableModalView.swift:25, CollectionModelView.swift:640, :685). That makes inner content effectively 68pt tall inside a 60pt outer frame — SwiftUI doesn't truncate, so the kebab can bleed into the row below in `LazyHStack(alignment: .top)`. The new inline branch already does the right thing (`.frame(height: 32)` only). Either drop the inner `height: 60` here and on the `Color.clear` line above, or move `.padding(.all, 4)` inside `RowDecoratorMenuView` so all three branches share the same outer sizing contract.

Sources/JoyfillUI/View/DecoratorView.swift
```diff
@@ -254,43 +254,18 @@ struct RowDecoratorMenuView: View {
...
-        } label: {
-            HStack(spacing: 4) {
-                if let icon = decorator.icon, !icon.isEmpty {
-                    DecoratorIconImage(iconName: icon, size: 14)
-                }
-                if let label = decorator.label, !label.isEmpty {
-                    Text(label)
-                        .font(.system(size: 12, weight: .medium))
-                        .lineLimit(1)
+            GeometryReader { geometry in
```
vivekjyani, line 254: Minor: `inlineMax`, `inlineVisible`, and `hasInlineOverflow` (lines 245, 248-249, just above the hunk) are dead after this refactor — `body` only references `displayable` and `exceedsLimit`. The PR description acknowledges this and defers cleanup, but it's a 4-line delete; leaving it just sets a trap for the next reader (someone might "fix" the inline branch to use `inlineVisible` again). Worth removing in this PR.

Sources/JoyfillUI/View/Fields/CollectionView/CollectionViewModel.swift
```diff
@@ -2255,6 +2255,16 @@ extension TableDataViewModelProtocol {
...
     }
+
+    func decoratorsCellWidth() -> CGFloat {
+        if decoratorConfig.visibleLimitInRows <= 1 {
+            return 40
+        } else if decoratorConfig.visibleLimitInRows == 2 {
+            return 80
+        } else {
+            return 100
+        }
```
vivekjyani, line 2266: Nit: the branches are `<= 1 → 40`, `== 2 → 80`, `else → 100`. If `visibleLimitInRows` is ever 0 (e.g. "hide all", or an unconfigured default), the else branch reserves 100pt for a column that renders nothing. Probably impossible in practice today, but `<= 1 → 40`, `== 2 → 80`, `>= 3 → 100`, default `40` would be safer.

JoyfillSwiftUIExample/JoyfillUITests/DecoratorUITests/DecoratorUITestSupport.swift
```diff
@@ -63,8 +63,9 @@ enum DecoratorUITestSupport {
     }
 
     /// Identifier for a single row-level decorator button (shown when exactly one is displayable).
+    /// Row and field buttons currently share the same DecoratorButton identifier format.
     static func rowDecoratorID(action: String) -> String {
-        return "row_decorator_\(action)"
+        return "decorator_button_\(action)"
```
vivekjyani, line 68: Resolving my earlier concern about the row/field identifier collision: field decorators only render inside the row edit form (`.sheet` presentation) and row decorators only render on the table/collection view itself, so the two scopes are never simultaneously in the visible accessibility tree. Sharing `DecoratorButton` and a single `decorator_button_<action>` identifier is intentional and correct — tests query by action in whichever screen they're on. Disregard the suggestion to split the identifiers.

vivekjyani: Correction on the rendering locations: field decorators are rendered in the field title (opposite side of the title), row decorators are rendered inline in each row, and column decorators are rendered in a sheet. Field and row decorators do appear on the same screen, but reusing DecoratorButton with the same identifier is intentional — it's shared component code.

Sources/JoyfillUI/View/Fields/CollectionView/CollectionViewModel.swift
vivekjyani: The first branch already handles this — visibleLimitInRows <= 1 returns 40, so 0 (and negatives) fall into the 40pt branch, not 100pt. The else only fires for >= 3. So the suggested change is functionally identical to what's already there. No change needed.

Sources/JoyfillUI/View/DecoratorView.swift
vivekjyani: Visually verified — the kebab icon and Color.clear are small enough that the 8pt layout overflow is just transparent space, nothing visible bleeds into the next row. Not changing this.

Sources/JoyfillUI/View/DecoratorView.swift
vivekjyani: removed unused code

Sources/JoyfillUI/View/Fields/TableView/TableModalView.swift
```diff
@@ -11,6 +11,22 @@ struct TableRowView : View {
...
+                    viewModel.tableDataModel.documentEditor?.reportDecoratorAction(
+                        fieldIdentifier: viewModel.tableDataModel.fieldIdentifier,
+                        action: decorator.action ?? "",
+                        rowIds: [rowDataModel.rowID]
+                    )
+                }
+                .padding(.all, 4)
+                .frame(width: viewModel.decoratorsCellWidth(), height: 60)
+                .background(Color.rowSelectionBackground(isSelected: isSelected, colorScheme: colorScheme))
+                .border(Color.tableCellBorderColor)
```
vivekjyani, line 28: **Border style inconsistent with sibling data cells.**

This decorator cell uses `.border(Color.tableCellBorderColor)` (default 1pt), but the data cells in the same `LazyHStack` (lines 34-37) use `.overlay(RoundedRectangle(cornerRadius: 0).stroke(Color.tableCellBorderColor, lineWidth: 1.5))`. So in the table the decorator cell will visually have a thinner / different border than the data cells next to it.

This wasn't a problem before the PR because the decorator lived in the sticky `rowsHeader` pane where everything used `.border()`. Now that it sits next to data cells in the scrolling area, it should match them. The header decorator cell (`colsHeader`, around line 295) already matches the data column headers (`.overlay(... .inset(by: 1).stroke(lineWidth: 1.5))`), so only the per-row case is off.

Fix: replace `.border(Color.tableCellBorderColor)` here with the same `.overlay(RoundedRectangle(cornerRadius: 0).stroke(Color.tableCellBorderColor, lineWidth: 1.5))` the sibling cells use.

(The collection sites in `CollectionModelView.swift` are fine — collection cells uniformly use `.border()`, so the decorator already matches there.)

JoyfillSwiftUIExample/JoyfillUITests/DecoratorUITests/Decorator.json
```diff
@@ -795,6 +805,15 @@
       "title" : "Multiple Choice",
       "_id" : "69709dc240f1e385f35abd36",
       "multi" : true
+    },
+    {
+      "title" : "Extra Text",
```
vivekjyani, line 810: **Unused test data.**

This new "Extra Text" field (`_id: 6a1f0c00a1b2c3d4e5f60719`) and its associated field-position entry plus decorator (added at line ~135 above) aren't referenced by any test code — `git grep '6a1f0c00a1b2c3d4e5f60719'` returns hits only inside this JSON file itself.

Either leftover scaffolding from local debugging that should be removed, or a test that exercises this field didn't get added with the JSON. The commit message ("Fix UI tests for decorator") suggests it was added intentionally for some test flow — worth confirming which it is and either wiring up the test or dropping the JSON additions.

vivekjyani: This is required for UI testing. Currently, when scrolling to the bottom, some content is obscured by a button, which is causing issues in our test cases.

Sources/JoyfillUI/View/Fields/TableView/TableModalView.swift
vivekjyani: Verified visually — the existing .border() looks fine in context and matches what the decorator cell had before. Not changing this.

VishnuBishnoi (approved):
## Review — NO-2108: Scale row decorator column width and render inline horizontally

**APPROVED** — clean execution of the decorator-in-scrollable-area migration. All substantive concerns are non-blocking.

---

### What this does

Moves the row decorator column out of the sticky left pane and into the horizontally scrolling data area. The column width steps between 40 / 80 / 100 pt depending on `visibleLimitInRows` (≤1 / 2 / ≥3). Inline rendering replaces the old VStack + `•••` overflow pattern, reusing `DecoratorButton` directly.

---

### Strengths

- **`decoratorsCellWidth()` on the protocol** — single source of truth for width. Every callsite (`colsHeader`, `getWidthForExpanderRow`, `rowWidth`) will automatically stay in sync. Good placement on `TableDataViewModelProtocol`.
- **`DecoratorButton` reuse** — eliminates the separate `singleDecoratorButton` path and the `row_decorator_<action>` accessibility ID. `DecoratorUITestSupport.rowDecoratorID` updated to match.
- **Dead code removal** — `inlineMax`, `inlineVisible`, `hasInlineOverflow`, `singleDecoratorButton` all gone cleanly.
- **`leftColumnWidth`** — correctly stops reserving 40 pt for the decorator column; decorator is now accounted for inside the scrolling area.
- **`getWidthForExpanderRow` parameter** — replacing the hardcoded `40` with `rowDecoratorsCellWidth: CGFloat = 40` is the right direction; default keeps call sites backward-compatible.

---

### Non-blocking notes

**N1 — Kebab button width doesn't adapt to `decoratorsCellWidth()`**

In `RowDecoratorMenuView`, when `exceedsLimit` is true:
```swift
kebabButton
    .frame(width: 40, height: 60)   // ← always 40 pt
```
But the outer cell is `decoratorsCellWidth()` wide — potentially 80 or 100 pt. The kebab ends up centered in a wider cell with wasted space on both sides. Since `RowDecoratorMenuView` doesn't currently receive the cell width, the simplest fix would be replacing the fixed frame with `.frame(maxWidth: .infinity, minHeight: 60)` so the tappable area fills the cell regardless of configured width.

**N2 — `alignment: .center` for row vs `.trailing` for field decorators**

PR #318 used `.trailing` alignment in the `GeometryReader` + `minWidth` pattern for field decorators. Here row decorators use `.center`. Both compile and function correctly; just worth confirming the centering is intentional (row cells are typically narrower and centered content may look better there).

**N3 — No UI test covering the width-scaling path**

The existing decorator tests exercise presence and tapping, but nothing verifies that the column renders at 80 pt when `visibleLimitInRows == 2`. Not a blocker, but a regression risk if `decoratorsCellWidth()` thresholds change later.

---

LGTM overall — the architecture is cleaner after this change. The kebab width (N1) would be worth a follow-up fix.

PR #321, opened 2026-05-01
NO-1695: Unify TableDateView font, single-line filter/bulk-edit layout, and full-pill tap target

vivekjyani:
## Context / Spec
This PR cleans up date rendering in table and collection cells so date values look consistent across formats, read well in compact contexts, and are easier to tap. It addresses three issues:
1. Date text rendered at inconsistent sizes because `minimumScaleFactor` shrank longer formats (e.g. date+time) while shorter ones stayed full size.
2. Date values in the filter search bars and the bulk-edit ("Edit Multiple Rows") sheets needed a compact, single-line presentation.
3. Only the date text was tappable, not the surrounding grey pill.

### Files touched
- `Sources/JoyfillUI/View/Fields/TableView/TableDateView.swift`
- `Sources/JoyfillUI/View/Fields/TableView/SearchBar.swift`
- `Sources/JoyfillUI/View/Fields/TableView/TableModalTopNavigationView.swift`
- `Sources/JoyfillUI/View/Fields/CollectionView/CollectionSearchBar.swift`
- `Sources/JoyfillUI/View/Fields/CollectionView/CollectionModalTopNavigationView.swift`

### Code changes
- Added an `isSingleLineLayout` flag to `TableDateView` that selects between two presentations of a filled date cell:
  - **single line, leading-aligned** — filter search bars + bulk-edit sheets
  - **date/time split onto two lines, centered** — regular table/collection cells
- Standardized date text to a consistent **16pt** font (quick view 15→16) and removed the old `lineLimit(2)` / `minimumScaleFactor(0.7)` / `allowsTightening` / `layoutPriority` that caused different formats to render at different sizes.
- Made the **entire grey date pill tappable** (not just the text): moved the pill's `frame`/`padding`/`background` inside the `Button` label, applied `.contentShape(Rectangle())` last, and added `.buttonStyle(.plain)`.
- Passed `isSingleLineLayout: true` from the table + collection filter search bars and the table + collection "Edit Multiple Rows" bulk-edit sheets.
- Minor layout cleanup around the clear (`xmark`) control and cell width; tighter vertical padding (4pt) for the two-line layout, 8pt for single-line.

## Implementation Decisions
- `isSingleLineLayout` is named for what it controls (layout/alignment), not the caller, since filter and bulk-edit share the same compact presentation. (This replaces the `isUsedForFilter` name used in the earlier version of this PR.)
- The new initializer argument defaults to `false`, so existing call sites and the regular two-line cell layout are unchanged.
- Removing `minimumScaleFactor` (rather than tuning it) is deliberate — scaling was the source of the inconsistent font sizes. Long values now truncate on a single line instead of shrinking.
- No fixed pill width/height: `TableDateView` is reused in places with different sizes, so the pill stays flexible to its container.
- Presentational/layout only; no data-model changes.

## Visuals
- UI-only change. Before/after captures from table + collection date cells (single-line vs two-line) and the tappable pill can be added on request.

## Public API / Docs
- `TableDateView` is an internal view; its initializer gained `isSingleLineLayout` (defaulted, non-breaking). No public/external SDK behavior changes.
- No documentation update required.

## Testing
- Automated tests: not added (UI layout / state-wiring change).
- Manual validation:
  - Table + collection date cells (two-line date/time)
  - Table + collection filter date input (single line)
  - Table + collection "Edit Multiple Rows" date field (single line)
  - Tapping anywhere on the date pill opens the picker
  - Clear (`xmark`) date action and focus/blur behavior in all modes

## Notes For Reviewers
- Focus review on the single-line vs two-line rendering and the full-pill tap target.
- Verify no regression in date clearing, focus/blur, and timezone-aware formatting.
- Note: the `isUsedForFilter` flag from the earlier version was renamed to `isSingleLineLayout`.


Sources/JoyfillUI/View/Fields/TableView/TableDateView.swift
```diff
@@ -56,27 +58,43 @@ struct TableDateView: View {
...
-                                .minimumScaleFactor(0.7)
-                                .allowsTightening(true)
-                                .layoutPriority(1)
+                            Group {
+                                if isSingleLineLayout {
+                                    Text(dateString)
+                                        .darkLightThemeColor()
+                                        .font(.system(size: 16))
+                                } else {
+                                    let parts = dateString.split(separator: " ")
```
vivekjyani, line 68: **Robustness — splitting on the first space is a silent assumption.** The two-line layout treats `dateString` as `<date> <time>` separated by exactly one space: `parts[0]` -> line 1, `parts[1...]` -> line 2. This is correct for *every current* `DateFormatType` because all date components are numeric (`MM/DD/YYYY`, `YYYY-MM-DD`, `DD-MM-YYYY`, ...), so the only space is the date/time boundary. But it's a hidden dependency: if a month-name format (`MMM d, yyyy` -> `Jun 15, 2026`) is ever added, or the `.string`/ISO path in `ValueUnion.dateTime` (which returns `"hh:mm a"` with an internal space) ever reaches this view, the split lands on the wrong token and the date gets chopped across two lines. Worth either a one-line comment documenting the "single space = date/time separator" assumption, or splitting on a known boundary rather than the first space.

**Minor:** `if parts.count >= 1` (line 70) is always true here — we're already inside `if !dateString.isEmpty`, and `split(separator:)` of a non-empty string yields >=1 element. Only the `>= 2` branch does real work; the first guard can be dropped.

Sources/JoyfillUI/View/Fields/TableView/TableDateView.swift
```diff
@@ -56,27 +58,43 @@ struct TableDateView: View {
...
+                                        }
+                                        if parts.count >= 2 {
+                                            Text(parts[1...].joined(separator: " "))
+                                                .darkLightThemeColor()
+                                                .font(.system(size: 16))
+                                        }
+                                    }
+                                }
+                            }
+                            .frame(maxWidth: .infinity, alignment: isSingleLineLayout ? .leading : .center)
```
vivekjyani, line 84: **Confirm intended:** `.frame(maxWidth: .infinity)` makes the grey pill span the full cell width in the regular two-line (non-filter) table/collection cells as well. Previously the pill hugged the date text and a trailing `Spacer()` pushed the xmark to the edge. So this changes the resting appearance of *every* normal date cell, not only the filter/bulk-edit ones. That looks like the deliberate "bigger tap target" goal — just flagging since the PR description frames the full-pill change purely around tappability, and the visual footprint change is broader than that.

vivekjyani (commented):
Automated code review (Claude). Overall this is a clean, well-scoped presentational change — approving in spirit. Two non-blocking notes left inline; details and a public-API assessment are in the PR thread summary.

Sources/JoyfillUI/View/Fields/TableView/TableDateView.swift
vivekjyani: Good catch on both.

The split only works today because all DateFormatTypes are numeric, so the single space is always the date/time boundary — a month-name or hh:mm a string would chop wrong. I'll switch to split(separator: " ", maxSplits: 1) and add a comment noting the assumption. No-space (date-only/time-only) cases already render fine on one line.

And yes, parts.count >= 1 is always true inside !dateString.isEmpty — I'll drop it.

Sources/JoyfillUI/View/Fields/TableView/TableDateView.swift
vivekjyani: yes intended

Sources/JoyfillUI/View/Fields/TableView/TableDateView.swift
```diff
@@ -56,27 +58,43 @@ struct TableDateView: View {
...
-                                .layoutPriority(1)
+                            Group {
+                                if isSingleLineLayout {
+                                    Text(dateString)
+                                        .darkLightThemeColor()
+                                        .font(.system(size: 16))
+                                } else {
+                                    let parts = dateString.split(separator: " ")
+
+                                    VStack(alignment: .leading, spacing: 3) {
```
VishnuBishnoi, line 70: **Accessibility / test-label risk.** Splitting into two stacked `Text`s changes the `Button`'s combined accessibility `.label`. `TimeZoneUITestCases` reads this exact label on `ChangeCellDateIdentifier` cells — both via `app.buttons[<dateTimeString>]` lookups (L437/519/611) and `XCTAssertEqual(cellButton.label, convertedTime)` (L573/1126/…). If SwiftUI joins the two texts with a comma, all of those fail, and `.normalizedSpaces` won't save it (it doesn't touch commas). UI Tests are skipped in CI, so this won't surface there.

Suggest pinning the label explicitly so it's decoupled from the visual layout:
```swift
.accessibilityElement(children: .ignore)
.accessibilityLabel(dateString)
```
That keeps the tests green and makes VoiceOver announce the full date as one phrase.

Sources/JoyfillUI/View/Fields/TableView/TableDateView.swift
VishnuBishnoi, line 68: `split(separator: " ")` assumes the only space separates date from time — true for all current `DateFormatType` patterns, but it'd break for any future format with an intra-date space (e.g. `"Sep 10, 2025 9:00 AM"` → `"Sep"` on line 1, `"10, 2025 9:00 AM"` on line 2). Since the format is known here, deriving date/time parts from the format would be more robust than string-splitting. Non-blocking given today's formats.

Sources/JoyfillUI/View/Fields/TableView/TableDateView.swift
```diff
@@ -56,27 +58,43 @@ struct TableDateView: View {
...
-                            Text(dateString)
-                                .darkLightThemeColor()
-                                .font(.system(size: 16))
-                                .lineLimit(2)
-                                .minimumScaleFactor(0.7)
-                                .allowsTightening(true)
-                                .layoutPriority(1)
+                            Group {
+                                if isSingleLineLayout {
+                                    Text(dateString)
```
VishnuBishnoi, line 63: Single-line branch: with `lineLimit`/`minimumScaleFactor` gone there's nothing forcing one line, so a long date+time will *wrap* and then clip against the fixed-height search bars (`.frame(height: 25)`/`40`). The PR description says these should "truncate on a single line" — add `.lineLimit(1)` (+ `.truncationMode(.tail)`, or `.minimumScaleFactor(...)` if shrink-to-fit is preferred) to actually enforce that.

Sources/JoyfillUI/View/Fields/TableView/TableDateView.swift
VishnuBishnoi, line 84: `.frame(maxWidth: .infinity, …)` (replacing the old content-hugging pill + `Spacer()`) makes the grey pill span the full cell width in regular two-line cells, not just hug the text. Bigger tap target is the goal, but it's a noticeable visual change — worth attaching before/after screenshots for the regular table/collection cell so reviewers can confirm the intended look.

Sources/JoyfillUI/View/Fields/TableView/TableDateView.swift
```diff
@@ -56,27 +58,43 @@ struct TableDateView: View {
...
+                            .contentShape(Rectangle())
                         }
-                        .contentShape(Rectangle())
-                        .padding(.horizontal, 12)
-                        .padding(.vertical, 8)
-                        .background(
-                            RoundedRectangle(cornerRadius: 8, style: .continuous)
-                                .fill(Color(uiColor: .secondarySystemFill))
-                        )
+                        .buttonStyle(.plain)
```
VishnuBishnoi, line 95: 👍 Full-pill tap target done correctly — pill visuals inside the label, `.contentShape(Rectangle())` applied last, `.buttonStyle(.plain)` to avoid the default button tint/highlight. This is the right way to make the whole pill hittable.

VishnuBishnoi (changes requested):
Reviewed in detail. The design direction is good — the full-pill tap target is implemented correctly (frame/padding/background moved inside the `Button` label, `.contentShape(Rectangle())` last, `.buttonStyle(.plain)`), and the consistent 16pt font / removal of `minimumScaleFactor` is a reasonable call. The `isSingleLineLayout` naming + defaulted arg keep existing call sites unchanged.

Requesting changes on one significant risk plus a couple of smaller items:

**1． (Main) The two-line split changes the date cell's accessibility `.label`, which likely breaks the TimeZone UI suite — and CI won't catch it.**
In the regular (non-single-line) cell, `Text(dateString)` becomes a `VStack` of two `Text`s. A `Button` wrapping two stacked `Text`s exposes a *combined* accessibility label, and SwiftUI typically joins stacked texts with a comma (e.g. `"06/15/2026, 10:30AM"`) rather than the single string `"06/15/2026 10:30AM"`. `TimeZoneUITestCases` depends on the old single-string label in several places:
- `app.buttons[convertedTime]` lookups by label — `testTableBlankDateTimeWithDifferentTimezone` (L437–438), `…BulkEdit` (L519–520), `testTableBlankDateTimeWithoutTimezone` (L611–612)
- `XCTAssertEqual(cellButton.label, convertedTime)` on `ChangeCellDateIdentifier` cells (L573, and the same pattern at L1126/1206/1235/1277)

These all use the `MM/dd/yyyy hh:mma` format (date **and** time → two parts → split). The `.normalizedSpaces` helper collapses whitespace but not commas, so a comma-joined label fails both the lookups and the equality checks. Because the **UI Tests CI job is skipped** (see checks), this would merge green and regress silently.

Recommended fix: make the cell's accessibility label explicit so it's independent of the visual split, e.g. on the two-line layout add `.accessibilityElement(children: .ignore)` + `.accessibilityLabel(dateString)` (or just `.accessibilityLabel(dateString)`). That preserves the tests *and* makes VoiceOver read the whole date as one phrase. Please also run `TimeZoneUITestCases` (and the Table/Collection date tests) locally and confirm green before merge.

**2． (Non-blocking) Single-line layout has no `.lineLimit(1)`.** The PR says long values should "truncate on a single line," but with `lineLimit`/`minimumScaleFactor` removed and nothing replacing them, `Text` will *wrap* — then get clipped by the fixed-height filter bars (`.frame(height: 25)` / `40`). Add `.lineLimit(1)` (and `.truncationMode(.tail)`, or `.minimumScaleFactor` if you prefer shrink-to-fit) to match the stated intent.

**3． (Non-blocking, forward-looking) The `split(separator: " ")` two-line approach is format-fragile.** It works for every current `DateFormatType` (single space between date and time), but a future month-name/locale format like `"Sep 10, 2025 9:00 AM"` would split across the wrong lines. Since you already have the format, consider deriving the date and time parts from it rather than splitting on spaces.

Minor visual note inline. No automated tests were added; given the skipped UI Tests job, the regression surface here is unguarded — hence the emphasis on the accessibility label + local verification.

Sources/JoyfillUI/View/Fields/TableView/TableDateView.swift
vivekjyani: Yes right

Sources/JoyfillUI/View/Fields/TableView/TableDateView.swift
vivekjyani: Good catch — that's intentional. Making the pill maxWidth: .infinity is what gives us the larger tap target from the "make the whole date pill tappable" change; the old text-hugging pill + Spacer() left most of the cell non-tappable when a date was present. So in the regular two-line cell the grey pill now fills the cell width and the date is centered in it (single-line/filter/bulk-edit stays left-aligned).

Sources/JoyfillUI/View/Fields/TableView/TableDateView.swift
vivekjyani: Good catch — confirmed. The single-line branch (filter search bar + bulk-edit) lost its line/scale modifiers, so a long date+time can wrap and clip inside the fixed-height field (25/40 pt). We don't want truncation here, so I'll go with shrink-to-fit instead: .lineLimit(1).minimumScaleFactor(0.5) on that branch — keeps it to one line and scales the text down to fit without cutting anything off.

Just to note, this only affects the single-line branch. The regular table/collection cells already handle long dates via the two-line date/time split (the else branch), so no change needed there.

Sources/JoyfillUI/View/Fields/TableView/TableDateView.swift
vivekjyani: Agreed, and thanks for the careful read. Confirmed the split is safe for every current DateFormatType — date parts are slash/dash separated and the time portion has no space (hh:mma → 11:30AM), so the single space is always the date/time boundary. The break case you describe needs a month-name/spaced-AM format, and none exist in the enum today. Since it's non-blocking, I'm keeping the string-split for this PR and have added a comment documenting the "single space = date/time separator" assumption so the constraint is explicit for whoever adds a new format. Deriving the parts from the known format is the right approach if/when such a format lands.

Sources/JoyfillUI/View/Fields/TableView/TableDateView.swift
vivekjyani: Fixed — pinned the label with .accessibilityElement(children: .ignore).accessibilityLabel(dateString) on the pill content, so the accessibility label stays the single date string regardless of the two-line visual split. Verified locally that app.buttons[convertedTime] resolves again and the label == convertedTime assertions pass; also reads as one phrase in VoiceOver now. Thanks for catching it and tracing it to the exact assertions.

VishnuBishnoi: Confirmed resolved — the pill content now carries `.accessibilityElement(children: .ignore)` + `.accessibilityLabel(dateString)`, so the button label is restored to the full date string (identical to the pre-split behavior). That makes `app.buttons[convertedTime]` resolve again and the `label == convertedTime` assertions hold, and it reads as one phrase in VoiceOver. Thanks for the local verification. 👍

Sources/JoyfillUI/View/Fields/TableView/TableDateView.swift
VishnuBishnoi: Heads-up: this one isn't in the current head (`809a420c8`) yet — I re-read the file and there is no `lineLimit`/`minimumScaleFactor` on the single-line branch. You mentioned going with `.lineLimit(1).minimumScaleFactor(0.5)` here; please include it before merge (or confirm you're intentionally skipping it). Non-blocking, so I'm not holding the approval on it.

VishnuBishnoi (approved):
**Approving — re-review of the new changes. My blocking concern is resolved.**

Cross-checked the diff since my last review (`2f9d6e514` → `809a420c8`):

- ✅ **Accessibility / test-label regression (the blocker)** — fixed exactly as suggested: the pill content now has `.accessibilityElement(children: .ignore)` + `.accessibilityLabel(dateString)`, restoring the button's label to the full date string regardless of the two-line visual split. This makes it equivalent to the pre-split behavior, so the `TimeZoneUITestCases` `app.buttons[convertedTime]` lookups and `label == convertedTime` assertions hold again. You also confirmed this locally — appreciated, since the UI Tests CI job is skipped.
- ✅ **Split fragility** — documented with a clear comment noting the single-space date/time assumption and when to revisit. Fine as a non-blocking, forward-looking note (`parts[1...].joined` already avoids any crash).
- ✅ **Full-width pill** — confirmed intentional (the larger tap target is the point of the change).

CI is green (Build + Unit Tests + Syntax) and the PR is mergeable.

**One non-blocking loose end** (replied in-thread): you mentioned adding `.lineLimit(1).minimumScaleFactor(0.5)` to the single-line (filter/bulk-edit) branch so long date+time values shrink instead of wrapping/clipping in the fixed-height fields — but that isn't in the current head yet. Please include it before merge, or confirm you're skipping it. Not holding the approval on it.

PR #322, opened 2026-05-04
NO-2115 Store blank chart points as null in JSON

vivekjyani:
## Context / Spec
Spec: when a coordinate (x or y) on a chart-field point is left blank — i.e. the user has not entered any value, not even zero — the saved JSON should not encode that coordinate as `0`. Today blanks silently coerce to `0`, which is indistinguishable from a real zero and corrupts downstream charting/analytics. The fix omits the coordinate from the JSON entirely when the user hasn't entered a value.

## What Changed
Two source files plus a UI test, all scoped to chart-point handling:

- **`Sources/JoyfillModel/JoyDoc.swift`**
  - `Point.setValue(_ value: CGFloat?, key:)` now removes the key from the dictionary when `value` is `nil` (previously it was a no-op, so clearing a coordinate left the prior value in place). With the key removed, the encoder emits no field at all for that coordinate.
  - `Point.init(id:)` no longer defaults `x` and `y` to `0`; new points start with `nil` coordinates so an untouched, freshly added point serializes without `x` / `y` keys.

- **`Sources/JoyfillUI/View/Fields/ChartView/ChartDetailView.swift`**
  - `setX` / `setY` use `number.map { CGFloat($0.doubleValue) }` instead of `?? 0`, so empty / unparseable input propagates as `nil` rather than being coerced to `0`.
  - `xBinding` / `yBinding` display an empty string when the underlying value is `nil` instead of formatting `0`, so a cleared field reads as blank in the UI.

- **`JoyfillSwiftUIExample/JoyfillUITests/ChartField/ChartFieldUITestCases.swift`**
  - New `testChartCoordinateValuesPersistAndClearToNil` UI test covering write → persist round-trip and clear → key-absent propagation through `onChange`.

## Design Decisions
- **Removing the key (not storing `.null`) when the setter receives `nil`.** The encoder iterates the dictionary, so omitting the key produces no field for that coordinate in the JSON. This is symmetric with `Point.init(from:)`, which uses `decodeIfPresent` and collapses both "missing key" and JSON `null` into "key absent" — meaning encode/decode is a fixed point, and re-saved documents stay stable across round-trips. Storing `.null` would have produced an asymmetric encoder (emits `null`) vs. decoder (drops it back to absent).
- **Fix is in the model setter, not only in the UI.** Doing it in `Point.setValue` makes the invariant universal — any future caller that assigns `point.x = nil` gets the right behavior, instead of every UI site having to remember to clear the dictionary entry manually.
- **Display binding shows empty string for `nil`, not `"0"`.** Otherwise the user sees `0` even though no value is stored, which is misleading and reintroduces the original confusion.
- **No change to `Point`'s public signatures.** Only the internal behavior of `setValue(CGFloat?, key:)` and `init(id:)` changed — both still accept the same arguments and produce a valid `Point`.

## Screenshots / Video
N/A — behavioral / serialization change only; no visual layout difference. Cleared input now reads as blank instead of `0`, which matches what the user typed.

## Public API Impact
No public API surface changes. `Point.init(id:)` and `Point.x` / `Point.y` setters keep the same signatures; only their internal handling of `nil` changes (previously `nil` was silently dropped, now it removes the key from the underlying dictionary). No SDK docs need updating.

## Test Coverage
UI test added: `testChartCoordinateValuesPersistAndClearToNil` in `ChartFieldUITestCases`. It walks the full path:
1. Enter values into the horizontal / vertical point textfields, navigate away, navigate back → values persist correctly.
2. Clear both textfields, navigate away, navigate back → asserts `valueElements.first.points.first.x` and `.y` are both `nil` via the `onChange` payload, and additionally pins down the contract that the cleared coordinates are omitted from the point's dictionary entirely (not stored as `.null` or `.double(0)`).

This covers the regression directly. No additional unit tests were added.

## Reviewer Notes
- The behavior change in `Point.setValue(CGFloat?, key:)` (previously a no-op for `nil`) only has two callers — the `x` and `y` setters — so the blast radius is contained.
- Existing documents that already store explicit `0` coordinates are unaffected; only newly cleared / freshly added blank points have their keys omitted from the JSON.

### DIFF
--- a/JoyfillSwiftUIExample/JoyfillUITests/ChartField/ChartFieldUITestCases.swift
+++ b/JoyfillSwiftUIExample/JoyfillUITests/ChartField/ChartFieldUITestCases.swift
@@ -106,6 +106,48 @@ final class ChartFieldUITestCases: JoyfillUITestsBaseClass {
         XCTAssertEqual(verticalTF.value as? String, "one")
         XCTAssertEqual(horizontalTF.value as? String, "two")
     }
+    
+    func testChartCoordinateValuesPersistAndClearToNil() {
+        goToChartDetailField(index: 0)
+
+        let horizontalPointsValue = app.textFields.matching(identifier: "HorizontalPointsValue").element(boundBy: 0)
+        let verticalPointsValue = app.textFields.matching(identifier: "VerticalPointsValue").element(boundBy: 0)
+
+        XCTAssertTrue(horizontalPointsValue.waitForExistence(timeout: 1), "Horizontal coordinate textfield not found")
+        XCTAssertTrue(verticalPointsValue.waitForExistence(timeout: 1), "Vertical coordinate textfield not found")
+
+        horizontalPointsValue.tap()
+        horizontalPointsValue.clearText()
+        horizontalPointsValue.typeText("11")
+
+        verticalPointsValue.tap()
+        verticalPointsValue.clearText()
+        verticalPointsValue.typeText("22")
+        app.dismissKeyboardIfVisible()
+
+        goBack()
+
+        goToChartDetailField(index: 0)
+        let horizontalAfterWrite = app.textFields.matching(identifier: "HorizontalPointsValue").element(boundBy: 0)
+        let verticalAfterWrite = app.textFields.matching(identifier: "VerticalPointsValue").element(boundBy: 0)
+
+        XCTAssertEqual(horizontalAfterWrite.value as? String, "11", "Horizontal coordinate did not persist after navigation")
+        XCTAssertEqual(verticalAfterWrite.value as? String, "22", "Vertical coordinate did not persist after navigation")
+
+        horizontalAfterWrite.tap()
+        horizontalAfterWrite.clearText()
+        verticalAfterWrite.tap()
+        verticalAfterWrite.clearText()
+
+        goBack()
+
+        goToChartDetailField(index: 0)
+        let horizontalAfterClear = app.textFields.matching(identifier: "HorizontalPointsValue").element(boundBy: 0)
+        let verticalAfterClear = app.textFields.matching(identifier: "VerticalPointsValue").element(boundBy: 0)
+
+        XCTAssertNil(onChangeResultValue().valueElements?.first?.points?.first?.x, "Horizontal coordinate should be nil after clearing")
+        XCTAssertNil(onChangeResultValue().valueElements?.first?.points?.first?.y, "Vertical coordinate should be nil after clearing")
+    }
 
     func testChartFieldOnFocusAndOnBlur() {
         goToChartDetailField(index: 0)
--- a/JoyfillSwiftUIExample/JoyfillUITests/ChartField/ChartFieldUITestCases.swift
+++ b/JoyfillSwiftUIExample/JoyfillUITests/ChartField/ChartFieldUITestCases.swift
@@ -147,6 +147,14 @@ final class ChartFieldUITestCases: JoyfillUITestsBaseClass {
 
         XCTAssertNil(onChangeResultValue().valueElements?.first?.points?.first?.x, "Horizontal coordinate should be nil after clearing")
         XCTAssertNil(onChangeResultValue().valueElements?.first?.points?.first?.y, "Vertical coordinate should be nil after clearing")
+
+        // The getter returns nil for both `.null` and "key absent". Pin down the
+        // actual contract — cleared coordinates must be omitted from the point's
+        // dictionary entirely (not stored as .null or .double(0)).
+        let clearedPoint = onChangeResultValue().valueElements?.first?.points?.first
+        XCTAssertNotNil(clearedPoint, "Point should still exist after clearing coordinates")
+        XCTAssertNil(clearedPoint?.dictionary["x"], "x key should be absent from point dictionary, not stored as null or 0")
+        XCTAssertNil(clearedPoint?.dictionary["y"], "y key should be absent from point dictionary, not stored as null or 0")
     }
 
     func testChartFieldOnFocusAndOnBlur() {
--- a/Sources/JoyfillModel/JoyDoc.swift
+++ b/Sources/JoyfillModel/JoyDoc.swift
@@ -1656,7 +1656,7 @@ public struct Point: Codable,Hashable, Equatable {
     ///   - key: The key to set the value for.
     mutating func setValue(_ value: CGFloat?, key: String) {
         guard let value = value else {
-            self.dictionary[key] = .null
+            self.dictionary[key] = nil
             return
         }
         self.dictionary[key] = .double(value)
--- a/Sources/JoyfillModel/JoyDoc.swift
+++ b/Sources/JoyfillModel/JoyDoc.swift
@@ -1594,7 +1594,10 @@ public struct Point: Codable,Hashable, Equatable {
         }
         let allKeys = container.allKeys
         for key in allKeys {
-            dictionary[key.stringValue] = try container.decodeIfPresent(ValueUnion.self, forKey: key)
+            // Use `decode` rather than `decodeIfPresent` so JSON `null` round-trips
+            // as `ValueUnion.null` instead of being silently dropped. `allKeys` only
+            // contains keys present in the payload, so `decode` cannot throw "missing key".
+            dictionary[key.stringValue] = try container.decode(ValueUnion.self, forKey: key)
         }
     }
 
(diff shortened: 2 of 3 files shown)


vivekjyani (commented):
Review comments inline. Main concern: test assertions and the implementation disagree on whether a cleared coordinate is stored as `.null` or omitted from the dictionary entirely.

Sources/JoyfillModel/JoyDoc.swift
```diff
@@ -1656,6 +1656,7 @@ public struct Point: Codable,Hashable, Equatable {
     ///   - key: The key to set the value for.
     mutating func setValue(_ value: CGFloat?, key: String) {
         guard let value = value else {
+            self.dictionary[key] = nil
```
vivekjyani, line 1659: **Test/implementation contradiction.** This writes `.null` into the dictionary on `nil`, but the new test asserts `XCTAssertNil(clearedPoint?.dictionary["x"])` — i.e. the key must be absent. After `setValue(nil, key:)`, `dictionary["x"]` is `.some(.null)`, not `nil`, so the test should fail on the in-memory point.

Pick one contract:
- **Store `.null`** (matches the PR description's stated design): keep this line and change the test assertions to `XCTAssertEqual(dict["x"], .null)`. Also un-revert `bbf29e1e` so `Point.init(from:)` preserves `.null` on decode — otherwise encode→decode is asymmetric and a re-saved doc loses the `null`.
- **Remove the key** (matches the test's comment): change this to `self.dictionary[key] = nil` and update the PR description. This is symmetric with the current decoder (`decodeIfPresent` already collapses missing/null into absent), so encode/decode is a fixed point.

Secondary nit on this same overload: the `String?` `setValue` just above is a silent no-op on `nil`, while this `CGFloat?` overload now writes `.null`. Two setters with the same shape have opposite behavior — `point.label = nil` does nothing, `point.x = nil` writes `null`. Worth aligning the two so future callers aren't surprised.

JoyfillSwiftUIExample/JoyfillUITests/ChartField/ChartFieldUITestCases.swift
```diff
@@ -106,6 +106,56 @@ final class ChartFieldUITestCases: JoyfillUITestsBaseClass {
...
+
+        horizontalAfterWrite.tap()
+        horizontalAfterWrite.clearText()
+        verticalAfterWrite.tap()
+        verticalAfterWrite.clearText()
+
+        goBack()
+
+        goToChartDetailField(index: 0)
+        let horizontalAfterClear = app.textFields.matching(identifier: "HorizontalPointsValue").element(boundBy: 0)
```
vivekjyani, line 145: Nit: `horizontalAfterClear` / `verticalAfterClear` are bound but never read — only `onChangeResultValue()` is asserted against. Either drop these two lines or use them to assert that the textfields render as empty strings after clear, which would also exercise the new empty-string branch in `xBinding` / `yBinding`.

JoyfillSwiftUIExample/JoyfillUITests/ChartField/ChartFieldUITestCases.swift
```diff
@@ -106,6 +106,56 @@ final class ChartFieldUITestCases: JoyfillUITestsBaseClass {
...
+        let verticalPointsValue = app.textFields.matching(identifier: "VerticalPointsValue").element(boundBy: 0)
+
+        XCTAssertTrue(horizontalPointsValue.waitForExistence(timeout: 1), "Horizontal coordinate textfield not found")
+        XCTAssertTrue(verticalPointsValue.waitForExistence(timeout: 1), "Vertical coordinate textfield not found")
+
+        horizontalPointsValue.tap()
+        horizontalPointsValue.clearText()
+        horizontalPointsValue.typeText("11")
+
+        verticalPointsValue.tap()
```
vivekjyani, line 123: Nit: keyboard isn't dismissed between the horizontal and vertical inputs — only once after both at L126. On iPhone form factors the keyboard can occlude the second textfield and cause flake. Consider `app.dismissKeyboardIfVisible()` between the two `tap()`/`typeText()` blocks, or guard each tap with a `hittable` check.

Sources/JoyfillModel/JoyDoc.swift
vivekjyani: Going with option 2 — removed the key instead of storing `.null`. Pushed in 71814d23: `setValue(CGFloat?, key:)` now does `self.dictionary[key] = nil`, which makes encode/decode symmetric (decoder's `decodeIfPresent` already collapses null/missing into absent), so re-saved docs stay stable. The existing test assertions (`XCTAssertNil(dict["x"])`) match this contract, and the PR description has been updated.

On the secondary nit: leaving the `String?` overload alone for now since this PR is scoped to chart points — happy to align both setters in a follow-up if preferred.

JoyfillSwiftUIExample/JoyfillUITests/ChartField/ChartFieldUITestCases.swift
vivekjyani: Acknowledged — leaving as-is for now. Haven't seen flake on the form factors this suite runs against, and the existing `app.dismissKeyboardIfVisible()` after the second input has been reliable. If we start seeing intermittent failures on iPhone, will add the intermediate dismiss then.

JoyfillSwiftUIExample/JoyfillUITests/ChartField/ChartFieldUITestCases.swift
vivekjyani: Acknowledged — leaving as-is for now. The model-level assertions via `onChangeResultValue()` already cover the contract this PR is scoped to (cleared coordinate → `nil` in the model, key absent from the dictionary). The UI binding's empty-string branch is straightforward enough that a model-level guarantee is sufficient here. Happy to tighten the test in a follow-up if we see the binding regress.

VishnuBishnoi (approved):
## Review — NO-2115: Store blank chart points as null in JSON

**APPROVED** — the fix is correct end-to-end. Key-absent serialization is the right contract here.

---

### What this does

Clears the `x = 0` / `y = 0` defaults from `Point.init(id:)` (now `nil`) and ensures that clearing a coordinate field removes the key from the backing dictionary entirely rather than storing `0` or JSON `null`.

---

### Correctness analysis

**`Point.init(id:)` — nil defaults**
Freshly created points now serialize without `x` / `y` keys at all. On decode, `decodeIfPresent` maps a missing key → `nil`, so the encode/decode round-trip is a fixed point. Correct.

**`Point.setValue(nil, key:)` — key removal**
```swift
guard let value = value else {
    self.dictionary[key] = nil   // ← removes the key in Swift dictionaries
    return
}
```
`[String: ValueUnion][key] = nil` removes the entry from the dictionary. This is the right approach and is symmetric with `decodeIfPresent` on the decode side. Storing JSON `null` (the previous attempt that was reverted) would have been technically decodable but inconsistent with the "key absent = no value" contract the rest of the model relies on.

**`xBinding` / `yBinding` — early-return on nil**
```swift
guard let x = point.x else { return "" }
```
Idiomatic and correct — cleared field shows an empty text field, and the number formatter is never called with a forced `0`.

**`setX` / `setY` — nil propagation**
```swift
point.x = number.map { CGFloat($0.doubleValue) }
```
Clean replacement for `CGFloat(number?.doubleValue ?? 0)`. When the field is empty (`number == nil`), `x` is set to `nil` and the key will be removed by `setValue`. Correct.

**New UI test — `testChartCoordinateValuesPersistAndClearToNil`**
Write 11/22 → navigate away → verify persistence → clear both → verify key-absent in `onChangeResultValue()`. Good coverage of the core bug.

---

### Non-blocking notes

**N1 — Cleared text fields not asserted in the UI layer**

`horizontalAfterClear` and `verticalAfterClear` are declared but never have any assertions made against them:
```swift
let horizontalAfterClear = app.textFields.matching(identifier: "HorizontalPointsValue").element(boundBy: 0)
let verticalAfterClear   = app.textFields.matching(identifier: "VerticalPointsValue").element(boundBy: 0)
// ← no XCTAssert on these two
```
Adding `XCTAssertEqual(horizontalAfterClear.value as? String, "")` and `XCTAssertEqual(verticalAfterClear.value as? String, "")` would close the loop between the model assertion (key absent) and the UI assertion (field shows empty string).

**N2 — `onChangeResultValue()` called four times**

The test calls it once per `XCTAssertNil` + once for `clearedPoint`. Confirm this is a stored-result getter (idempotent) rather than one that replays the change handler each call. If it replays, consolidate into a single `let result = onChangeResultValue()` at the top of the clear-assertions block.

**N3 — Chart render test with nil coordinates missing**

No test verifies that the chart view renders gracefully when `x` or `y` is `nil` on an existing persisted point (e.g. loaded from a document where one coordinate was cleared). Low priority, but worth tracking as a follow-up.

---

The core fix is solid and the serialization contract is now internally consistent. N1 is worth a follow-up commit.

PR #323, opened 2026-05-05
NO-2124: Add decorator support for pencil and pen-to-square icons

vivekjyani:
## 1) Spec Context
This change adds support for two additional decorator icon IDs (`pencil` and `pen-to-square`) so they render consistently in the SDK and are also available in the example app icon picker.

## 2) What Changed
- Added icon ID handling in [`DecoratorView`](Sources/JoyfillUI/View/DecoratorView.swift) for:
  - asset mapping: `decorator_pencil`, `decorator_pen_to_square`
  - SF Symbol fallback mapping: `pencil`, `square.and.pencil`
- Added new asset image sets:
  - `Sources/JoyfillUI/Resources/Assets.xcassets/decorator_pencil.imageset/*`
  - `Sources/JoyfillUI/Resources/Assets.xcassets/decorator_pen_to_square.imageset/*`
- Updated demo icon catalogue in:
  - `JoyfillSwiftUIExample/JoyfillExample/Decorator Example/DecoratorAPIDemoView.swift`

## 3) Implementation Decisions
- Used dedicated XCAsset image sets with template rendering intent so these icons follow existing tint/theming behavior.
- Kept SF Symbol fallback mappings aligned with the same IDs for resilient rendering behavior.

## 4) Visual Evidence
- Screenshot/video: not attached in this PR from CLI.
- The example app (`DecoratorAPIDemoView`) includes both IDs for quick visual verification.

## 5) Public API / Docs
- No public API surface change.
- Behaviorally, supported decorator `icon` values now include `pencil` and `pen-to-square`.
- If external docs list supported icon IDs, they should be updated to include these two values.

## 6) Testing
- No new automated tests added in this PR (asset + mapping addition).
- Manual verification path is available via the example app icon picker.

## 7) Reviewer Notes
- Suggested review focus: icon ID mapping correctness and asset naming consistency.
- This PR intentionally keeps scope minimal (no unrelated refactors).


Sources/JoyfillUI/View/DecoratorView.swift
```diff
@@ -49,6 +49,8 @@ enum DecoratorIcon {
         case "plus":         return "decorator_plus"
         case "print":        return "decorator_print"
         case "flag":         return "decorator_flag"
+        case "pencil":       return "decorator_pencil"
+        case "pen-to-square": return "decorator_pen_to_square"
```
vivekjyani, line 53: Formatting nit — the surrounding cases pad return values into a fixed column (e.g. line 52 `case "pencil":       return ...`). `"pen-to-square"` is the longest key in this switch, so the single space here breaks the column alignment that the rest of the switch maintains. Either add extra padding on this line (and line 86) so the column lines up, or re-pad the rest of the switch to accommodate the new max-width key. Cosmetic only; happy to leave as-is if you'd rather not churn the surrounding lines.

Sources/JoyfillUI/View/DecoratorView.swift
```diff
@@ -80,6 +82,8 @@ enum DecoratorIcon {
         case "plus":         return "plus"
         case "print":        return "printer.fill"
         case "flag":         return "flag.fill"
+        case "pencil":       return "pencil"
+        case "pen-to-square": return "square.and.pencil"
```
vivekjyani, line 86: Same column-alignment nit as line 53 — `pen-to-square` breaks the formatted column. Worth fixing in tandem with the bundledAssetName switch so both stay symmetrical.

JoyfillSwiftUIExample/JoyfillExample/Decorator Example/DecoratorAPIDemoView.swift
```diff
@@ -41,6 +41,8 @@ private let iconCatalogue: [IconOption] = [
     .init(id: "circle-info", symbol: "info.circle.fill"),
     .init(id: "filter",      symbol: "line.3.horizontal.decrease.circle.fill"),
     .init(id: "paper-plane", symbol: "paperplane.fill"),
+    .init(id: "pencil",      symbol: "pencil"),
+    .init(id: "pen-to-square", symbol: "square.and.pencil"),
```
vivekjyani, line 45: Same column-alignment nit — the longer `"pen-to-square"` ID has only one space before `symbol:` while all sibling rows are aligned to a fixed column. Cosmetic, but if you're touching this file anyway it's a one-character fix.

JoyfillSwiftUIExample/JoyfillExample/Decorator Example/DecoratorAPIDemoView.swift
```diff
@@ -41,6 +41,8 @@ private let iconCatalogue: [IconOption] = [
...
 ]
```
vivekjyani, line 46: Pre-existing, not introduced by this PR, but worth flagging while we're here: the comment on line 18 claims `iconCatalogue` *matches* the SDK's `DecoratorIcon` mapping. Even after this PR, the demo is still missing `import`, `rotate`, `folder-open`, `magnet`, and `add` — all of which are mapped in `DecoratorView.swift`. Either sync the catalogue (so the demo can exercise every supported ID) or soften the comment to say the catalogue is a curated subset. Not blocking this PR; could be a follow-up.

vivekjyani (commented):
Clean, scoped change — adds two new icon IDs (`pencil`, `pen-to-square`) consistently to both the bundled-asset and SF Symbol fallback paths in the SDK, plus the example app's icon picker. The Contents.json structure for the new imagesets matches the existing pattern exactly (template rendering intent, 1x SVG with empty 2x/3x slots).

A few minor inline notes below — all nits, none blocking. The behavior change itself is straightforward and the asset attribution (Font Awesome 7.2.0 license header) is preserved.

Sources/JoyfillUI/View/DecoratorView.swift
vivekjyani: Acknowledged — leaving as-is. Re-padding the rest of the switch just to accommodate one longer key would churn unrelated lines and bloat the diff for a purely cosmetic gain. The single-space form is unambiguous and the switch still reads cleanly. Will keep this in mind if we ever do a broader formatting pass on this file.

Sources/JoyfillUI/View/DecoratorView.swift
vivekjyani: Same as above — keeping it consistent with line 53 (single-space) rather than re-padding the whole switch. No functional impact and the diff stays minimal.

JoyfillSwiftUIExample/JoyfillExample/Decorator Example/DecoratorAPIDemoView.swift
vivekjyani: Leaving as-is to stay consistent with the SDK switches above. Cosmetic-only and not worth re-aligning the surrounding rows for.

JoyfillSwiftUIExample/JoyfillExample/Decorator Example/DecoratorAPIDemoView.swift
vivekjyani: Good catch on the pre-existing drift, but out of scope for this PR — this change is intentionally narrow (just adding pencil / pen-to-square). Happy to sync the demo catalogue with the full SDK mapping (and update the line 18 comment) in a separate follow-up.

VishnuBishnoi (approved):
## Review — NO-2124: Add decorator support for pencil and pen-to-square icons

**APPROVED** — minimal, well-scoped PR that follows existing patterns exactly. One factual issue worth fixing before the next batch of icon additions.

---

### What this does

Adds `pencil` and `pen-to-square` as valid decorator `icon` values. Each gets:
- A named asset (`decorator_pencil` / `decorator_pen_to_square`) with an SVG at 1x and an SF Symbol fallback.
- An entry in the demo app icon picker for manual visual verification.

---

### Correctness

**`DecoratorIcon.assetName(for:)`**
```swift
case "pencil":        return "decorator_pencil"
case "pen-to-square": return "decorator_pen_to_square"
```
Naming is consistent with every existing entry in the switch. ✓

**`DecoratorIcon.sfSymbolName(for:)`**
```swift
case "pencil":        return "pencil"
case "pen-to-square": return "square.and.pencil"
```
Both `pencil` and `square.and.pencil` are valid SF Symbols available since iOS 13. ✓

**`Contents.json` structure**
Both new imagesets are byte-for-byte identical in structure to existing ones (`decorator_paper_plane`, `decorator_camera`, etc.): 1x filename, empty 2x/3x entries, `template-rendering-intent: template`. Perfect structural consistency. ✓

**Demo catalogue**
```swift
.init(id: "pencil",      symbol: "pencil"),
.init(id: "pen-to-square", symbol: "square.and.pencil"),
```
Uses the correct SF Symbol names in the demo picker. ✓

---

### Issue — Font Awesome version mismatch (non-blocking, but worth fixing)

Every existing decorator SVG in the repo is from **Font Awesome Free 6.5.2** (Copyright 2024):
```
<!--! Font Awesome Free 6.5.2 by @fontawesome ... Copyright 2024 Fonticons, Inc. -->
```

The two new SVGs are from **Font Awesome Free 7.2.0** (Copyright 2026):
```
<!--! Font Awesome Free 7.2.0 by @fontawesome ... Copyright 2026 Fonticons, Inc. -->
```

This is the only inconsistency in the entire PR. Path data between major versions can differ subtly (stroke weight, viewBox proportions, corner radii), which could make these two icons look slightly off compared to the rest of the set. Not a correctness bug, but visible in side-by-side comparisons.

**Recommendation:** Source the `pencil` and `pen-to-square` SVGs from FA 6.5.2 to stay consistent with the existing 20+ icons, OR open a follow-up ticket to migrate all decorator SVGs to FA 7.x together. Don't mix versions within the same icon family.

---

### Minor notes

**N1 — No automated tests**
Acknowledged in the PR. Since no test exists for `DecoratorIcon.assetName(for:)` or `sfSymbolName(for:)` for any icon, this is a pre-existing gap, not introduced here. Worth tracking as a follow-up — a unit test on the two `switch` functions would prevent silent regressions when new IDs are added.

**N2 — SVG files have no trailing newline**
Both new SVGs end without `\n`. All existing decorator SVGs in the repo follow the same convention (single-line SVGs), so this is consistent — not a new problem.

---

LGTM with the version note above. The code change itself is exactly right; the only ask is aligning the SVG source version with the rest of the asset catalogue.

PR #324, opened 2026-05-05
NO-2120 Update schema for chart x/y nullability and image multi flag

vivekjyani:
## 1. Context / Spec
The bundled `joyfill-schema.swift` was out of sync with the latest backend schema for chart points and image columns. Chart points need to support sparse data (missing or null x/y), and image columns (image-type cells in table/collection fields) need a `multi` flag to indicate multi-image support. Bumping the embedded schema version to reflect the change.

## 2. What Changed
**File:** `Sources/JoyfillUI/View/joyfill-schema.swift`

- **Chart point** (`x`, `y`):
  - Type changed from `"number"` to `["number", "null"]`
  - Removed from `required` array — only `_id` is now required
- **ImageColumn** (image-type column in table/collection):
  - Added optional `multi: boolean` property
- **Schema version**:
  - `$joyfillSchemaVersion` bumped `1.0.0` → `1.0.1`

**File:** `JoyfillSwiftUIExample/JoyfillTests/SchemaValidation/SchemaValidationTests.swift`

- Updated `expectedSchemaVersion` constant from `"1.0.0"` to `"1.0.1"` to match the bumped schema (resolves 49 test failures in the `SchemaValidationTests` suite — all of them were asserting against this single constant).

## 3. Design Decisions
- Used JSON Schema's union type `["number", "null"]` rather than dropping the type constraint, so non-null values are still validated as numbers.
- Kept `_id` required on chart points to preserve point identity even when coordinates are absent.
- `multi` is optional (not in `required`) on `ImageColumn` to remain backwards-compatible with existing column definitions.
- Bumped version to **1.0.1** (patch) because all changes are backwards-compatible relaxations: relaxing `required` and adding optional fields. `JoyfillSchemaManager.validateVersion` only compares major versions, so this bump flows through without code changes and old documents continue to validate.

## 4. Screenshots / Video
N/A — schema-only change, no UI impact.

## 5. Public API Impact
No public Swift API changes. `joyfillSchema` is a `public var` string constant; its value is updated but the symbol/signature is unchanged. Swift models already support the new shapes — `Point.x`/`Point.y` are already `CGFloat?`, and `multi: Bool?` already exists on the column-data types in `JoyDoc.swift`. No docs update required.

## 6. Test Coverage
- Existing `SchemaValidationTests` suite (49 tests) continues to pass after the version-constant update; nothing in those tests depends on chart x/y being required or on `ImageColumn` lacking `multi`, so the relaxations don't break them.
- No new positive-coverage tests added for the relaxed shapes (chart point with `x: null` / missing `x`/`y`, ImageColumn with `multi: true`). Can be added as a follow-up if desired.

## 7. Reviewer Notes
- Diff is small and localized; please verify the schema delta matches the canonical backend schema source.
- The `null` allowance on `x`/`y` means downstream chart rendering must already tolerate missing coordinates — verified that current consumers (`ChartDetailView`, formula context in `JoyfillDocContext`) treat them as optional, and no force-unwraps exist in `ChartView/`.
- Version bump is **patch-level** (1.0.0 → 1.0.1); no migration or consumer impact expected.
- The fallback `documentVersion ?? "1.0.0"` in `JoyfillSchemaManager.swift:69` is intentional ("treat undefined as v1.x.x") and was deliberately left as-is — major-version comparison makes the literal irrelevant.

### DIFF
--- a/JoyfillSwiftUIExample/JoyfillTests/SchemaValidation/SchemaValidationTests.swift
+++ b/JoyfillSwiftUIExample/JoyfillTests/SchemaValidation/SchemaValidationTests.swift
@@ -12,7 +12,7 @@ import Joyfill
 
 final class SchemaValidationTests: XCTestCase {
     /// The schema version expected in `SchemaValidationError.details`.
-    private let expectedSchemaVersion = "1.0.0"
+    private let expectedSchemaVersion = "1.0.1"
     private let sdkVersion = "3.0.0-rc17"
     
     func documentEditor(document: JoyDoc) -> DocumentEditor {
--- a/JoyfillSwiftUIExample/JoyfillTests/SchemaValidation/SchemaValidationTests.swift
+++ b/JoyfillSwiftUIExample/JoyfillTests/SchemaValidation/SchemaValidationTests.swift
@@ -341,6 +341,39 @@ final class SchemaValidationTests: XCTestCase {
             XCTAssertEqual(expectedSchemaVersion, err?.details.schemaVersion)
         }
     }
+
+    // MARK: - Schema 1.0.1 relaxations
+
+    // ChartPoint: x/y are now optional and may be null. Only _id is required.
+    func testChartPoint_WithNullOrMissingCoordinates_ShouldValidate() {
+        let cases: [(String, [String: Any])] = [
+            ("null x and y",     ["_id": "p1", "x": NSNull(), "y": NSNull()]),
+            ("missing x and y",  ["_id": "p2"]),
+            ("null x only",      ["_id": "p3", "x": NSNull(), "y": 5]),
+            ("missing y only",   ["_id": "p4", "x": 1])
+        ]
+        for (label, point) in cases {
+            var docDict = minimalValidDocumentDictionary()
+            let chartField: [String: Any] = [
+                "_id": "chart1",
+                "type": "chart",
+                "file": "file1",
+                "xTitle": "X", "yTitle": "Y",
+                "xMin": 0, "xMax": 100,
+                "yMin": 0, "yMax": 100,
+                "value": [
+                    [
+                        "_id": "series1",
+                        "points": [point]
+                    ]
+                ]
+            ]
+            docDict["fields"] = [chartField]
+            let doc = JoyDoc(dictionary: docDict)
+            let err = JoyfillSchemaManager().validateSchema(document: doc)
+            XCTAssertNil(err, "ChartPoint with \(label) should validate against the relaxed schema")
+        }
+    }
 }
 
 private class MockFormChangeEvent: FormChangeEvent {
--- a/Sources/JoyfillUI/View/joyfill-schema.swift
+++ b/Sources/JoyfillUI/View/joyfill-schema.swift
@@ -1877,6 +1877,9 @@ public var joyfillSchema = """
             "type": "string",
             "minLength": 1
           },
+          "multi": {
+            "type": "boolean"
+          },
           "type": {
             "type": "string",
             "const": "image"
@@ -2257,16 +2260,14 @@ public var joyfillSchema = """
             "type": "string"
           },
           "y": {
-            "type": "number"
+            "type": ["number", "null"]
           },
           "x": {
-            "type": "number"
+            "type": ["number", "null"]
           }
         },
         "required": [
-          "_id",
-          "y",
-          "x"
+          "_id"
         ]
       },
       "CollectionField": {
--- a/Sources/JoyfillUI/View/joyfill-schema.swift
+++ b/Sources/JoyfillUI/View/joyfill-schema.swift
@@ -2594,7 +2594,7 @@ public var joyfillSchema = """
         ]
       }
     },
-    "$joyfillSchemaVersion": "1.0.0"
+    "$joyfillSchemaVersion": "1.0.1"
   }
 
 """


Sources/JoyfillUI/View/joyfill-schema.swift
```diff
@@ -1877,6 +1877,9 @@ public var joyfillSchema = """
             "type": "string",
             "minLength": 1
           },
+          "multi": {
+            "type": "boolean"
+          },
```
vivekjyani, line 1882: **No positive test coverage for the relaxations.** The PR description acknowledges this. A typo here (or in the `["number", "null"]` union below for chart x/y) would go uncaught — existing `SchemaValidationTests` only assert the version string, not that a doc using the new shape validates.

Two cheap additions would close the gap:
- A doc with `points: [{ "_id": "p1", "x": null, "y": null }]` validates successfully.
- An `ImageColumn` with `"multi": true` validates successfully.

Sources/JoyfillUI/View/joyfill-schema.swift
```diff
@@ -2257,16 +2260,14 @@ public var joyfillSchema = """
...
-            "type": "number"
+            "type": ["number", "null"]
           }
         },
         "required": [
-          "_id",
-          "y",
-          "x"
+          "_id"
         ]
```
vivekjyani, line 2271: **`required: ["_id"]` — intentional?** With both `x` and `y` now optional, a point with only `_id` (no label, no coordinates) is valid. Confirm this matches the canonical backend schema. If not, consider keeping `label` required, or adding an `anyOf` constraint that at least one of `x`/`y`/`label` must be present — otherwise this allows completely empty data points.

Sources/JoyfillUI/View/joyfill-schema.swift
```diff
@@ -2593,7 +2594,7 @@ public var joyfillSchema = """
...
       }
     },
-    "$joyfillSchemaVersion": "1.0.0"
+    "$joyfillSchemaVersion": "1.0.1"
```
vivekjyani, line 2597: **Semver: should this be 1.1.0 instead of 1.0.1?**

Adding an optional property and relaxing `required` are *consumer-side* backwards-compatible — but a v1.0.1 document containing `x: null` will fail strict validation against a v1.0.0 schema in any tool that doesn't restrict comparison to major version (other SDKs, backend validators, external consumers).

The "major-only comparison" argument in the PR description holds for this SDK (`JoyfillSchemaManager.validateVersion`), but the bump signals nothing to other ecosystems. Low priority — worth confirming with the backend team that 1.0.x patch convention is what they expect.

JoyfillSwiftUIExample/JoyfillTests/SchemaValidation/SchemaValidationTests.swift
```diff
@@ -12,7 +12,7 @@ import Joyfill
 
 final class SchemaValidationTests: XCTestCase {
     /// The schema version expected in `SchemaValidationError.details`.
-    private let expectedSchemaVersion = "1.0.0"
+    private let expectedSchemaVersion = "1.0.1"
```
vivekjyani, line 15: **Duplicate `SchemaValidationTests` class in the test target.**

There's another file at `JoyfillSwiftUIExample/JoyfillTests/SchemaValidationTests.swift:7` that also defines `final class SchemaValidationTests: XCTestCase` and was not updated by this PR.

Pre-existing, not introduced here — but if both are in the test target's build phase, this is a name collision (non-deterministic test discovery or link error). Worth confirming which file is canonical and either removing or renaming the other while you're touching this file.

vivekjyani (commented):
Review of the schema bump — most issues inline below.

One issue that can't be inlined because the file isn't in this PR's diff:

### Formula context silently coerces null x/y to 0 (potential bug)

`Sources/JoyfillUI/ViewModels/JoyfillDocContext.swift:974-975`

```swift
structuredPoint["x"] = pointDict["x"] ?? .number(0)
structuredPoint["y"] = pointDict["y"] ?? .number(0)
```

The PR description claims formula context "treats them as optional," but missing keys fall back to `.number(0)`. With the schema now permitting sparse points, formulas like `SUM`/`AVERAGE` over chart points will silently arithmetic against 0 for missing data, skewing results.

Note: an explicit `.null` value in the dictionary survives `??` (only a missing key triggers it), so behavior depends on whether the source produces missing keys vs. explicit nulls. Worth either changing the fallback to `.null` or documenting why `0` is the correct default for the formula path.

Sources/JoyfillUI/View/joyfill-schema.swift
vivekjyani: _id was the only required field before this change too — we only removed x and y. Confirmed the backend's canonical schema also requires only _id; intentionally not adding an anyOf since the SDK schema mirrors backend and an empty point is harmless to the renderer.

Sources/JoyfillUI/View/joyfill-schema.swift
vivekjyani: Thanks for flagging. We've decided to go with 1.0.1 for this change — the version is set to mirror the canonical backend schema, so it's intentional on our side. Will keep the minor-vs-patch point in mind for future bumps.

JoyfillSwiftUIExample/JoyfillTests/SchemaValidation/SchemaValidationTests.swift
vivekjyani: Verified — only the file under JoyfillTests/SchemaValidation/ is in the test target's build phase; the other isn't compiled, so there's no actual collision. Pre-existing and out of scope for this PR — leaving as-is.

Sources/JoyfillUI/View/joyfill-schema.swift
vivekjyani: Unit test added

vivekjyani: testChartPoint_WithNullOrMissingCoordinates_ShouldValidate — parameterized over null x/y, missing x/y, and mixed cases, so a typo in the ["number","null"] union or a regression on required would be caught.

VishnuBishnoi (approved):
## Review — NO-2120: Update schema for chart x/y nullability and image multi flag

**APPROVED** — clean, correctly scoped schema sync. All changes are backwards-compatible, align with the Swift model that already exists, and are well-tested.

---

### Schema changes (`joyfill-schema.swift`)

**Chart point `x` / `y` — nullability**
```json
"y": { "type": ["number", "null"] }   // was: "number"
"x": { "type": ["number", "null"] }   // was: "number"
"required": ["_id"]                    // was: ["_id", "y", "x"]
```
Correct on all fronts:
- `["number", "null"]` is standard JSON Schema syntax for a union type. Non-null values are still constrained to numbers — the type guard is preserved, not dropped.
- Removing `x`/`y` from `required` while keeping `_id` required preserves point identity for sparse data. Consistent with PR #322 (`Point.init(id:)` now defaults `x`/`y` to `nil`, `Point.setValue(nil, key:)` removes the key).
- Round-trip is clean: key-absent → `decodeIfPresent` → `nil` in the model → `setValue(nil)` removes key → key-absent in JSON → validates against the relaxed schema.

**`ImageColumn` — `multi` flag**
```json
"multi": { "type": "boolean" }   // new, optional (not in required)
```
Simple, backwards-compatible addition. `multi: Bool?` already exists in the Swift model; the schema was just behind. Not in `required`, so existing column definitions without the key continue to validate.

**Version bump: `1.0.0` → `1.0.1`**
Appropriate patch-level bump — all changes are backwards-compatible relaxations (widened types, added optional field). Confirmed that `JoyfillSchemaManager.validateVersion` only compares major versions (`if majorVersion > supportedMajorVersion`), so the bump is transparent to all existing document validation paths. The `documentVersion ?? "1.0.0"` fallback on line 69 remains correct; major-version comparison makes the exact literal irrelevant.

---

### Test changes (`SchemaValidationTests.swift`)

**`expectedSchemaVersion` update**
`"1.0.0"` → `"1.0.1"` — necessary mechanical update; the 49 affected tests all assert against this one constant and are otherwise unchanged. ✓

**New test: `testChartPoint_WithNullOrMissingCoordinates_ShouldValidate`**
Four parameterised cases:

| Label | x | y |
|---|---|---|
| `null x and y` | `NSNull()` | `NSNull()` |
| `missing x and y` | absent | absent |
| `null x only` | `NSNull()` | `5` |
| `missing y only` | `1` | absent |

Covers all combinations of the two orthogonal axes (null vs. absent, x vs. y). The loop label is propagated into the `XCTAssertNil` message, which makes failures immediately actionable. Pattern is consistent with the existing suite's `minimalValidDocumentDictionary()` usage. ✓

---

### Non-blocking notes

**N1 — No negative test for the retained type constraint**
The union type `["number", "null"]` still rejects non-number, non-null values (e.g. `x: "string"`, `x: true`). There's no test asserting that the old guard is still in place. A single case like:
```swift
("invalid string x", ["_id": "p5", "x": "not-a-number"]),
```
with `XCTAssertNotNil(err, "...")` would lock down the contract and prevent accidental future widening to a bare `{}` type.

**N2 — `multi: true` on ImageColumn not tested**
Acknowledged in the PR description and deferred. Low risk given the property is a simple optional boolean with no logic attached to it in the schema. Fine as a follow-up.

**N3 — `JoyfillSchemaManager()` instantiated per loop iteration**
Minor: the test creates a fresh `JoyfillSchemaManager()` in every `for` iteration body. Hoisting it to `let manager = JoyfillSchemaManager()` before the loop is a trivial cleanup with no functional impact.

---

Schema delta matches the stated intent, aligns with the existing Swift model, and the test coverage is solid for the primary change. LGTM.

PR #325, opened 2026-05-06
NO-2127: Show decorators, tooltip, and required indicator on readonly forms and when title is hidden

vivekjyani:
## 1. Context
On readonly forms (`mode == .readonly` or `field.disabled == true`), the field header was being disabled wholesale by SwiftUI's `.disabled()` propagation, which made decorators non-tappable and hid them in table/collection bulk-edit sheets. Additionally, when `titleDisplay == "none"`, the entire `FieldHeaderView` was skipped, so decorators, the required asterisk, and the tooltip button never rendered even though they should still be visible/interactive.

## 2. What changed
- **[FieldHeaderView.swift](Sources/JoyfillUI/View/FieldHeaderView.swift)** — Always render the header `HStack`; title is now an optional child instead of gating the whole view. Decorators, required asterisk, and tooltip render regardless of title visibility. Added `.environment(\.isEnabled, true)` on `FieldDecoratorsView` so decorators stay tappable even when the parent field is `.disabled()`. Removed the `model.mode == .fill` guard around decorators.
- **[DocumentEditor.swift](Sources/JoyfillUI/ViewModels/DocumentEditor.swift)** — Always build a `FieldHeaderModel`; when `titleDisplay == "none"`, pass `title: nil` instead of skipping the model entirely so decorators/tooltip/required still surface.
- **[TableModalTopNavigationView.swift](Sources/JoyfillUI/View/Fields/TableView/TableModalTopNavigationView.swift)** & **[CollectionModalTopNavigationView.swift](Sources/JoyfillUI/View/Fields/CollectionView/CollectionModalTopNavigationView.swift)** — Removed `mode == .fill` gating in the bulk-edit sheet so decorators show on readonly tables/collections.
- **[TableViewModel.swift](Sources/JoyfillUI/View/Fields/TableView/TableViewModel.swift)** & **[CollectionViewModel.swift](Sources/JoyfillUI/View/Fields/CollectionView/CollectionViewModel.swift)** — `showRowDecorators` no longer requires `.fill` mode.

## 3. Why it's done this way
- Used `.environment(\.isEnabled, true)` rather than restructuring the `.disabled()` chain at every call site — this is a localized override scoped to decorator buttons inside the header, keeps the rest of the readonly behavior (inputs disabled) intact, and avoids touching the public field-rendering API.
- Kept building `FieldHeaderModel` unconditionally and made `title` optional, instead of introducing a separate "header-without-title" code path. Smaller diff, single source of truth for header content.
- Mode (`.fill`/`.readonly`) is still tracked on the model; only the rendering gates were removed, so future consumers that need mode-aware behavior aren't blocked.

## 4. Screenshot / video
N/A — behavioral fix; visually the header now appears in readonly mode and when titles are hidden, with decorators tappable. Happy to attach a clip if needed.

## 5. Public API
No public API changes. `FieldHeaderModel.title` was already `String?`; existing initializers and call sites are unchanged. No docs update required.

## 6. Tests
- Added [DecoratorAPIUITests.swift](JoyfillSwiftUIExample/JoyfillUITests/DecoratorUITests/DecoratorAPIUITests.swift) covering decorator tap interactions in readonly mode.
- Existing decorator/header UI tests continue to pass.
- Manual verification on readonly forms, `titleDisplay == "none"` fields, and table/collection bulk-edit sheets.

## 7. Notes for reviewer
- The `.environment(\.isEnabled, true)` override is intentionally narrow — only on the `FieldDecoratorsView` inside the header. If you'd prefer a different mechanism (e.g., lifting the `.disabled()` boundary), happy to discuss.
- Decorator visibility is now driven purely by `decorators.isEmpty` and `isDisplayable` — `mode` is no longer a gate.

### DIFF
--- a/JoyfillSwiftUIExample/JoyfillUITests/DecoratorUITests/DecoratorAPIUITests.swift
+++ b/JoyfillSwiftUIExample/JoyfillUITests/DecoratorUITests/DecoratorAPIUITests.swift
@@ -29,6 +29,9 @@ private let cL2ColBarcode    = "69e8827e72fb2149cfd944a4"
 
 // Text field
 private let fpTextField = "6970918d350238d0738dd5c9"
+private let fpNumberField = "69709192a4e9c0c281851275"
+private let fpDateField = "6970919451f1adb30fcff9ea"
+private let fpDropdownField = "6970919a6ce635a6b14913ed"
 
 // MARK: - Decorator base (disables animations for all decorator tests)
 
@@ -779,5 +782,99 @@ final class DecoratorCollectionAPIUITests: DecoratorAPIUITestsBase {
         S.openCollectionNestedRowEditForm(rowIndex: 1, boundBy: 1, in: app)
         XCTAssertTrue(waitUntil(2) { !flag.exists })
         S.dismissRowEditForm(in: app)
+    }
+}
+
+// MARK: - Readonly Decorator Coverage
+
+final class DecoratorReadonlyAPIUITests: DecoratorAPIUITestsBase {
+
+    override func getJSONFileNameForTest() -> String { "Decorator" }
+
+    override func getGotoLaunchArguments() -> [(String, String?)] {
+        return [("--disable-animations", nil), ("--mode", "readonly")]
+    }
+
+    private func slightSwipeUp() {
+        let start = app.windows.firstMatch.coordinate(withNormalizedOffset: CGVector(dx: 0.5, dy: 0.72))
+        let end = app.windows.firstMatch.coordinate(withNormalizedOffset: CGVector(dx: 0.5, dy: 0.58))
+        start.press(forDuration: 0.01, thenDragTo: end)
+    }
+
+    private func addTapAndAssertFocus(path: String, action: String) {
+        typealias S = DecoratorUITestSupport
+        let button = app.buttons[S.fieldDecoratorID(action: action)]
+
+        S.run(S.addCommand(path: path,
+                           decorators: [S.decorator(action: action, icon: "flag", label: "Flag")]),
+              in: app)
+
+        XCTAssertTrue(waitUntil(3) { button.exists && button.isHittable })
+        button.tap()
+
+        XCTAssertTrue(waitUntil(2) {
+            S.decoratorFocusEvent(action: action, in: self.focusBlurOptionalResults()) != nil
+        })
+
+        S.run(S.removeCommand(path: path, action: action), in: app)
+    }
+
+    func testReadonly_TextDecorator_Tappable() {
+        addTapAndAssertFocus(path: "\(pageID)/\(fpTextField)", action: "ro_text")
+    }
+
+    func testReadonly_NumberDecorator_Tappable() {
+        addTapAndAssertFocus(path: "\(pageID)/\(fpNumberField)", action: "ro_number")
+    }
+
+    func testReadonly_DateDecorator_Tappable() {
+        addTapAndAssertFocus(path: "\(pageID)/\(fpDateField)", action: "ro_date")
+    }
+
+    func testReadonly_DropdownDecorator_Tappable() {
+        slightSwipeUp()
+        addTapAndAssertFocus(path: "\(pageID)/\(fpDropdownField)", action: "ro_dropdown")
+    }
+
+    func testReadonly_TableRowDecorator_Tappable() {
+        typealias S = DecoratorUITestSupport
+        let action = "ro_table_row"
+        let rowsPath = "\(pageID)/\(fpTable)/rows"
+
+        S.openTableDetailView(in: app)
+        let rowButton = app.buttons.matching(identifier: S.rowDecoratorID(action: action)).element(boundBy: 0)
+
+        S.run(S.addCommand(path: rowsPath,
+                           decorators: [S.decorator(action: action, icon: "flag", label: "Flag")]),
+              in: app)
+        XCTAssertTrue(waitUntil(3) { rowButton.exists && rowButton.isHittable })
+
+        rowButton.tap()
+        XCTAssertTrue(waitUntil(2) {
+            S.decoratorFocusEvent(action: action, in: self.focusBlurOptionalResults()) != nil
+        })
+
+        S.run(S.removeCommand(path: rowsPath, action: action), in: app)
+    }
+
+    func testReadonly_CollectionRowDecorator_Tappable() {
+        typealias S = DecoratorUITestSupport
+        let action = "ro_collection_row"
+        let rootRowsPath = "\(pageID)/\(fpCollection)/rows"
+
+        S.openCollectionDetailView(in: app)
+        let rowButton = app.buttons.matching(identifier: S.rowDecoratorID(action: action)).element(boundBy: 0)
+
+        S.run(S.addCommand(path: rootRowsPath,
+                           decorators: [S.decorator(action: action, icon: "flag", label: "Flag")]),
+              in: app)
+        XCTAssertTrue(waitUntil(3) { rowButton.exists && rowButton.isHittable })
+
+        rowButton.tap()
+        XCTAssertTrue(waitUntil(2) {
+            S.decoratorFocusEvent(action: action, in: self.focusBlurOptionalResults()) != nil
+        })
+
+        S.run(S.removeCommand(path: rootRowsPath, action: action), in: app)
     }
 }
--- a/Sources/JoyfillUI/View/FieldHeaderView.swift
+++ b/Sources/JoyfillUI/View/FieldHeaderView.swift
@@ -20,53 +20,53 @@ struct FieldHeaderView: View {
     }
     
     var body: some View {
-        if let title = fieldHeaderModel?.title {
-            HStack(alignment: .center, spacing: 4) {
+        HStack(alignment: .center, spacing: 4) {
+            if let title = fieldHeaderModel?.title {
                 Text("\(title)")
                     .font(.headline.bold())
-                
-                if fieldHeaderModel?.required == true {
-                    Image(systemName: "asterisk")
-                        .foregroundColor(isFilled ? .gray : .red)
-                        .imageScale(.small)
+            }
+            
+            if fieldHeaderModel?.required == true {
+                Image(systemName: "asterisk")
+                    .foregroundColor(isFilled ? .gray : .red)
+                    .imageScale(.small)
+            }
+            
+            Spacer()
+            
+            if let model = fieldHeaderModel,
+               !model.decorators.isEmpty {
+                FieldDecoratorsView(
+                    decorators: model.decorators,
+                    visibleLimit: model.visibleLimitInFields
+                ) { decorator in
+                    onDecoratorTap?(decorator)
                 }
-
-                Spacer()
-
-                if let model = fieldHeaderModel,
-                   !model.decorators.isEmpty {
-                    FieldDecoratorsView(
-                        decorators: model.decorators,
-                        visibleLimit: model.visibleLimitInFields
-                    ) { decorator in
-                        onDecoratorTap?(decorator)
-                    }
-                    // Re-enable the header even when the parent field is .disabled() —
-                    // decorators must stay interactive on readonly forms.
-                    .environment(\.isEnabled, true)
-                }
-                let tipDescription = fieldHeaderModel?.tipDescription ?? ""
-                let tipTitle = fieldHeaderModel?.tipTitle ?? ""
-                if let tipVisible = fieldHeaderModel?.tipVisible {
-                    if tipVisible == true && !(tipDescription.isEmpty && tipTitle.isEmpty) {
-                        Button(action: {
-                            if let tipTitle = fieldHeaderModel?.tipTitle,
-                               let tipDescription = fieldHeaderModel?.tipDescription {
-                                alertMessage = tipTitle
-                                alertDescription = tipDescription
-                                showAlert = true
-                            }
-                        }, label: {
-                            Image(systemName: "i.circle")
-                        })
-                        .accessibilityIdentifier("ToolTipIdentifier")
-                        .alert(isPresented: $showAlert) {
-                            Alert(
-                                title: Text(alertMessage ?? ""),
-                                message: Text(alertDescription ?? ""),
-                                dismissButton: .default(Text("Dismiss"))
-                            )
+                // Re-enable the header even when the parent field is .disabled() —
+                // decorators must stay interactive on readonly forms.
+                .environment(\.isEnabled, true)
+            }
+            let tipDescription = fieldHeaderModel?.tipDescription ?? ""
+            let tipTitle = fieldHeaderModel?.tipTitle ?? ""
+            if let tipVisible = fieldHeaderModel?.tipVisible {
+                if tipVisible == true && !(tipDescription.isEmpty && tipTitle.isEmpty) {
+                    Button(action: {
+                        if let tipTitle = fieldHeaderModel?.tipTitle,
+                           let tipDescription = fieldHeaderModel?.tipDescription {
+                            alertMessage = tipTitle
+                            alertDescription = tipDescription
+                            showAlert = true
                         }
+                    }, label: {
+                        Image(systemName: "i.circle")
+                    })
+                    .accessibilityIdentifier("ToolTipIdentifier")
+                    .alert(isPresented: $showAlert) {
+                        Alert(
+                            title: Text(alertMessage ?? ""),
+                            message: Text(alertDescription ?? ""),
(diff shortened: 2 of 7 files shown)


Sources/JoyfillUI/View/FieldHeaderView.swift
```diff
@@ -20,51 +20,53 @@ struct FieldHeaderView: View {
     }
     
     var body: some View {
-        if let title = fieldHeaderModel?.title {
-            HStack(alignment: .center, spacing: 4) {
+        HStack(alignment: .center, spacing: 4) {
```
vivekjyani, line 23: **Empty header HStack adds vertical space when nothing is rendered.**

This `HStack` now renders unconditionally. Previously the `if let title = fieldHeaderModel?.title` guard caused the view to return nothing when title was nil.

`FieldHeaderView` is rendered inside a parent `VStack` in every field (e.g. `TextView`, `NumberView`, `DropdownView`, `MultiLineTextView`, etc.). For a field with `titleDisplay == "none"` AND no required indicator AND no decorators AND no tooltip, the always-on `HStack` is still counted as a real subview, so the VStack's default spacing applies — adding a visible gap above the input where there used to be none.

Suggest gating the `HStack` on "any header content present" (title || required || non-empty decorators || a valid tooltip) and returning `EmptyView()` otherwise.

Sources/JoyfillUI/View/FieldHeaderView.swift
```diff
@@ -20,51 +20,53 @@ struct FieldHeaderView: View {
...
-                        .accessibilityIdentifier("ToolTipIdentifier")
-                        .alert(isPresented: $showAlert) {
-                            Alert(
-                                title: Text(alertMessage ?? ""),
-                                message: Text(alertDescription ?? ""),
-                                dismissButton: .default(Text("Dismiss"))
-                            )
+                // Re-enable the header even when the parent field is .disabled() —
+                // decorators must stay interactive on readonly forms.
+                .environment(\.isEnabled, true)
```
vivekjyani, line 48: **Tooltip button is not given the same `.environment(\.isEnabled, true)` override.**

Decorators are explicitly re-enabled here, but the tooltip `Button` below (line 53) is not. On a readonly/disabled field the parent `.disabled()` will propagate and the tooltip will render but be non-tappable.

The commit message says "tool tip stays same as earlier," so this may be intentional — but a visible `i.circle` button that does nothing on tap is a worse UX than not showing it. Either apply the same `.environment(\.isEnabled, true)` override to the tooltip Button, or hide the tooltip entirely on readonly forms.

Sources/JoyfillUI/ViewModels/DocumentEditor.swift
```diff
@@ -657,7 +657,9 @@ extension DocumentEditor {
         let fieldEditMode: Mode = ((fieldData?.disabled == true) || (mode == .readonly) ? .readonly : .fill)
         let decorators = fieldData?.decorators?.filter({ $0.isDisplayable }).map(DecoratorLocal.init(from:)) ?? []
         
-        var fieldHeaderModel = (fieldPosition.titleDisplay == nil || fieldPosition.titleDisplay != "none") ? FieldHeaderModel(title: fieldData?.title, required: fieldData?.required, tipDescription: fieldData?.tipDescription, tipTitle: fieldData?.tipTitle, tipVisible: fieldData?.tipVisible, decorators: decorators, mode: fieldEditMode, visibleLimitInFields: decoratorConfig.visibleLimitInFields) : nil
+        let showTitle = (fieldPosition.titleDisplay == nil || fieldPosition.titleDisplay != "none")
+        
+        var fieldHeaderModel = FieldHeaderModel(title: showTitle ? fieldData?.title : nil, required: fieldData?.required, tipDescription: fieldData?.tipDescription, tipTitle: fieldData?.tipTitle, tipVisible: fieldData?.tipVisible, decorators: decorators, mode: fieldEditMode, visibleLimitInFields: decoratorConfig.visibleLimitInFields)
```
vivekjyani, line 662: **`FieldHeaderModel.mode` is now dead state.**

Every `model.mode == .fill` gate was removed in this PR, but `mode: fieldEditMode` is still passed into the model here (and again at line 836), and `FieldHeaderModel` still stores it. Nothing reads it anymore.

Either drop the `mode` property + initializer arg, or add a comment explaining why it's preserved. The PR description mentions "future consumers that need mode-aware behavior" — but YAGNI; the property can come back when actually needed and removing it now keeps the model honest.

Sources/JoyfillUI/View/Fields/TableView/TableModalTopNavigationView.swift
```diff
@@ -237,13 +237,13 @@ struct EditMultipleRowsSheetView: View {
...
-            if viewModel.tableDataModel.mode == .fill {
-                let decorators = viewModel.tableDataModel.getTableCellDecorators(rowIds: viewModel.tableDataModel.selectedRows, columnId: col.id ?? "")
-                if !decorators.isEmpty {
-                    FieldDecoratorsView(decorators: decorators, visibleLimit: viewModel.decoratorConfig.visibleLimitInFields) { decorator in
-                        viewModel.tableDataModel.documentEditor?.reportDecoratorAction(fieldIdentifier: viewModel.tableDataModel.fieldIdentifier, action: decorator.action ?? "", rowIds: viewModel.tableDataModel.selectedRows, columnId: col.id)
-                    }
+            
+            let decorators = viewModel.tableDataModel.getTableCellDecorators(rowIds: viewModel.tableDataModel.selectedRows, columnId: col.id ?? "")
+            if !decorators.isEmpty {
+                FieldDecoratorsView(decorators: decorators, visibleLimit: viewModel.decoratorConfig.visibleLimitInFields) { decorator in
```
vivekjyani, line 245: **Bulk-edit decorators may need the same `.environment(\.isEnabled, true)` override.**

The `mode == .fill` gate was dropped so decorators now render in readonly bulk-edit, but they don't get the same isEnabled override that `FieldHeaderView` got. Worth confirming the bulk-edit sheet isn't wrapped by a propagating `.disabled()` from a readonly parent — otherwise decorators will render but be non-tappable, repeating the bug this PR fixes elsewhere.

Also: can multi-row selection / bulk edit even be entered on a readonly table? If not, this whole code path is unreachable in readonly and the change is dead. Please confirm reachability — otherwise the parallel change in `CollectionModalTopNavigationView.swift` is also unused.

(Same comment applies to `CollectionEditMultipleRowsSheetView` in `CollectionModalTopNavigationView.swift`.)

JoyfillSwiftUIExample/JoyfillUITests/DecoratorUITests/DecoratorAPIUITests.swift
```diff
@@ -29,6 +29,9 @@ private let cL2ColBarcode    = "69e8827e72fb2149cfd944a4"
 
 // Text field
 private let fpTextField = "6970918d350238d0738dd5c9"
+private let fpNumberField = "69709192a4e9c0c281851275"
+private let fpDateField = "6970919451f1adb30fcff9ea"
+private let fpDropdownField = "6970919a6ce635a6b14913ed"
```
vivekjyani, line 34: **Verify these field IDs exist in the `Decorator` JSON fixture.**

`fpNumberField`, `fpDateField`, `fpDropdownField` are new constants used by the four readonly tests below. If any are stale/wrong the tests will fail with a generic "button not hittable" timeout rather than a clear "field not found" error, which makes future failures hard to diagnose. Worth a quick grep against the JSON to confirm.

JoyfillSwiftUIExample/JoyfillUITests/DecoratorUITests/DecoratorAPIUITests.swift
```diff
@@ -781,3 +784,97 @@ final class DecoratorCollectionAPIUITests: DecoratorAPIUITestsBase {
...
+
+    override func getJSONFileNameForTest() -> String { "Decorator" }
+
+    override func getGotoLaunchArguments() -> [(String, String?)] {
+        return [("--disable-animations", nil), ("--mode", "readonly")]
+    }
+
+    private func slightSwipeUp() {
+        let start = app.windows.firstMatch.coordinate(withNormalizedOffset: CGVector(dx: 0.5, dy: 0.72))
+        let end = app.windows.firstMatch.coordinate(withNormalizedOffset: CGVector(dx: 0.5, dy: 0.58))
```
vivekjyani, line 800: Minor: `slightSwipeUp()` is only used in `testReadonly_DropdownDecorator_Tappable`. If the dropdown is below the fold here but the existing fill-mode dropdown decorator tests don't need a swipe, those existing tests may be exercising a different code path or a different field instance. A one-line comment noting why only this readonly test needs the swipe would help future readers.

vivekjyani (commented):
Inline review comments below. Core fix looks sound; flagging a few concerns and minor cleanups.

Sources/JoyfillUI/View/FieldHeaderView.swift
vivekjyani: When the header has any content (title, required, decorators, or tooltip), the HStack renders with non-zero height and everything looks correct — including the case you're calling out (title hidden + decorators present). No issue there.

Sources/JoyfillUI/View/FieldHeaderView.swift
vivekjyani: No need to enable tool tip for now

Sources/JoyfillUI/View/Fields/TableView/TableModalTopNavigationView.swift
vivekjyani: We do not support bulk edit and row form opening in read only mode

JoyfillSwiftUIExample/JoyfillUITests/DecoratorUITests/DecoratorAPIUITests.swift
vivekjyani: Yes exists

Sources/JoyfillUI/ViewModels/DocumentEditor.swift
vivekjyani: preserved for future mode-aware logic

JoyfillSwiftUIExample/JoyfillUITests/DecoratorUITests/DecoratorAPIUITests.swift
vivekjyani: The dropdown field happens to fall below the fold on the test device in this readonly fixture layout — the existing fill-mode dropdown tests use a different page/scroll position so they don't hit it. Helper is intentionally narrow (`slightSwipeUp` does what it says on the tin), so I'd rather not add a comment that'll go stale if the fixture layout changes. Happy to add one if you feel strongly.

VishnuBishnoi (approved):
## Review — NO-2127: Show decorators, tooltip, and required indicator on readonly forms and when title is hidden

**APPROVED** — the approach is correct and the core fix works. One gap between the PR's stated goals and the implementation needs a follow-up, flagged as C1 below.

---

### Architecture overview

The PR makes two orthogonal fixes:

1. **`titleDisplay == "none"` gate removal** — `DocumentEditor.swift` now always builds a `FieldHeaderModel`, passing `title: nil` when the display is suppressed. `FieldHeaderView` moves the title into an optional `if let` child inside the `HStack` instead of gating the entire view on it. Decorators, required asterisk, and tooltip now render regardless of title visibility.

2. **`mode == .fill` gate removal** — Strips the `.fill` guard from `FieldHeaderView`, `TableViewModel.showRowDecorators`, `CollectionViewModel.showRowDecorators(forSchemaKey:)`, and both modal bulk-edit top navigation views. The `.environment(\.isEnabled, true)` on `FieldDecoratorsView` counteracts the `isEnabled = false` injected by `.disabled(listModel.fieldEditMode == .readonly)` in `FormView`.

---

### What's correct

**`DocumentEditor.swift` — always-build model**
```swift
let showTitle = (fieldPosition.titleDisplay == nil || fieldPosition.titleDisplay != "none")
var fieldHeaderModel = FieldHeaderModel(title: showTitle ? fieldData?.title : nil, ...)
```
Clean ternary, single model path, `mode` still threaded through for future consumers. Both occurrences (lines ~657 and ~831) updated consistently. ✓

**`FieldHeaderView.swift` — HStack always rendered**
Moving the `if let title` check inside the `HStack` is the right structural change. The `.padding(.bottom, fieldHeaderModel?.title == nil ? 8 : 0)` on `FieldDecoratorsView` is a thoughtful UX touch — adds breathing room between the decorator row and field content when there's no title above it. ✓

**`.environment(\.isEnabled, true)` on `FieldDecoratorsView`**
`FormView` applies `.disabled(listModel.fieldEditMode == .readonly)` to every field view, which injects `isEnabled = false` into the full subtree including `FieldHeaderView`. Overriding that key on just `FieldDecoratorsView` is the correct surgical approach — re-enables the decorator buttons without touching the surrounding readonly behavior. ✓

**`TableViewModel` / `CollectionViewModel` — `showRowDecorators`**
Single-line removals of the `&& tableDataModel.mode == .fill` clause. Clean. ✓

**Modal top navigation views**
Removing the `if viewModel.tableDataModel.mode == .fill` wrapper from both `EditMultipleRowsSheetView` and `CollectionEditMultipleRowsSheetView` is straightforward. ✓

---

### C1 — Tooltip button missing `.environment(\.isEnabled, true)` (notable gap)

The PR description explicitly lists "tooltip" as a goal:
> *"decorators, tooltip, and required indicator on readonly forms"*

But in the updated `FieldHeaderView`, `.environment(\.isEnabled, true)` is applied **only** to `FieldDecoratorsView`:

```swift
if let model = fieldHeaderModel, !model.decorators.isEmpty {
    FieldDecoratorsView(...)
        .environment(\.isEnabled, true)   // ← decorators ✓
}
// ...
Button(action: { ... }, label: { Image(systemName: "i.circle") })
    .accessibilityIdentifier("ToolTipIdentifier")
    // ← no .environment(\.isEnabled, true) here ✗
```

Since `FieldHeaderView` sits inside a `.disabled(fieldEditMode == .readonly)` scope, the tooltip `Button` inherits `isEnabled = false` and is **rendered but not tappable** in readonly mode. The fix is one line:

```swift
Button(action: { ... }) { ... }
    .accessibilityIdentifier("ToolTipIdentifier")
    .environment(\.isEnabled, true)   // add this
    .alert(...)
```

---

### C2 — Bulk-edit sheet `FieldDecoratorsView` not re-enabled (minor)

In `EditMultipleRowsSheetView` and `CollectionEditMultipleRowsSheetView`, the `FieldDecoratorsView` for bulk-selected cells is now shown without a mode gate, but also without `.environment(\.isEnabled, true)`. If the sheet is accessible in readonly mode (depends on whether row-selection is also ungated), those decorators would be visible but non-tappable — same root cause as C1. Lower priority since the sheet may not be reachable in readonly mode, but worth confirming.

---

### Non-blocking notes

**N1 — Required asterisk without title is visually orphaned**
When `titleDisplay == "none"` and `required == true`, the HStack renders: `[asterisk] [Spacer] [decorators?]`. A lone asterisk at the far left with no label text next to it is contextually ambiguous. Consider either suppressing the asterisk when there's no title, or keeping it adjacent to the decorator area on the right side.

**N2 — No test for `titleDisplay == "none"` decorator rendering**
The new `DecoratorReadonlyAPIUITests` class tests readonly mode well (text, number, date, dropdown, table row, collection row). But the other half of the fix — "decorators still render when title is hidden" — has no automated coverage. A test that sets a field's title display to "none" and asserts the decorator button is still hittable would close this gap.

**N3 — `slightSwipeUp()` uses raw normalized window coordinates**
```swift
let start = app.windows.firstMatch.coordinate(withNormalizedOffset: CGVector(dx: 0.5, dy: 0.72))
let end   = app.windows.firstMatch.coordinate(withNormalizedOffset: CGVector(dx: 0.5, dy: 0.58))
```
Fragile against layout changes or device size variations. Using a named scroll container identifier + `scrollToElement` or a `swipeUp()` on the identified scroll view would be more resilient.

---

### Summary

| Area | Status |
|---|---|
| `titleDisplay == "none"` always builds `FieldHeaderModel` | ✅ Correct |
| `FieldHeaderView` HStack always rendered | ✅ Correct |
| `.environment(\.isEnabled, true)` on `FieldDecoratorsView` | ✅ Correct mechanism |
| `mode == .fill` guard removed from view models + modal sheets | ✅ Correct |
| Tooltip button re-enabled on readonly | ❌ Missing — C1 |
| Readonly decorator UI tests (6 cases) | ✅ Good coverage |

LGTM on the core approach. The tooltip gap (C1) is a one-line fix against a stated requirement.

PR #326, opened 2026-05-11
NO-2131: Reset duplicated mobile-view page controls + add UI tests

vivekjyani:
## Context / Spec
This PR addresses `NO-2131` by aligning duplicated **mobile-view** pages with expected default controls after duplication. In mobile view, a duplicated page should remain manageable and re-duplicable with both duplicate modes available.

## What Changed (Files / Code)
  - In `duplicatePage(...)`, when duplicating an alternate (mobile) page, the duplicated alternate page is now explicitly reset to:
    - `deletable = true`
    - `copyable = [.withValues, .withoutValues]`
- `JoyfillSwiftUIExample/JoyfillUITests/PageDeletableCopyable/PageDeletableCopyableMobileViewUITests.swift`
  - Added UI coverage for mobile-view duplication behavior and post-duplication delete behavior.
- `JoyfillSwiftUIExample/JoyfillUITests/PageDeletableCopyable/PageDeletableCopyableMobileView.json`
  - Added dedicated fixture used by mobile-view UI tests.
- `JoyfillSwiftUIExample/JoyfillExample.xcodeproj/project.pbxproj`
  - Registered the new fixture and new UI test file in the test target/resources.

## Key Implementation Decision
We reset `deletable` and `copyable` on the duplicated alternate page at duplication time rather than inheriting potentially restrictive source values. This guarantees duplicated mobile-view pages have predictable, spec-aligned controls regardless of original-page state.

## Screenshot / Video
No screenshot/video attached for this PR since the change is primarily behavior + automated UI test coverage in the page navigation sheet. Happy to attach a short recording if reviewers want visual confirmation.

## Public API / Docs Impact
- Public API changes: **None**
- Documentation updates needed: **No**

## Test Coverage
- Added UI tests:
  - `testDuplicateInMobileViewProducesPageWithDeletableAndFullCopyable`
  - `testDeleteDuplicatedPageInMobileViewRemovesIt`
- Coverage note: This PR covers the mobile-view duplicate/delete flows described above.

## Reviewer Notes
- Focus review on `duplicatePage(...)` alternate-page behavior in `DocumentEditor` and the new mobile-view UI test expectations.
- Change is intentionally small and scoped to `NO-2131` behavior.

### DIFF
--- a/JoyfillSwiftUIExample/JoyfillUITests/PageDeletableCopyable/PageDeletableCopyableMobileView.json
+++ b/JoyfillSwiftUIExample/JoyfillUITests/PageDeletableCopyable/PageDeletableCopyableMobileView.json
@@ -1,0 +1,236 @@
+{
+  "_id": "69fda955b580cf6fab953a4e",
+  "type": "template",
+  "stage": "published",
+  "identifier": "doc_pdc_mobile_001",
+  "name": "PDC Mobile",
+  "createdOn": 1758527982824,
+  "files": [
+    {
+      "_id": "file_pdc_mobile_001",
+      "metadata": {},
+      "name": "",
+      "version": 1,
+      "pages": [
+        {
+          "_id": "page_pdc_mobile_001",
+          "name": "Web Only Page",
+          "deletable": true,
+          "copyable": [
+            "with-values",
+            "without-values"
+          ],
+          "fieldPositions": [],
+          "metadata": {},
+          "hidden": false,
+          "width": 816,
+          "height": 1056,
+          "cols": 8,
+          "rowHeight": 8,
+          "layout": "grid",
+          "presentation": "normal",
+          "margin": 0,
+          "padding": 24,
+          "borderWidth": 0
+        }
+      ],
+      "pageOrder": [
+        "page_pdc_mobile_001"
+      ],
+      "views": [
+        {
+          "_id": "view_pdc_mobile_001",
+          "type": "mobile",
+          "pages": [
+            {
+              "_id": "page_pdc_mobile_001",
+              "name": "Mobile View Page",
+              "fieldPositions": [
+                {
+                  "_id": "69fda9611467c1db59e4d3e9",
+                  "type": "block",
+                  "displayType": "original",
+                  "fontSize": 28,
+                  "fontWeight": "bold",
+                  "x": 0,
+                  "y": 0,
+                  "width": 3,
+                  "height": 4,
+                  "field": "69fda96121a7adca05d300ed"
+                },
+                {
+                  "_id": "69fda96d0c839f53f00e39fc",
+                  "type": "text",
+                  "displayType": "original",
+                  "x": 0,
+                  "y": 4,
+                  "width": 4,
+                  "height": 9,
+                  "field": "69fda96d180c2668269fa855"
+                },
+                {
+                  "_id": "69fda977525cf2212e2bc466",
+                  "type": "number",
+                  "displayType": "original",
+                  "x": 0,
+                  "y": 13,
+                  "width": 4,
+                  "height": 8,
+                  "field": "69fda9770d8a0f4198bcdd6b"
+                },
+                {
+                  "_id": "69fda97ce247af94be7fac7c",
+                  "type": "date",
+                  "displayType": "original",
+                  "format": "MM/DD/YYYY hh:mma",
+                  "x": 0,
+                  "y": 21,
+                  "width": 4,
+                  "height": 7,
+                  "field": "69fda97c01d09f836499f4d0"
+                },
+                {
+                  "_id": "69fda9833003b431d0ee26b0",
+                  "type": "table",
+                  "displayType": "original",
+                  "x": 0,
+                  "y": 28,
+                  "width": 4,
+                  "height": 15,
+                  "field": "69fda9839f94ae3df377125b"
+                }
+              ],
+              "metadata": {},
+              "hidden": false,
+              "width": 375,
+              "height": 667,
+              "cols": 4,
+              "rowHeight": 8,
+              "layout": "grid",
+              "presentation": "normal",
+              "margin": 0,
+              "padding": 16,
+              "borderWidth": 0
+            }
+          ],
+          "pageOrder": [
+            "page_pdc_mobile_001"
+          ]
+        }
+      ]
+    }
+  ],
+  "fields": [
+    {
+      "file": "file_pdc_mobile_001",
+      "_id": "69fda96121a7adca05d300ed",
+      "type": "block",
+      "title": "Heading Text",
+      "value": "Mobile View",
+      "identifier": "field_69fda96121a7adca05d300ed"
+    },
+    {
+      "file": "file_pdc_mobile_001",
+      "_id": "69fda96d180c2668269fa855",
+      "type": "text",
+      "title": "Text",
+      "identifier": "field_69fda96d180c2668269fa855",
+      "value": "Mobile View"
+    },
+    {
+      "file": "file_pdc_mobile_001",
+      "_id": "69fda9770d8a0f4198bcdd6b",
+      "type": "number",
+      "title": "Number",
+      "identifier": "field_69fda9770d8a0f4198bcdd6b",
+      "value": 999999
+    },
+    {
+      "file": "file_pdc_mobile_001",
+      "_id": "69fda97c01d09f836499f4d0",
+      "type": "date",
+      "title": "Date Time",
+      "identifier": "field_69fda97c01d09f836499f4d0",
+      "value": 1778231739198,
+      "tz": "Asia/Calcutta"
+    },
+    {
+      "file": "file_pdc_mobile_001",
+      "_id": "69fda9839f94ae3df377125b",
+      "type": "table",
+      "title": "Table",
+      "tableColumns": [
+        {
+          "_id": "69fd9b33f281c145236324f4",
+          "type": "text",
+          "title": "Text Column",
+          "width": 0,
+          "deleted": false,
+          "identifier": "69fda9839f94ae3df377125b_column_69fd9b33f281c145236324f4"
+        },
+        {
+          "_id": "69fd9b33b274a8e580fbc89f",
+          "type": "dropdown",
+          "title": "Dropdown Column",
+          "deleted": false,
+          "width": 0,
+          "options": [
+            {
+              "_id": "69fd9b330d0e14af562203f6",
+              "value": "Yes",
+              "deleted": false
+            },
+            {
+              "_id": "69fd9b338ec7e34ced2a9ae3",
+              "value": "No",
+              "deleted": false
+            },
+            {
+              "_id": "69fd9b33a67afdc2c212156d",
+              "value": "N/A",
+              "deleted": false
+            }
+          ],
+          "identifier": "69fda9839f94ae3df377125b_column_69fd9b33b274a8e580fbc89f"
+        },
+        {
+          "_id": "69fd9b33b147eebab75f60ca",
+          "type": "text",
+          "title": "Text Column",
+          "width": 0,
+          "deleted": false,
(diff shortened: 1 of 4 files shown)


vivekjyani (commented):
Behavior change in `duplicatePage(...)` is small, focused, and mirrors the existing reset on the main `duplicatedPage` (lines 1050–1052), so the fix is consistent with what's already there. A few nits below — none are blocking.

**Summary of findings:**
- One dead helper (`tapDuplicateMode`) in the new UI test file.
- A test helper that asserts-but-still-returns-`nil`, forcing every caller to write a `guard ... else { return }`.
- The duplicate-mode alert opened in test step 4 is never dismissed.
- Minor naming nit on the alternate-page variable.

Sources/JoyfillUI/ViewModels/DocumentEditor.swift
```diff
@@ -1062,6 +1062,8 @@ extension DocumentEditor {
                 var originalAltPage = altView.pages![originalAlternatePageIndex]
                 let originalAltPageID = originalAltPage.id
                 originalAltPage.id = duplicatedPage.id
+                originalAltPage.deletable = true
+                originalAltPage.copyable = [.withValues, .withoutValues]
```
vivekjyani, line 1066: Nit (naming): by this point `originalAltPage.id` has already been reassigned to `duplicatedPage.id` on line 1064, so the variable no longer represents the *original* alt page — it's the about-to-be-inserted *duplicated* alt page. The reset (`deletable`/`copyable`) is correct, but the name reads as if we're mutating the source. Consider renaming to `duplicatedAltPage` for symmetry with `duplicatedPage` above. (Pre-existing naming; flagging since this PR is the one that adds semantically-loaded resets to it.)

JoyfillSwiftUIExample/JoyfillUITests/PageDeletableCopyable/PageDeletableCopyableMobileViewUITests.swift
```diff
@@ -0,0 +1,121 @@
...
+        let alert = app.alerts.firstMatch
+        XCTAssertTrue(alert.waitForExistence(timeout: 3), "Duplicate mode dialog should appear.", file: file, line: line)
+        if alert.buttons[title].exists {
+            alert.buttons[title].tap()
+        } else if app.buttons[title].exists {
+            app.buttons[title].tap()
+        } else {
+            XCTFail("Could not find duplicate mode button '\(title)'.", file: file, line: line)
+        }
+    }
```
vivekjyani: Dead code: `tapDuplicateMode(_:)` is defined here but never called by either test in this file. Both tests either inspect the alert directly (`testDuplicateInMobileView...`, lines 83–88) or interact with the delete-confirmation alert (`testDeleteDuplicatedPageInMobileView...`). Please remove this helper, or wire it in if it was intended to be used (e.g., in the first test after asserting `hasWithValues`/`hasWithoutValues`).

JoyfillSwiftUIExample/JoyfillUITests/PageDeletableCopyable/PageDeletableCopyableMobileViewUITests.swift
```diff
@@ -0,0 +1,121 @@
...
+
+    private func mobileViewPageRowOrFail(file: StaticString = #filePath, line: UInt = #line) -> XCUIElement? {
+        let pageRows = app.buttons.matching(identifier: "PageSelectionIdentifier")
+        XCTAssertTrue(pageRows.firstMatch.waitForExistence(timeout: 8), "Page rows should appear in the sheet.", file: file, line: line)
+
+        let row = pageRows.allElementsBoundByIndex.first { row in
+            row.label.contains("Mobile View Page") || row.staticTexts["Mobile View Page"].exists
+        }
+        XCTAssertNotNil(row, "Mobile-view page row should be visible — confirms mobile view is active.", file: file, line: line)
+        return row
```
vivekjyani: Helper shape nit: `mobileViewPageRowOrFail` returns `XCUIElement?` and uses `XCTAssertNotNil`, but `XCTAssertNotNil` does **not** stop test execution. That forces every caller to add a `guard let ... else { return }` (see lines 54 and 95) — and if a future caller forgets, the test will continue past the failed assertion and crash on the next use.

Two cleaner shapes:
1. Use `XCTUnwrap` and `throws`: `private func mobileViewPageRowOrFail(...) throws -> XCUIElement` + mark tests `throws`, then `let row = try mobileViewPageRowOrFail()`.
2. Return non-optional and `fatalError` / `XCTFail`+`fatalError` if not found.

Option 1 is idiomatic XCTest.

JoyfillSwiftUIExample/JoyfillUITests/PageDeletableCopyable/PageDeletableCopyableMobileViewUITests.swift
```diff
@@ -0,0 +1,121 @@
...
+        XCTAssertTrue(dupDuplicate.exists,
+                      "Duplicated page must show duplicate button (copyable reset).")
+        dupDuplicate.tap()
+
+        let alert = app.alerts.firstMatch
+        XCTAssertTrue(alert.waitForExistence(timeout: 3), "Duplicate dialog should appear on the duplicated page.")
+        let hasWithValues = alert.buttons["With Values"].exists || app.buttons["With Values"].exists
+        let hasWithoutValues = alert.buttons["Without Values"].exists || app.buttons["Without Values"].exists
+        XCTAssertTrue(hasWithValues, "Duplicated page's copyable must include with-values.")
+        XCTAssertTrue(hasWithoutValues, "Duplicated page's copyable must include without-values.")
```
vivekjyani, line 93: The duplicate-mode dialog opened on line 81 is never dismissed before the test ends. XCUITest relaunches the app between test methods so this likely won't cross-pollute, but leaving an open system alert at teardown can flake on CI (the alert can intercept the next launch's events on some iOS versions). Suggest dismissing it explicitly after the assertions:

```swift
if alert.buttons["Cancel"].exists { alert.buttons["Cancel"].tap() }
```

or whichever cancel affordance the duplicate-mode sheet exposes.

JoyfillSwiftUIExample/JoyfillUITests/PageDeletableCopyable/PageDeletableCopyableMobileView.json
```diff
@@ -0,0 +1,236 @@
+{
+  "_id": "69fda955b580cf6fab953a4e",
+  "type": "template",
+  "stage": "published",
+  "identifier": "doc_pdc_mobile_001",
+  "name": "PDC Mobile",
```
vivekjyani, line 6: Minor: the web-only `pages[0]` carries `name: "Web Only Page"` while `views[0].pages[0]` (same `_id`) carries `name: "Mobile View Page"`. That asymmetry is intentional for the test (it's how `mobileViewPageRowOrFail` confirms mobile view is the active rendering), so this is fine — but worth a one-line comment in the fixture or in the test helper so a future maintainer doesn't "fix" the names to match and silently break the discriminator.

Sources/JoyfillUI/ViewModels/DocumentEditor.swift
vivekjyani: Thanks for the catch — I see the reasoning, but I'd prefer to keep originalAltPage as-is.

The variable is initialized from altView.pages![originalAlternatePageIndex] (the original alt-view page), and originalAltPage.id = duplicatedPage.id is just an id reassignment so the alt-view entry stays linked to its web-side counterpart. The underlying page data (fieldPositions, structure, etc.) is still sourced from the original — we're not constructing a new page, just patching identifiers and the spec-mandated deletable / copyable defaults on the same struct before insertion.

I read the name as describing data provenance ("this came from the original alt page") rather than final role in the document, which is consistent with how the surrounding code reads. Renaming here would also create an inconsistency with the rest of the duplication flow unless we sweep similar spots, which feels out of scope for this PR.

JoyfillSwiftUIExample/JoyfillUITests/PageDeletableCopyable/PageDeletableCopyableMobileViewUITests.swift
vivekjyani: Good catch — the helper was left over from an earlier draft of the test and never wired in. Removed it in the follow-up commit. The existence assertions on lines 85-88 cover what the helper would have added, and the lingering alert hasn't shown any flakiness across local runs, so I'm leaving the dismissal out for now.

JoyfillSwiftUIExample/JoyfillUITests/PageDeletableCopyable/PageDeletableCopyableMobileViewUITests.swift
vivekjyani: Thanks — fair point on the shape. I'd prefer to keep it as-is for this PR though.

Both current call sites (lines 54 and 95) use guard let originalRow = mobileViewPageRowOrFail() else { return }, so the early-return is explicit at the call site and the failed XCTAssertNotNil is recorded before the test exits. The "future caller forgets" risk is real but contained — this file only has two test methods and isn't a shared utility, so I don't think the throws / XCTUnwrap refactor pays for itself here.

JoyfillSwiftUIExample/JoyfillUITests/PageDeletableCopyable/PageDeletableCopyableMobileViewUITests.swift
vivekjyani: Good call — added an explicit alert.buttons["Cancel"].tap() after the existence assertions (guarded by .exists so it's a no-op if the affordance ever changes). Cheap insurance against the CI flake mode you described.

JoyfillSwiftUIExample/JoyfillUITests/PageDeletableCopyable/PageDeletableCopyableMobileView.json
vivekjyani: Fair point — leaving the names as-is though. The fixture asymmetry is contained to this single test file and mobileViewPageRowOrFail is the only consumer of the "Mobile View Page" label, so a future maintainer "fixing" the JSON would have to also be touching the helper at the same time to cause any harm. Happy to add the comment if it bites us later, but I'd rather not annotate every intentional fixture quirk preemptively.

VishnuBishnoi (commented):

**The tests do not prove the fix is necessary.**
The fixture's mobile-view page already has `deletable: true` and `copyable: [with-values, without-values]`. Even without the source change, the duplicated alt page would inherit those defaults and both tests would pass. To actually exercise the reset behavior, either:
- modify the fixture so the alt page starts with `deletable: false` and an empty/partial `copyable`, OR
- add a second fixture with restrictive defaults and assert the duplicate is still fully enabled.

As-is, regressing the `originalAltPage.deletable = true` / `copyable = [...]` lines would not break the test suite.

**Brittle duplicate-row index.** `pageRows.element(boundBy: initialCount)` assumes the duplicate lands at the end of the list. The implementation inserts at `originalIdx + 1`, not at the end. It only works here because the fixture has one page. A future multi-page fixture silently breaks this. Prefer locating by the new page's id or `index(of: originalRow) + 1`.

**Localization-brittle assertions.** `alert.buttons["With Values"].exists` / `"Without Values"` match by visible label. If the project ever localizes, tests break. Use accessibility identifiers if the alert buttons expose them.

**Missing inverse-assertion test.** No test asserts that the *source* alt page's `deletable`/`copyable` are unchanged after duplication. A defensive guard would catch an accidental in-place mutation.

**Cite the spec in code.** The reset is unconditional — if a customer relies on duplicates inheriting source restrictions, this is a behavior change. A one-line comment referencing NO-2131 at the new lines would help future maintainers.

**Fixture name coupling.** Tests detect the mobile-view row by matching the label "Mobile View Page". Add an accessibility identifier or use page id to decouple from the display string.

PR #327, opened 2026-05-13
NO-2140 Include deleted row data in rowDelete change logs

vivekjyani:
## 1. Context / Spec
[NO-2140] When a table/collection row is deleted, the `field.value.rowDelete` change log only carried the `rowId` (and `parentPath`/`schemaId` for nested rows). Consumers replaying or auditing changes had no way to recover the deleted row's actual values. This PR enriches the change log with the full deleted row payload.

## 2. What Changed
- **`Sources/JoyfillUI/ViewModels/DocumentEditor+ChangeHandler.swift`**
  - Top-level delete path (`deleteRow`-style flow): capture each row's `anyDictionary` snapshot into `deletedRowsByID` before `setDeleted()` runs, then pass it through to `onChangeForDelete`.
  - Nested delete path: same snapshotting, but the snapshot is taken at the moment of `elements.remove(at:)` (both direct and recursive removal).
  - `deleteRowRecursively` now returns the removed `ValueElement?` instead of `Bool`, so the caller can attach the snapshot to the change log.
  - `onChangeForDelete` / `onChangeForDeleteNestedRow` accept `rowsByID: [String: [String: Any]]` and embed the snapshot in the `change` dict under a new `"row"` key alongside the existing `"rowId"` / `"parentPath"` / `"schemaId"`.
- **Tests**
  - `JoyfillSwiftUIExample/JoyfillUITests/TableField/TableFieldTests.swift` — asserts the delete changelog now contains the full deleted row properties.
  - `JoyfillSwiftUIExample/JoyfillUITests/CollectionField/CollectionFieldSearchFilterTests.swift` — same verification for the collection/nested case.

## 3. Design Decisions
- **Snapshot at delete-time, not after.** For the top-level path the row is mutated by `setDeleted()`, and for the nested path it's removed from the array — in both cases the original payload is gone by the time `onChangeForDelete*` runs. Capturing `anyDictionary` into a `[String: [String: Any]]` keyed by `rowId` keeps the snapshot intact without holding `ValueElement` references that could be mutated downstream.
- **Recursive function now returns `ValueElement?`.** Returning the removed element is strictly more information than `Bool` and avoids a second tree walk; callers that only cared about success can still treat `nil` as "not found."
- **Iterating `Array(children.keys)`** in `deleteRowRecursively` to avoid mutating-while-iterating the dictionary when a nested write-back happens.
- **Snapshot only when `shouldSendEvent` is true.** No need to allocate dictionaries for silent/internal deletes.
- **Key name `"row"`** in the change payload — matches the singular `"rowId"` already present and reads naturally next to it.

## 4. Screenshots / Video
N/A — change is in the emitted change-log payload, no UI difference.

## 5. Public API Impact
- No public Swift API changed — `onChangeForDelete` / `onChangeForDeleteNestedRow` and `deleteRowRecursively` are all `private`.
- **Change-log schema is extended**: `change` dicts for `target: "field.value.rowDelete"` now include an additional `"row"` key with the deleted row's full dictionary. Existing consumers reading only `rowId` are unaffected, but downstream docs describing the rowDelete payload should be updated to mention the new `"row"` field.

## 6. Test Coverage
- UI tests updated in both `TableFieldTests` and `CollectionFieldSearchFilterTests` to assert the full deleted-row properties are present in the changelog (covers top-level and nested cases).
- No new logic branches beyond snapshot capture; existing delete coverage exercises both `shouldSendEvent` paths.

## 7. Reviewer Notes
- Please double-check the recursive nested case — the snapshot is taken inside `deleteRowRecursively` via the returned `ValueElement?`, so the `childrens` of the deleted row are included in the snapshot (which is intentional, so a deleted parent's subtree is recoverable).
- If the backend/spec expects a different key name than `"row"` for the embedded payload, that's a one-line rename in `onChangeForDelete` / `onChangeForDeleteNestedRow`.

### DIFF
--- a/Sources/JoyfillUI/ViewModels/DocumentEditor+ChangeHandler.swift
+++ b/Sources/JoyfillUI/ViewModels/DocumentEditor+ChangeHandler.swift
@@ -30,6 +30,7 @@ extension DocumentEditor {
         guard var lastRowOrder = field.rowOrder else {
             return elements
         }
+        var deletedRowsByID = [String: [String: Any]]()
 
         for row in rowIDs {
             guard let index = elements.firstIndex(where: { $0.id == row }) else {
@@ -37,14 +38,17 @@ extension DocumentEditor {
                 return elements
             }
             var element = elements[index]
+            if shouldSendEvent {
+                deletedRowsByID[row] = element.anyDictionary
+            }
             element.setDeleted()
             elements[index] = element
             lastRowOrder.removeAll(where: { $0 == row })
         }
         fieldMap[fieldId]?.value = ValueUnion.valueElementArray(elements)
         fieldMap[fieldId]?.rowOrder = lastRowOrder
-        guard shouldSendEvent else { return  elements }
-        onChangeForDelete(fieldIdentifier: fieldIdentifier, rowIDs: rowIDs)
+        guard shouldSendEvent else { return elements }
+        onChangeForDelete(fieldIdentifier: fieldIdentifier, rowIDs: rowIDs, rowsByID: deletedRowsByID)
         return elements
     }
     
@@ -56,40 +60,47 @@ extension DocumentEditor {
         }
         guard var elements = field.valueToValueElements else { return [] }
         guard !rowIDs.isEmpty else { return elements }
+        var deletedRowsByID = [String: [String: Any]]()
         
         for row in rowIDs {
             if let index = elements.firstIndex(where: { $0.id == row }) {
-                elements.remove(at: index)
+                let deletedElement = elements.remove(at: index)
+                if shouldSendEvent {
+                    deletedRowsByID[row] = deletedElement.anyDictionary
+                }
             } else {
-                _ = deleteRowRecursively(rowId: row, in: &elements)
+                if let deletedElement = deleteRowRecursively(rowId: row, in: &elements) {
+                    if shouldSendEvent {
+                        deletedRowsByID[row] = deletedElement.anyDictionary
+                    }
+                }
             }
         }
         fieldMap[fieldId]?.value = ValueUnion.valueElementArray(elements)
         guard shouldSendEvent else { return elements }
         var parentPath = computeParentPath(targetParentId: parentRowId, nestedKey: nestedKey, in: [rootSchemaKey : elements]) ?? ""
-        onChangeForDeleteNestedRow(fieldIdentifier: fieldIdentifier, rowIDs: rowIDs, parentPath: parentPath, schemaId: nestedKey)
+        onChangeForDeleteNestedRow(fieldIdentifier: fieldIdentifier, rowIDs: rowIDs, parentPath: parentPath, schemaId: nestedKey, rowsByID: deletedRowsByID)
         return elements
     }
 
-    private func deleteRowRecursively(rowId: String, in elements: inout [ValueElement]) -> Bool {
+    private func deleteRowRecursively(rowId: String, in elements: inout [ValueElement]) -> ValueElement? {
         for i in 0..<elements.count {
             if elements[i].id == rowId {
-                elements.remove(at: i)
-                return true
+                return elements.remove(at: i)
             }
             if var children = elements[i].childrens {
-                for key in children.keys {
+                for key in Array(children.keys) {
                     if var nestedElements = children[key]?.valueToValueElements {
-                        if deleteRowRecursively(rowId: rowId, in: &nestedElements) {
+                        if let deletedElement = deleteRowRecursively(rowId: rowId, in: &nestedElements) {
                             children[key]?.value = ValueUnion.valueElementArray(nestedElements)
                             elements[i].childrens = children
-                            return true
+                            return deletedElement
                         }
                     }
                 }
             }
         }
-        return false
+        return nil
     }
     
     /// Duplicates specified rows in a table field.
@@ -1316,7 +1327,7 @@ extension DocumentEditor {
         events?.onChange(changes: changes, document: document)
     }
 
-    private func onChangeForDelete(fieldIdentifier: FieldIdentifier, rowIDs: [String]) {
+    private func onChangeForDelete(fieldIdentifier: FieldIdentifier, rowIDs: [String], rowsByID: [String: [String: Any]]) {
         guard let context = makeFieldChangeContext(for: fieldIdentifier) else { return }
         
         let event = FieldChangeData(fieldIdentifier: fieldIdentifier)
@@ -1324,6 +1335,7 @@ extension DocumentEditor {
         var changes = [Change]()
         
         for targetRow in targetRowIndexes {
+            let rowObject = rowsByID[targetRow.id] ?? [:]
             var change = Change(v: 1,
                                 sdk: "swift",
                                 target: "field.value.rowDelete",
@@ -1334,7 +1346,10 @@ extension DocumentEditor {
                                 fieldId: event.fieldIdentifier.fieldID,
                                 fieldIdentifier: context.fieldIdentifier,
                                 fieldPositionId: context.fieldPositionID,
-                                change: ["rowId": targetRow.id],
+                                change: [
+                                    "rowId": targetRow.id,
+                                    "row": rowObject
+                                ],
                                 createdOn: Date().timeIntervalSince1970)
             changes.append(change)
         }
@@ -1342,7 +1357,7 @@ extension DocumentEditor {
 //        refreshField(fieldId: fieldIdentifier.fieldID, fieldIdentifier: fieldIdentifier)
     }
     
-    private func onChangeForDeleteNestedRow(fieldIdentifier: FieldIdentifier, rowIDs: [String], parentPath: String, schemaId: String) {
+    private func onChangeForDeleteNestedRow(fieldIdentifier: FieldIdentifier, rowIDs: [String], parentPath: String, schemaId: String, rowsByID: [String: [String: Any]]) {
         guard let context = makeFieldChangeContext(for: fieldIdentifier) else { return }
         
         let event = FieldChangeData(fieldIdentifier: fieldIdentifier)
@@ -1350,6 +1365,7 @@ extension DocumentEditor {
         var changes = [Change]()
         
         for targetRow in targetRowIndexes {
+            let rowObject = rowsByID[targetRow.id] ?? [:]
             var change = Change(v: 1,
                                 sdk: "swift",
                                 target: "field.value.rowDelete",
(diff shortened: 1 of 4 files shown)


Sources/JoyfillUI/ViewModels/DocumentEditor+ChangeHandler.swift
```diff
@@ -56,40 +60,47 @@ extension DocumentEditor {
...
+                    deletedRowsByID[row] = deletedElement.anyDictionary
+                }
             } else {
-                _ = deleteRowRecursively(rowId: row, in: &elements)
+                if let deletedElement = deleteRowRecursively(rowId: row, in: &elements) {
+                    if shouldSendEvent {
+                        deletedRowsByID[row] = deletedElement.anyDictionary
+                    }
+                }
             }
```
vivekjyani, line 82: **Bug / pre-existing edge case now made visible.** Unlike the top-level `deleteRows` (which early-returns when a row isn't found at line 36–39), this nested loop silently continues when neither `firstIndex` nor `deleteRowRecursively` finds the row. The `rowID` is still passed to `onChangeForDeleteNestedRow`, which will then emit a `field.value.rowDelete` change with `row: [:]` for a row that was never actually deleted.

Before this PR the empty payload wasn't observable, but now `row: {}` in the change-log is a clear signal of this mismatch. Suggest one of:

```swift
for row in rowIDs {
    if let index = elements.firstIndex(where: { $0.id == row }) {
        let deletedElement = elements.remove(at: index)
        if shouldSendEvent { deletedRowsByID[row] = deletedElement.anyDictionary }
    } else if let deletedElement = deleteRowRecursively(rowId: row, in: &elements) {
        if shouldSendEvent { deletedRowsByID[row] = deletedElement.anyDictionary }
    } else {
        Log("Row not found: \(row)", type: .error)
        // either `continue` and filter rowIDs before emit, or bail like deleteRows does
    }
}
```

and then filter the `rowIDs` passed to `onChangeForDeleteNestedRow` to only the IDs present in `deletedRowsByID`, so consumers don't see phantom `rowDelete` events with empty `row` payloads.

vivekjyani (commented):
Reviewed the change-log enrichment. The snapshot approach is sound and the tests cover both top-level and nested paths. Left a couple of inline notes — one about an edge case in the nested path where a missing row still emits a `rowDelete` change (now with an empty `row` payload), and a smaller consistency note about snapshot timing.

Sources/JoyfillUI/ViewModels/DocumentEditor+ChangeHandler.swift
```diff
@@ -30,21 +30,25 @@ extension DocumentEditor {
...
         for row in rowIDs {
             guard let index = elements.firstIndex(where: { $0.id == row }) else {
                 Log("Row not found: \(row)", type: .error)
                 return elements
             }
             var element = elements[index]
+            if shouldSendEvent {
+                deletedRowsByID[row] = element.anyDictionary
+            }
             element.setDeleted()
```
vivekjyani: Snapshot is captured **before** `setDeleted()` runs, so the emitted `row` payload has `deleted: false` (confirmed by the new `TableFieldTests` assertion). That's a reasonable choice — consumers get the row "as it was at delete time" — but worth being explicit in a comment here, because:

1. The nested path snapshots *at* `elements.remove(at:)`, so its semantics differ subtly (no `deleted` flag was ever set, snapshot is the live row).
2. A future refactor that moves `setDeleted()` above the snapshot capture would silently flip the payload to `deleted: true` and break consumers without breaking the test (the test only checks `deleted: false`).

A one-line `// snapshot pre-mutation so the change-log records the row as it existed before delete` here would lock the intent in.

Sources/JoyfillUI/ViewModels/DocumentEditor+ChangeHandler.swift
```diff
@@ -56,40 +60,47 @@ extension DocumentEditor {
...
+    private func deleteRowRecursively(rowId: String, in elements: inout [ValueElement]) -> ValueElement? {
         for i in 0..<elements.count {
             if elements[i].id == rowId {
-                elements.remove(at: i)
-                return true
+                return elements.remove(at: i)
             }
             if var children = elements[i].childrens {
-                for key in children.keys {
+                for key in Array(children.keys) {
```
vivekjyani: Nit: the `Array(children.keys)` defensive copy isn't strictly necessary here — the inner block only mutates existing values (`children[key]?.value = ...`), it doesn't insert/remove keys, so iterating `children.keys` directly is safe in Swift. Harmless either way, but if it was changed specifically to avoid a crash, that crash wouldn't have occurred. Up to you whether to revert or keep as defensive.

Sources/JoyfillUI/ViewModels/DocumentEditor+ChangeHandler.swift
```diff
@@ -56,40 +60,47 @@ extension DocumentEditor {
...
         return elements
     }
 
-    private func deleteRowRecursively(rowId: String, in elements: inout [ValueElement]) -> Bool {
+    private func deleteRowRecursively(rowId: String, in elements: inout [ValueElement]) -> ValueElement? {
         for i in 0..<elements.count {
             if elements[i].id == rowId {
-                elements.remove(at: i)
-                return true
+                return elements.remove(at: i)
```
vivekjyani, line 94: Worth noting (and the PR description already calls this out): the returned `ValueElement` for a recursively-deleted row carries its full `childrens` subtree along with it. The collection test happens to delete a leaf row so `children: [:]`, but if a parent row with descendants is deleted, the emitted `row` payload will contain the entire subtree. That's likely the desired behavior for replay/audit, but it means the change-log payload size is now unbounded in the depth/breadth of the deleted subtree. Worth a test that explicitly covers "delete a parent with grandchildren" to lock in the expectation.

Sources/JoyfillUI/ViewModels/DocumentEditor+ChangeHandler.swift
```diff
@@ -56,40 +60,47 @@ extension DocumentEditor {
...
+                if let deletedElement = deleteRowRecursively(rowId: row, in: &elements) {
+                    if shouldSendEvent {
+                        deletedRowsByID[row] = deletedElement.anyDictionary
+                    }
+                }
             }
         }
         fieldMap[fieldId]?.value = ValueUnion.valueElementArray(elements)
         guard shouldSendEvent else { return elements }
         var parentPath = computeParentPath(targetParentId: parentRowId, nestedKey: nestedKey, in: [rootSchemaKey : elements]) ?? ""
```
vivekjyani, line 86: Unrelated nit (pre-existing): `var parentPath` can be `let parentPath` — it's never mutated after assignment. Feel free to leave for a separate cleanup PR.

VishnuBishnoi (commented):

**Semantic inconsistency between top-level and nested snapshots:**
The top-level path snapshots `element.anyDictionary` *before* `setDeleted()` runs, so the embedded row has `"deleted": false` (the `TableFieldTests` assertion confirms this). The nested path uses `remove(at:)`, which does not toggle the flag — the snapshot reflects whatever `deleted` was on the row beforehand. Net result: a `rowDelete` event embeds a row payload claiming the row is not deleted. Consider either (a) stripping `deleted` from the snapshot before embedding, or (b) snapshotting after `setDeleted()` so the top-level path emits `"deleted": true` consistently. Pick one — current behavior diverges by path.

**Subtree-recovery claim is not exercised by tests.**
The PR description says "deleted parent's subtree is recoverable" via the recursive snapshot. But the collection test deletes a leaf row whose `children` is `[:]`. Add a test that deletes a parent with at least one nested child and asserts the `children` dict appears in the embedded `"row"` payload — otherwise the recursion's subtree path is uncovered.

**Cross-platform key naming.** Please confirm with the web/JS SDK that `"row"` matches the schema — if other SDKs emit `"deletedRow"` / `"payload"` / etc., align before this ships. Cheap to change now, expensive to change after consumers integrate.

**Concurrency.** The `[String: [String: Any]]` payload propagates into `Change.change: [String: Any]`. Fine for serialization, but Swift 6 strict-concurrency mode will flag this as non-Sendable when an `onChange` consumer hops actors. Worth a note for the eventual Swift 6 migration.

**Unit-test gap.** All coverage is end-to-end UI. A pure unit test for `deleteRowRecursively` at depth ≥ 2 (asserting both the returned `ValueElement?` and that its `childrens` are intact) would catch any regression cheaply.

**Duplicate `rowIDs` edge case.** If callers pass the same id twice, the second occurrence falls through to `deleteRowRecursively` and returns `nil`, silently dropping a snapshot. Probably not reachable today, but a precondition assert or de-duplication at the entry point would be defensive.

Sources/JoyfillUI/ViewModels/DocumentEditor+ChangeHandler.swift
vivekjyani: Good catch — agreed it can be let. Skipping in this PR to keep the diff scoped to the rowDelete payload change; will pick it up in a separate cleanup pass.

Sources/JoyfillUI/ViewModels/DocumentEditor+ChangeHandler.swift
vivekjyani: Good catch on the future-refactor hazard. Went with the stronger fix: snapshot **after** `setDeleted()` so the payload carries `"deleted": true`, and updated the `TableFieldTests` assertion to match. Reordering `setDeleted()` will now break the test instead of silently flipping the payload.
Scoped to the top-level path in this PR — leaving the nested `deleteNestedRows` semantics for a separate change.

Sources/JoyfillUI/ViewModels/DocumentEditor+ChangeHandler.swift
vivekjyani: You're right — reverted to direct `for key in children.keys`. The inner block only mutates values of existing keys and returns immediately after the mutation, so there's no iteration hazard to defend against. Keeping the `Array(...)` copy would have implied a risk that doesn't exist.

Sources/JoyfillUI/ViewModels/DocumentEditor+ChangeHandler.swift
vivekjyani: Added `testDeleteNestedCollectionItemPayloadIncludesSubtree` in `DocumentEditor+ChangeHandlerTests` — deletes an L1 row that owns three L2 grandchildren and asserts the embedded `row["children"][schemaKey]["value"]` contains all three IDs. A future change that drops `childrens` from the recursive return will now break this test.

Sources/JoyfillUI/ViewModels/DocumentEditor+ChangeHandler.swift
vivekjyani: Fixed the main issue — `deleteNestedRows` now only passes IDs that were actually deleted to `onChangeForDeleteNestedRow`, and short-circuits the emit when nothing was deleted. Skipping the `Log` line for now since it's diagnostic-only.

PR #328, opened 2026-05-13
NO-2145: Support row form in readonly mode with disabled cells

vivekjyani:
## Context
In readonly mode, the row form was never reachable, so users couldn't see cell decorators that only render inside the form. Spec: allow the row form to open in readonly so decorators are visible, but every editor inside must be disabled.

## What changed
- `TableViewModel.showSingleClickEditButton` / `CollectionViewModel.showSingleClickEditButton`: dropped `mode == .fill` gate so the pencil opens the form in readonly too
- `EditMultipleRowsSheetView` / `CollectionEditMultipleRowsSheetView`: single `.disabled(mode == .readonly)` on the column `ForEach`; decorator buttons inside `columnTitle` / `fieldTitle` use `.environment(\.isEnabled, true)` to stay tappable
- Hid the table "More" popover and the in-form "+" insert-row button in readonly (their actions all mutate data)
- `TableTextRowFormView`: readonly text now uses `.frame(maxWidth: .infinity)` so empty cells don't collapse to zero width
- Example app: `SampleFormFooterController` marked `@MainActor` to fix "Publishing changes from background threads" warning
- Added `RowFormReadonlyUITests` in `DecoratorAPIUITests.swift` covering the contract above (table, collection root, nested L1)

## Design decisions
- Single `.disabled` on the `ForEach` instead of per-cell — keeps the override surface to one line per file
- `.environment(\.isEnabled, true)` chosen over restructuring into Groups because it preserves the existing view tree and only re-enables the decorator subtree
- Row selector and "More" popover stay hidden in readonly — bulk operations don't make sense without edit capability; single-click pencil is the entry point

## Screenshots
_Attach: readonly row form showing decorators + disabled editors._

## Public API
No public API changes. No docs update needed.

## Tests
Automated: new `RowFormReadonlyUITests` class launches with `--mode readonly` and exercises the row form contract end-to-end across three scopes:
- `testTable_CellDecoratorInsideRowFormStaysTappable` — table row form
- `testCollection_RootCellDecoratorInsideRowFormStaysTappable` — collection root row form
- `testCollection_NestedL1CellDecoratorStaysTappable` — collection nested L1 row form

Each test:
1. Adds a common/specific cell decorator via the existing `DecoratorUITestSupport` bridge.
2. Opens the row form via the single-click pencil (only entry point in readonly).
3. Asserts the column-title decorator is hittable and that tapping it fires `onFocus` — proves the `.environment(\.isEnabled, true)` override works through the parent `.disabled`.
4. Iterates every column in that row form's schema via the `assertCellsNotTappable` helper: `waitForExistence` → `isEnabled == false` → `tap()` → `keyboards.element` must not appear.
5. Cells that swap to a static `Text` in readonly (`TableTextRowFormView` → collection text, `TableBarcodeView` → table barcode) are checked via absence of the editable element instead, since they never surface as an interactive control.

Manual: verified readonly form opens, all cell types are non-editable, decorators remain tappable, "+" / "More" / row selector all hidden.

## Notes for reviewer
- `.disabled` propagates by default; the `.environment(\.isEnabled, true)` override on `FieldDecoratorsView` is the only way decorators stay live. If you move decorators around, keep that modifier on the new home.
- Same shape applied to both table and collection paths — diffs mirror each other on purpose.
- The UI tests use `Decorator.json` and the existing pasteboard-free `decorator_command_input` bridge; no new infrastructure was added.

### DIFF
--- a/JoyfillSwiftUIExample/JoyfillExample/Footer Example/SampleFormFooter.swift
+++ b/JoyfillSwiftUIExample/JoyfillExample/Footer Example/SampleFormFooter.swift
@@ -10,6 +10,7 @@ import UIKit
 
 // MARK: - Footer coordinator (validate / navigate + FormChangeEvent)
 
+@MainActor
 final class SampleFormFooterController: ObservableObject, FormChangeEvent {
 
     weak var documentEditor: DocumentEditor?
@@ -57,39 +58,45 @@ final class SampleFormFooterController: ObservableObject, FormChangeEvent {
 
     // MARK: - FormChangeEvent
 
-    func onChange(changes: [Change], document: JoyDoc) {
-        guard let editor = self.documentEditor else { return }
-        
-        let validation = editor.validate()
-        let validities = validation.fieldValidities
-        let total = validities.count
-        let completed = validities.filter { $0.status == .valid }.count
-        
-        self.fieldPaths = Self.buildPaths(from: validities)
-        if self.currentFieldIndex >= self.fieldPaths.count {
-            self.currentFieldIndex = self.fieldPaths.isEmpty ? -1 : self.fieldPaths.count - 1
+    nonisolated func onChange(changes: [Change], document: JoyDoc) {
+        Task { @MainActor in
+            guard let editor = self.documentEditor else { return }
+
+            let validation = editor.validate()
+            let validities = validation.fieldValidities
+            let total = validities.count
+            let completed = validities.filter { $0.status == .valid }.count
+
+            self.fieldPaths = Self.buildPaths(from: validities)
+            if self.currentFieldIndex >= self.fieldPaths.count {
+                self.currentFieldIndex = self.fieldPaths.isEmpty ? -1 : self.fieldPaths.count - 1
+            }
+            self.completedText = "\(completed) of \(total) Completed"
+            self.navigationEnabled = validation.status == .invalid
         }
-        self.completedText = "\(completed) of \(total) Completed"
-        self.navigationEnabled = validation.status == .invalid
     }
 
-    func onFocus(event: Event) {
+    nonisolated func onFocus(event: Event) {
         let focusedPageID = event.pageEvent?.page.id ?? event.fieldEvent?.pageID
-        updateFooterVisibility(for: focusedPageID)
+        Task { @MainActor in
+            self.updateFooterVisibility(for: focusedPageID)
+        }
     }
 
-    func onBlur(event: Event) {
+    nonisolated func onBlur(event: Event) {
         guard let pageEvent = event.pageEvent,
               pageEvent.type == "page.blur",
               let blurredPageID = pageEvent.page.id else { return }
 
-        if blurredPageID == footerPageID {
-            isFooterVisible = false
+        Task { @MainActor in
+            if blurredPageID == self.footerPageID {
+                self.isFooterVisible = false
+            }
         }
     }
-    func onUpload(event: Joyfill.UploadEvent) {}
-    func onCapture(event: Joyfill.CaptureEvent) {}
-    func onError(error: Joyfill.JoyfillError) {}
+    nonisolated func onUpload(event: Joyfill.UploadEvent) {}
+    nonisolated func onCapture(event: Joyfill.CaptureEvent) {}
+    nonisolated func onError(error: Joyfill.JoyfillError) {}
 
     // MARK: - Paths
 
--- a/Sources/JoyfillUI/View/Fields/TableView/TableModalTopNavigationView.swift
+++ b/Sources/JoyfillUI/View/Fields/TableView/TableModalTopNavigationView.swift
@@ -16,7 +16,7 @@ struct TableModalTopNavigationView: View {
 
             Spacer()
 
-            if !viewModel.tableDataModel.selectedRows.isEmpty {
+            if !viewModel.tableDataModel.selectedRows.isEmpty && viewModel.tableDataModel.mode == .fill {
                 Button(action: {
                     showingPopover = true
                 }) {
@@ -245,6 +245,7 @@ struct EditMultipleRowsSheetView: View {
                 FieldDecoratorsView(decorators: decorators, visibleLimit: viewModel.decoratorConfig.visibleLimitInFields) { decorator in
                     viewModel.tableDataModel.documentEditor?.reportDecoratorAction(fieldIdentifier: viewModel.tableDataModel.fieldIdentifier, action: decorator.action ?? "", rowIds: viewModel.tableDataModel.selectedRows, columnId: col.id)
                 }
+                .environment(\.isEnabled, true)
             }
         }
         .padding(.bottom, -8)
@@ -291,19 +292,21 @@ struct EditMultipleRowsSheetView: View {
                                 .disabled(viewModel.tableDataModel.shouldDisableMoveDown)
                                 .accessibilityIdentifier("LowerRowButtonIdentifier")
                                 
-                                Button(action: {
-                                    viewModel.insertBelowFromBulkEdit()
-                                }, label: {
-                                    ZStack {
-                                        RoundedRectangle(cornerRadius: 6)
-                                            .stroke(.blue, lineWidth: 1)
-                                            .frame(width: 27, height: 27)
-                                        
-                                        Image(systemName: "plus")
-                                            .foregroundStyle(.blue)
-                                    }
-                                })
-                                .accessibilityIdentifier("PlusTheRowButtonIdentifier")
+                                if viewModel.tableDataModel.mode == .fill {
+                                    Button(action: {
+                                        viewModel.insertBelowFromBulkEdit()
+                                    }, label: {
+                                        ZStack {
+                                            RoundedRectangle(cornerRadius: 6)
+                                                .stroke(.blue, lineWidth: 1)
+                                                .frame(width: 27, height: 27)
+
+                                            Image(systemName: "plus")
+                                                .foregroundStyle(.blue)
+                                        }
+                                    })
+                                    .accessibilityIdentifier("PlusTheRowButtonIdentifier")
+                                }
                             } else {
                                 Spacer()
                             }
@@ -612,6 +615,7 @@ struct EditMultipleRowsSheetView: View {
                     }
                     .id(col.id)
                 }
+                .disabled(viewModel.tableDataModel.mode == .readonly)
                 Spacer()
             }
             .padding(.all, 16)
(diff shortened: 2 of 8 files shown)


JoyfillSwiftUIExample/JoyfillExample/Footer Example/SampleFormFooter.swift
```diff
@@ -57,39 +58,45 @@ final class SampleFormFooterController: ObservableObject, FormChangeEvent {
...
-        let validities = validation.fieldValidities
-        let total = validities.count
-        let completed = validities.filter { $0.status == .valid }.count
-        
-        self.fieldPaths = Self.buildPaths(from: validities)
-        if self.currentFieldIndex >= self.fieldPaths.count {
-            self.currentFieldIndex = self.fieldPaths.isEmpty ? -1 : self.fieldPaths.count - 1
+    nonisolated func onChange(changes: [Change], document: JoyDoc) {
+        Task { @MainActor in
+            guard let editor = self.documentEditor else { return }
```
vivekjyani, line 63: Behavior change worth calling out: `onChange` was synchronous before — by the time it returned, `completedText` / `navigationEnabled` / `fieldPaths` were updated. After this change the work runs on a later main-actor tick via `Task { @MainActor }`, so any caller that reads controller state right after the editor finishes a change will see stale values for one run-loop turn. SwiftUI views that observe the `@Published` properties will still update correctly, so this is probably fine in practice — but flagging because the example app is a reference implementation that downstream users copy.

Also: under a high-frequency event source (bulk import, scripted changes) each `onChange` enqueues a separate `Task` that runs `editor.validate()`. There's no coalescing — N changes produce N validations. If that ever shows up as a hot spot, debounce inside the Task. The clean fix would be marking `FormChangeEvent` methods `@MainActor` in the protocol, but that's a public-API change and out of scope here.

Sources/JoyfillUI/View/Fields/TableView/TableModalTopNavigationView.swift
```diff
@@ -245,6 +245,7 @@ struct EditMultipleRowsSheetView: View {
                 FieldDecoratorsView(decorators: decorators, visibleLimit: viewModel.decoratorConfig.visibleLimitInFields) { decorator in
                     viewModel.tableDataModel.documentEditor?.reportDecoratorAction(fieldIdentifier: viewModel.tableDataModel.fieldIdentifier, action: decorator.action ?? "", rowIds: viewModel.tableDataModel.selectedRows, columnId: col.id)
                 }
+                .environment(\.isEnabled, true)
```
vivekjyani, line 248: Using `.environment(\.isEnabled, true)` to override an ancestor `.disabled(true)` is fragile. Apple's `disabled(_:)` docs state: "The higher views in a view hierarchy can override the value you set on this view." `.disabled` is implemented via internal AND-merging of the environment; writing `\.isEnabled = true` directly via `.environment` happens to win in current builds only because the two modifiers don't share storage exactly. This behavior has shifted across iOS versions before.

A more robust pattern: hoist `FieldDecoratorsView` out of the disabled subtree entirely. For example, split `columnTitle(col)` so the decorators render in a sibling that lives *above* the `.disabled` `ForEach` (or wrap only the cell editor in `.disabled`, not the title row). Then decorators don't inherit the disabled environment and you can drop this override.

At minimum: please add a UI test that taps a decorator on a readonly row form — the PR description calls out this modifier as load-bearing, and without a test a future SwiftUI change can silently regress it.

(Same applies to the mirror at `CollectionModalTopNavigationView.swift:319`.)

Sources/JoyfillUI/View/Fields/TableView/TableModalTopNavigationView.swift
```diff
@@ -612,6 +615,7 @@ struct EditMultipleRowsSheetView: View {
                     }
                     .id(col.id)
                 }
+                .disabled(viewModel.tableDataModel.mode == .readonly)
```
vivekjyani, line 618: Two observations on the scope of this `.disabled`:

1. **Redundant with cell-level mode propagation.** The `cellModel` constructed at line ~398 already receives `editMode: viewModel.tableDataModel.mode`, and each cell view (`TableTextRowFormView`, `TableDateView`, etc.) branches on `editMode == .readonly`. The `.disabled` here is belt-and-suspenders, which is fine — just noting it's not the primary defense.

2. **`Apply All` sits outside this disabled subtree.** Lines ~341–366 render the Apply All button when `selectedRows.count != 1`, and it would mutate data via `viewModel.bulkEdit(changes:)`. Today it's unreachable in readonly because `showRowSelector = (mode == .fill)` and the More popover is now gated, so multi-row selection can't happen. But that's a fragile invariant maintained across three different files. Consider adding `.disabled(viewModel.tableDataModel.mode == .readonly)` (or wrapping in `if mode == .fill`) directly on the Apply All button so the guarantee is local and survives future entry points.

The chevron up/down buttons remaining enabled in readonly is correct — they just navigate the selection.

(Same applies to `CollectionModalTopNavigationView.swift:724`.)

Sources/JoyfillUI/View/Fields/TableView/TableTextRowFormView.swift
```diff
@@ -31,6 +31,7 @@ struct TableTextRowFormView: View {
                     .font(.system(size: 15))
                     .lineLimit(1)
                     .padding(.leading, 4)
+                    .frame(maxWidth: .infinity, alignment: .leading)
```
vivekjyani, line 34: Minor: the branch this lands in is `viewMode == .quickView || editMode == .readonly`. `TableTextRowFormView` is only referenced from `CollectionModalTopNavigationView.swift:635` where viewMode is `.modalView`, so the `quickView` half of the condition is dead in practice and your `.frame(maxWidth: .infinity)` only affects the row-form readonly path — i.e. the intended path. No change requested, but if anyone ever wires this view into a quick-view context the wide frame will follow. Worth a short comment, or split the branch to `editMode == .readonly` only since that's the case this PR targets.

vivekjyani (commented):
Reviewed end-to-end. Approach is sound — single `.disabled` at the column-list level, decorator subtree re-enabled, action buttons gated by `mode == .fill`. A few notes inline:

1. The `.environment(\.isEnabled, true)` re-enable trick is fragile against `.disabled(true)` ancestors — please add at least a snapshot/UI test so a future SwiftUI behavior shift surfaces in CI rather than in prod.
2. The example footer's `nonisolated` + `Task { @MainActor }` switch is correct for the warning, but it converts what was a synchronous handler into a deferred one and removes coalescing — worth being aware of for bulk-change scenarios.
3. `Apply All` and the row chevrons live outside the new `.disabled` subtree. Currently unreachable in readonly because `showRowSelector` and the More popover are both gated to `.fill`, but `Apply All` should still be defensively gated on `mode == .fill` in case another entry point opens the bulk sheet with >1 row in readonly later.

No public API changes, change is mirrored cleanly across Table and Collection paths. LGTM with the notes above addressed (the snapshot/test ask is the only one I'd consider blocking).

JoyfillSwiftUIExample/JoyfillExample/Footer Example/SampleFormFooter.swift
vivekjyani: Thanks for the heads-up. The deferred-tick behavior is fine for our use case — the controller only drives SwiftUI views via @Published, so the UI reacts off the property change regardless of which run-loop tick it lands on. No imperative reader is calling into the editor and then reading controller state synchronously.

Agreed on the coalescing point — if validate() ever shows up as a hot spot under bulk changes we can debounce inside the Task. Not seeing it in practice today, so leaving as-is.

Marking the FormChangeEvent protocol @MainActor would be the clean fix but it's a public API change, so out of scope here.

Sources/JoyfillUI/View/Fields/TableView/TableModalTopNavigationView.swift
vivekjyani: Cells weren't handling readonly themselves — most of them (date, number, dropdown, multiSelect, image, signature) only check viewMode == .quickView, not editMode == .readonly. In the row form viewMode is .modalView, so without .disabled on the ForEach they stay editable. That's the actual reason we added it.

Not gating Apply All — row selection isn't available in readonly mode, so the multi-select state that shows Apply All can't be reached. No need for the extra guard.

Sources/JoyfillUI/View/Fields/TableView/TableModalTopNavigationView.swift
vivekjyani: The .environment(\.isEnabled, true) pattern is a standard SwiftUI workaround and has been stable in practice — Apple's "higher views override" doc applies to stacked .disabled modifiers (which AND-merge), not direct environment writes which bypass that merging. Withdrawing the "fragile / shifts across iOS versions" framing, that was speculative.

Restructuring decorators out of the disabled subtree is more churn than it's worth for the layout we have.

Sources/JoyfillUI/View/Fields/TableView/TableTextRowFormView.swift
vivekjyani: viewMode and editMode are different concerns — viewMode is modal vs preview, editMode is fill vs readonly. The OR covers both cases by design. The only caller (CollectionModalTopNavigationView.swift:635) passes viewMode = .modalView, so today the change only affects the readonly path. Leaving the condition as-is — if a future caller does pass .quickView, the wide frame is probably the right behavior there too.

VishnuBishnoi (commented):

**Cross-PR conflict (blocking when both #328 and #329 merge):**
After #329 lands, `TableTextRowFormView` becomes the text cell in the table row form too, and in readonly it renders as a static `Text` — so `("text", app.textFields["EditRowsTextFieldIdentifier"])` in `testTable_CellDecoratorInsideRowFormStaysTappable` will not exist and `waitForExistence` will fail. Use the same absence-of-TextField pattern that `testCollection_RootCellDecoratorInsideRowFormStaysTappable` uses for the table text cell. Whichever PR lands second must update this test.

**Coverage gaps:**
1. The "More" popover is gated to `mode == .fill` in the diff but no test asserts its absence in readonly after row selection. A regression that re-enables it would not be caught.
2. No regression test that opening the readonly row form does not attempt to auto-focus a disabled cell (`navigationFocusColumnId` should be a no-op).
3. PR description placeholder for the screenshot — `TableTextRowFormView` got `.frame(maxWidth: .infinity, alignment: .leading)` on the readonly `Text` branch. Verify visually that non-empty cells aren't pushed/aligned differently in fill mode.

**Code-comment ask:** The `.environment(\.isEnabled, true)` override depends on undocumented SwiftUI behavior — please put the warning currently in the PR description ("If you move decorators around, keep that modifier on the new home") into a source comment at both call sites. Future readers won't see the PR description.

**Nit on `assertCellsNotTappable`:** the helper taps a disabled cell when `isHittable` to actively probe — worth a one-line comment that this checks "the OS did not interpret the tap as activation", not just that `isEnabled == false`.

PR #329, opened 2026-05-15
NO-2139: Fix spurious rowUpdate events on row form text cell focus

vivekjyani:
## 1. Context

Tapping a text cell in the table/collection row edit form (just to open the keyboard, without typing) produced multiple `field.value.rowUpdate` events and **no** `focus` event. The same flow on a 1-row form emitted ~4 spurious `rowUpdate` callbacks before any user input — noisy for `onChange` consumers and visibly misleading in the example app's changelog.

While debugging this, it became obvious that the example app's changelog screen was hard to read for testing — every change in a bulk emission rendered as its own card, with no grouping or filtering — so this PR also refactors that screen.

## 2. What changed

**SDK fix** (the actual bug):
- `Sources/JoyfillUI/View/Fields/TableView/TableModalTopNavigationView.swift` — replaced the custom `TextField` + `Binding<String>` block for `.text` cells with `TableTextRowFormView` (the same component the collection form already used). Removed the now-dead `@FocusState focusedColumnIndex` and `triggerInlineTextFocus()`.
- `Sources/JoyfillUI/View/Fields/TableView/TableTextRowFormView.swift` — swapped `TextEditor` → `TextField`. Removed the iOS-15/16 `#available` split (only existed to hide `TextEditor`'s scroll background).

**UI test migration** (consequence of TextEditor → TextField):
- `JoyfillSwiftUIExample/JoyfillUITests/Collection/CollectionFieldTests.swift`, `CollectionFieldSearchFilterTests.swift`, `ChangeUITests/OnChangeHandlerUITests.swift` — 18 lookups switched from `app.textViews[\"EditRowsTextFieldIdentifier\"]` to `app.textFields[...]`. Table tests already used `textFields` and stay.

**Example app — Changelog screen refactor** (DX, not SDK):
- `JoyfillSwiftUIExample/JoyfillExample/ChangeManager.swift` — replaced `displayedChangelogs: [String]` with structured `displayedEntries: [ChangelogEntry]`. One `onChange` call (even bulk with N changes) is now one entry.
- `JoyfillSwiftUIExample/JoyfillExample/ChangelogView.swift` — rewritten: kind badges, summary line, expandable detail with structured key/value pairs and a raw-JSON toggle, kind filter chips, search box, copy/export menu, toast.
- `FormContainerView.swift` / `UserAccessTokenTextFieldView.swift` — badge counters updated to read the new property.

## 3. Implementation decisions

- **Why use `TableTextRowFormView` for table too?** The collection form already used it correctly with focus/blur emission and an `onChange(of: text)`-only `didChange` path. The table form had a parallel custom binding whose setter fired `bulkEdit` on every render-driven refresh — root cause of the spurious rowUpdates. Unifying onto one component fixes the bug and removes a code-path divergence.
- **Why `TextField` and not `TextEditor`?** Row form cells are conceptually single-line. `TextEditor` shipped a built-in scroll background that needed `scrollContentBackground(.hidden)` (iOS 16+) plus a `UITextView.appearance()` workaround for iOS 15 — only ever there to make `TextEditor` look like a `TextField`. `TextField` removes the iOS version split entirely.
- **Why combine bulk-emit changes into one changelog card?** A single `onChange(changes:document:)` call is one event from the SDK's perspective; rendering each `Change` in the array as a separate UI card was misleading and made bulk-edit testing tedious. The new entry preserves the array internally (`items: [[String: Any]]`) and shows `×N` on the card.
- **Why a structured changelog type?** A `[String]` of pre-formatted JSON blobs forces every consumer to re-parse to filter, summarize, or group. A typed `ChangelogEntry { kind, timestamp, items }` makes the UI cheap and accurate, and the `copyText` computed property keeps the copy/export output unchanged.

## 4. Screenshot / video

Before vs after of the example app's changelog screen, plus a recording of the row form focus flow showing no spurious events, will be attached as a follow-up comment. Not embedded here because the change spans behavior + UI and a single still doesn't capture it.

## 5. Public API / docs

No public API change.
- `ChangeManager.displayedChangelogs: [String]` → `displayedEntries: [ChangelogEntry]` is an **example-app internal** property — not part of the SDK surface.
- `TableTextRowFormView` is internal to `JoyfillUI`.
- No SDK type, public method, or callback signature changed.
- No docs update needed.

## 6. Tests

- **UI tests:** 18 lookups migrated as part of this PR; selectors compile and resolve against the new element type. UI test suite has **not** been re-run end-to-end yet — any collection test that typed strings containing `\\n` into this cell may now assert against truncated text (`TextField` strips newlines). I'll flag specific failures and fix on the test side; the production behavior change is intentional.
- **Unit tests:** No SDK-level behavior changed in a way that requires new unit tests (the fix is wiring/binding, not logic). The existing `OnChangeHandlerUITests` covers the rowUpdate emission count end-to-end.
- **Manual:** verified focus → 1 `focus` event, 0 `rowUpdate`; typing → one `rowUpdate` per change as before; blur → 1 `blur` event.

## 7. Notes for reviewer

- **iOS 15 compatibility:** removing the `#available(iOS 16.0, *)` branch is safe — that branch only chose between two `TextEditor` variants. `TextField` works identically on iOS 15+.
- **Single-line input is a real UX change.** Row-form text cells can no longer accept `\\n`. If any customer JSON has pre-existing newlines in text-cell values, SwiftUI will display them collapsed when the user edits — the underlying value is unchanged unless the user types over it.
- **`focusedColumnIndex` / `triggerInlineTextFocus()` removed** because they were only bound to the now-deleted `TextField`'s `.focused(...)`. Auto-focus on form open continues to work via the existing `navigationFocusColumnId` environment value that `TableTextRowFormView.autoFocusIfNeeded()` reads on appear.

### DIFF
--- a/JoyfillSwiftUIExample/JoyfillExample/ChangeManager.swift
+++ b/JoyfillSwiftUIExample/JoyfillExample/ChangeManager.swift
@@ -7,13 +7,93 @@ import JoyfillModel
 import JoyfillAPIService
 import Joyfill
 
+/// Kinds of events surfaced in the test changelog view.
+enum ChangelogKind: String {
+    case change
+    case focus
+    case blur
+    case pageFocus
+    case pageBlur
+    case upload
+    case capture
+
+    var label: String {
+        switch self {
+        case .change:    return "Change"
+        case .focus:     return "Focus"
+        case .blur:      return "Blur"
+        case .pageFocus: return "Page Focus"
+        case .pageBlur:  return "Page Blur"
+        case .upload:    return "Upload"
+        case .capture:   return "Capture"
+        }
+    }
+}
+
+/// One entry per event callback. A single `onChange` invocation — including
+/// a bulk edit that produces many `Change` rows — collapses to ONE entry.
+struct ChangelogEntry: Identifiable {
+    let id = UUID()
+    let timestamp: Date
+    let kind: ChangelogKind
+    /// Structured payload entries displayed in the expanded view (each item is
+    /// rendered as its own row card). For `change`, this is the list of changes.
+    /// For single events (focus/blur/upload/capture), it has one item.
+    let items: [[String: Any]]
+
+    var count: Int { items.count }
+
+    /// Compact one-line summary shown on the collapsed card.
+    var summary: String {
+        switch kind {
+        case .change:
+            let targets = items.compactMap { $0["target"] as? String }
+            let uniqueTargets = Array(Set(targets)).sorted()
+            let fields = Array(Set(items.compactMap { $0["fieldId"] as? String })).sorted()
+            let rowsCount = items.reduce(0) { acc, item in
+                if let inner = item["change"] as? [String: Any] {
+                    if let rowIds = inner["rowIds"] as? [String] { return acc + rowIds.count }
+                    if inner["rowId"] != nil { return acc + 1 }
+                }
+                return acc
+            }
+            var parts: [String] = []
+            if !uniqueTargets.isEmpty { parts.append(uniqueTargets.joined(separator: ", ")) }
+            if !fields.isEmpty { parts.append("field: \(fields.first!)\(fields.count > 1 ? " (+\(fields.count - 1))" : "")") }
+            if rowsCount > 0 { parts.append("\(rowsCount) row\(rowsCount == 1 ? "" : "s")") }
+            return parts.joined(separator: " · ")
+        default:
+            let first = items.first ?? [:]
+            if let fieldID = first["fieldID"] as? String { return "field: \(fieldID)" }
+            if let pageDict = first["page"] as? [String: Any], let pageName = pageDict["name"] as? String {
+                return "page: \(pageName)"
+            }
+            return ""
+        }
+    }
+
+    /// Multi-line plain-text dump used by Copy / Export.
+    var copyText: String {
+        let ts = DateFormatter.timestamp.string(from: timestamp)
+        let header = "[\(ts)] \(kind.label)\(count > 1 ? " (\(count))" : "")"
+        let body = items.enumerated().map { idx, item -> String in
+            let prefix = count > 1 ? "  #\(idx + 1) " : "  "
+            let json = (try? JSONSerialization.data(withJSONObject: item, options: [.prettyPrinted, .sortedKeys]))
+                .flatMap { String(data: $0, encoding: .utf8) } ?? "<unserializable>"
+            return prefix + json.replacingOccurrences(of: "\n", with: "\n  ")
+        }.joined(separator: "\n")
+        return header + "\n" + body
+    }
+}
+
 class ChangeManager: ObservableObject {
     private let apiService: APIService?
     var showScan: (@escaping (ValueUnion) -> Void) -> Void
     private let showImagePicker:  (@escaping ([String]) -> Void) -> Void
-    
-    // Published property to display changelogs on screen
-    @Published var displayedChangelogs: [String] = []
+
+    /// Structured entries — one per event callback. Bulk `onChange` (many
+    /// changes in one call) shows up as a single entry with `items.count > 1`.
+    @Published var displayedEntries: [ChangelogEntry] = []
     @Published var showChangelogView: Bool = false
 
     init(apiService: APIService? = nil, showImagePicker:  @escaping(@escaping ([String]) -> Void) -> Void, showScan: @escaping (@escaping (ValueUnion) -> Void) -> Void) {
@@ -50,6 +130,12 @@ class ChangeManager: ObservableObject {
             }
         }
     }
+
+    private func append(_ entry: ChangelogEntry) {
+        DispatchQueue.main.async {
+            self.displayedEntries.append(entry)
+        }
+    }
 }
 
 extension ChangeManager: FormChangeEvent {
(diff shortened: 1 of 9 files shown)


Sources/JoyfillUI/View/Fields/TableView/TableModalTopNavigationView.swift
```diff
@@ -492,44 +491,11 @@ struct EditMultipleRowsSheetView: View {
...
-                                    .font(.system(size: 15))
-                                    .accessibilityIdentifier("EditRowsTextFieldIdentifier")
-                                    .padding(.horizontal, 10)
-                                    .frame(height: 40)
+                                TableTextRowFormView(cellModel: Binding.constant(cellModel), isUsedForBulkEdit: isUsedForBulkEdit)
+                                    .frame(minHeight: 40)
                                     .cellBorder(isFocused: isFocused)
-                                    .focused($focusedColumnIndex, equals: colIndex)
-                                
+                                    .accessibilityIdentifier("EditRowsTextFieldIdentifier")
```
vivekjyani, line 500: Duplicate accessibility identifier. `TableTextRowFormView` already sets `.accessibilityIdentifier("EditRowsTextFieldIdentifier")` directly on the inner `TextField` (TableTextRowFormView.swift:39). Re-applying it on the wrapper here adds an outer accessibility element with the same identifier, which can produce two matches for `app.textFields["EditRowsTextFieldIdentifier"]` in XCUI and made the readonly branch (the `Text` view) also pick up the same identifier — a problem since the old custom `TextField` was never used in readonly mode but this wrapper is. Suggest dropping this modifier and relying on the one inside `TableTextRowFormView`.

Sources/JoyfillUI/View/Fields/TableView/TableTextRowFormView.swift
```diff
@@ -34,40 +34,22 @@ struct TableTextRowFormView: View {
...
-                            updateFieldValue(newText: newValue)
-                        }
-                        .focused($isTextFieldFocused)
-                        .onChange(of: isTextFieldFocused) { focused in
-                            if focused {
-                                cellModel.didFocusBlur?(.focus, cellModel.data)
-                            } else {
-                                cellModel.didFocusBlur?(.blur, cellModel.data)
-                            }
+                TextField("", text: $text)
```
vivekjyani, line 38: Worth confirming: `_text = State(initialValue: cellModel.wrappedValue.data.title)` only runs on init. If `cellModel.data.title` is updated externally while this view is on-screen (e.g., a formula-driven recompute), `text` won't reflect it. In the modal path this is masked by the `viewID = UUID()` re-creation on row-navigation in `EditMultipleRowsSheetView`, so behavior matches the old `TextEditor` path — but if this component is ever embedded somewhere without that recreation, you'd see stale values. Not blocking; flagging so future callers know.

JoyfillSwiftUIExample/JoyfillExample/ChangeManager.swift
```diff
@@ -7,13 +7,93 @@ import JoyfillModel
...
+            let rowsCount = items.reduce(0) { acc, item in
+                if let inner = item["change"] as? [String: Any] {
+                    if let rowIds = inner["rowIds"] as? [String] { return acc + rowIds.count }
+                    if inner["rowId"] != nil { return acc + 1 }
+                }
+                return acc
+            }
+            var parts: [String] = []
+            if !uniqueTargets.isEmpty { parts.append(uniqueTargets.joined(separator: ", ")) }
+            if !fields.isEmpty { parts.append("field: \(fields.first!)\(fields.count > 1 ? " (+\(fields.count - 1))" : "")") }
```
vivekjyani: `fields.first!` is guarded by `!fields.isEmpty` so it's safe, but it reads as a footgun. `fields[0]` (still safe) or `if let first = fields.first` is friendlier to anyone scanning for force-unwraps.

JoyfillSwiftUIExample/JoyfillExample/ChangelogView.swift
```diff
@@ -2,146 +2,434 @@
...
+    @State private var searchText: String = ""
+    @State private var expandedEntryIDs: Set<UUID> = []
+    @State private var copyToast: String? = nil
+
+    private var filteredEntries: [ChangelogEntry] {
+        changeManager.displayedEntries.reversed().filter { entry in
+            if let k = selectedKind, entry.kind != k { return false }
+            if !searchText.isEmpty {
+                let needle = searchText.lowercased()
+                if entry.copyText.lowercased().contains(needle) { return true }
```
vivekjyani, line 21: `filteredEntries` is a computed property recomputed on every render, and it calls `entry.copyText` per entry — which does pretty-printed JSON serialization of every item. With a busy session (a bulk edit producing dozens of entries, plus filtering on each keystroke in the search field), this can get noticeably expensive. Two cheap mitigations: (a) lowercase `searchText` once outside the closure; (b) precompute a lighter searchable string on `ChangelogEntry` (e.g., joined `fieldId`/`target` values plus item count) rather than recomputing full JSON. Not a correctness issue, just an example-app DX nit since the goal of this refactor was to make the log readable during testing.

JoyfillSwiftUIExample/JoyfillExample/ChangelogView.swift
```diff
@@ -2,146 +2,434 @@
...
+
+    private var keyValues: [(String, String)] {
+        var pairs: [(String, String)] = []
+        if kind == .change, let inner = item["change"] as? [String: Any] {
+            let preferredOrder = ["rowId", "rowIds", "columnId", "value", "schemaId", "parentPath"]
+            for key in preferredOrder {
+                if let v = inner[key] { pairs.append((key, stringify(v))) }
+            }
+            for (k, v) in inner where !preferredOrder.contains(k) {
+                pairs.append((k, stringify(v)))
```
vivekjyani, line 344: `for (k, v) in inner where !preferredOrder.contains(k)` iterates a Swift dictionary, whose order isn't guaranteed. Two cards showing the same `change` payload can render their tail keys in different orders across renders. Wrap with `.sorted(by: { $0.key < $1.key })` for stable display. (Same applies to the `else` branch at line 348 if you ever extend it.)

vivekjyani (commented):
Nice fix — unifying the table form on `TableTextRowFormView` removes a real source of spurious `rowUpdate` emissions, and the bulk-edit accumulator semantics are preserved by `EditMultipleRowsSheetView.didChange` (lines 401–411). A few small comments inline. Most are example-app polish; the duplicate accessibility identifier is the one worth changing before merge.

Sources/JoyfillUI/View/Fields/TableView/TableModalTopNavigationView.swift
vivekjyani: Good catch — agreed on dropping the outer modifier; the inner one on TableTextRowFormView's TextField is the source of truth. To be precise: it doesn't actually produce two matches under app.textFields[...] (the wrapper isn't a textField element in XCUI), but you're right that in the readonly branch it would attach to the staticText, which is a latent footgun. Will remove in a follow-up commit.

Sources/JoyfillUI/View/Fields/TableView/TableTextRowFormView.swift
vivekjyani: Accurate, and worth flagging that this is pre-existing behavior — the TextEditor version had the same one-shot State(initialValue:). Today the viewID = UUID() reset on row navigation is what keeps it in sync. Leaving as-is for this PR since no current caller exercises the stale path; happy to close the gap with an onChange(of: cellModel.data.title) sync in a follow-up if/when we embed this view elsewhere.

JoyfillSwiftUIExample/JoyfillExample/ChangelogView.swift
vivekjyani: Fair point. Will precompute a lowercase searchHaystack on ChangelogEntry (joined target/fieldId/rowId/columnId/page.name/kind) at construction time, and lowercase searchText once outside the filter loop. That replaces the per-keystroke JSON serializations with a single contains per entry. Doing it as a follow-up commit on this branch.

JoyfillSwiftUIExample/JoyfillExample/ChangeManager.swift
vivekjyani: Done

JoyfillSwiftUIExample/JoyfillExample/ChangelogView.swift
vivekjyani: Fair flag, though in practice Swift 5 iterates a dictionary in the same order on every read within a process as long as the contents don't change — so within a session the keys stay stable on re-renders. The only variation is across app launches (different hash seeds), which isn't really observable on a debug screen. Happy to add the .sorted for cosmetic determinism since it's a one-liner, but flagging that it's not actually causing flicker today.

VishnuBishnoi (commented):

**Block-on-CI:** The PR description acknowledges "UI test suite has not been re-run end-to-end yet." Given 18 selector migrations + a real behavior change (`TextEditor` → `TextField` strips `\n`), please run the full UI suite before merge. Any collection/table test that types or asserts multi-line content will silently change behavior.

**Scope conflation.** The PR bundles two changes:
1. *Bug fix:* unify on `TableTextRowFormView` to eliminate the binding `set:` that fired on every render. This is the actual NO-2139 fix.
2. *UX change:* `TextEditor` → `TextField`, removing multi-line support from row-form text cells.

These are separable. The bug fix does not require dropping `TextEditor`. If multi-line is intentionally being removed, it deserves its own ticket and a customer-facing note — any existing JSON with newlines in table/collection text cells will display collapsed on edit. Consider splitting, or at minimum add a CHANGELOG entry.

**Missing regression test for the actual bug.** The PR says "existing `OnChangeHandlerUITests` covers the rowUpdate emission count end-to-end." Please point to (or add) the specific assertion that, after `textField.tap()` (focus only, no typing), the `rowUpdate` count is 0 and `focus` count is 1. If no such assertion exists today, this PR should add it — otherwise the regression that motivated the PR is uncovered.

**Auto-focus coverage.** `triggerInlineTextFocus()` / `focusedColumnIndex` are removed; auto-focus now relies on `TableTextRowFormView.autoFocusIfNeeded()` reading `navigationFocusColumnId`. Confirm there's a test that opens the row form with a column focus intent and asserts the text field is focused (keyboard up). If not, add one — easy regression surface.

**Cross-PR collision with #328:** After this PR lands, the table form's text cell renders as static `Text` in readonly, so `testTable_CellDecoratorInsideRowFormStaysTappable` in #328 will fail on the `("text", app.textFields["EditRowsTextFieldIdentifier"])` lookup. Coordinate with #328 author on which lands second and update that test to use absence-of-TextField (same pattern as the existing collection-root test).

**Minor — example-app polish:**
- `filteredEntries` recomputes `reversed() + filter` per body re-render. Fine at small N; with large logs it becomes O(N) per keystroke in the search box. Cache if you care.
- `ChangelogKind` raw values are derived from case names. The JSON export uses `kind.rawValue` — locking down explicit `String` raw values prevents silent breakage if a case is renamed.

PR #330, opened 2026-05-18
NO-2152: Set decorate flag from row decorators when missing in JSON

vivekjyani:
## Context
Backends may ship `rowDecorators` (or row-specific `decorators.all` on individual rows) without the corresponding `decorate: true` flag. Previously this rendered as a silent no-op — decorator data was loaded but the row decorator column stayed hidden, so the kebab menu never appeared. The fix infers `decorate` from the actual decorator payload at load time, only when JSON omits the flag.

## What changed
- `Sources/JoyfillUI/ViewModels/DocumentEditor+Decorators.swift`
  - New `normalizeDecorateFlag(field:)` (internal) — when `decorate` is missing, sets it from the actual data: `true` if any displayable row decorator exists (common or row-specific), `false` otherwise.
  - New private `hasAnyDisplayableRowDecorator(field:, schemaKey:)` — load-path scan. Walks raw JSON dicts (`field.dictionary["value"]`, `rowDict["children"][k]["value"]`, `rowDict["decorators"]["all"]`) and short-circuits on first match. Avoids materializing `[ValueElement]` via `valueToValueElements`, which would otherwise double load time on docs with thousands of rows since the view model later builds those structs again. Reuses `Decorator.isDisplayable` for the displayability check.
  - `updateFieldMap()` now runs the normalization on each field before storing into `fieldMap`. Covers both initial load and page duplication.
- `JoyfillSwiftUIExample/JoyfillTests/DecoratorAPI/DecoratorPublicAPITests.swift`
  - 14 new tests under `// MARK: - Decorate flag normalization at load`.

## Behavior

| `decorate` in JSON | Decorators present | Result |
|---|---|---|
| `nil` | yes | `true` |
| `nil` | no | `false` |
| `true` | any | stays `true` |
| `false` | any | stays `false` |

After first load, `decorate` is always a decided boolean — no `nil` survives.

## Decisions
- **Only fill `nil`, never override `true`/`false`.** Explicit values in JSON are intentional state and must be respected. The scan only runs when the flag is absent — making this a fill-in-the-blank, not a re-derivation.
- **Always settle to a decided boolean.** When JSON omits `decorate` and the scan finds nothing, write `false` (rather than leaving `nil`). This makes every subsequent load idempotent — the guard sees `decorate != nil` and skips the row scan entirely. Cost moves to first-load only.
- **Mutate (not compute).** Setting `decorate` in `fieldMap` matches the existing pattern from `ensureDecorateEnabled` (write-API path). Single source of truth — downstream readers (`tableDataModel.decorate`, `schema[sk].decorate`) don't need to know normalization happened.
- **Normalize at `updateFieldMap`** (not in TableDataModel/CollectionViewModel init). Centralizes the fix at the single entry point where fields flow from raw JSON into the editor's working state. No coupling to view models.
- **Per-schema for collections.** Each schema's `decorate` is evaluated independently from rows that belong to that schema, since the row decorator column is a per-schema concern.
- **Never flip `true → false` at load.** That direction stays the responsibility of `ensureDecorateDisabledIfEmpty` after explicit removals via the API.

## Screenshot / video
N/A — pure model-layer fix; no UI/visual change beyond the row decorator column now appearing for payloads where it was previously hidden.

## Public API
No public API changes. `normalizeDecorateFlag` is `internal`; `decorate` / `rowDecorators` shapes are unchanged. No doc updates needed.

## Tests
14 cases covering: table (common, row-specific, nil/true/false flag, non-displayable, empty), collection (root common, child common, root row-specific, child row-specific in nested schema, both schemas empty, explicit true, explicit false), and idempotence (subsequent `updateFieldMap` calls don't re-flip or re-scan; explicit assertion on `editor.document.fields` locks the write-back via `fieldMap.didSet`). No follow-up tests planned.

## Notes for reviewer
- First load: O(rows) scan per table/collection field with `decorate == nil`. After that, every field has a decided boolean → all subsequent loads (page duplication, etc.) skip the scan via the guard.
- **Perf:** The load-path scan operates on raw JSON dicts rather than the typed `valueToValueElements` accessor. Verified on a 10k-row template: load times stay at baseline; without this, normalization added ~2.8s of duplicate `ValueElement` construction.
- Pre-existing concern (not addressed here): `updateFieldMap` on page duplication re-reads from `document.fields`, which can overwrite in-memory decorator changes made via the API. Worth flagging separately if it bites.

### DIFF
--- a/JoyfillSwiftUIExample/JoyfillTests/DecoratorAPI/DecoratorPublicAPITests.swift
+++ b/JoyfillSwiftUIExample/JoyfillTests/DecoratorAPI/DecoratorPublicAPITests.swift
@@ -713,7 +713,9 @@ final class DecoratorPublicAPITests: XCTestCase {
         XCTAssertEqual(editor.field(fieldID: "tbl-1")?.decorate, true)
     }
 
-    func testNormalize_table_noDecoratorsAnywhere_decorateStaysNil() {
+    func testNormalize_table_noDecoratorsAnywhere_decorateBecomesFalse() {
+        // When JSON omits the flag and the scan finds nothing, we write `false`
+        // as a decided sentinel so subsequent loads skip the scan.
         let dict = makeMinimalJoyDocDict(fields: [[
             "_id": "tbl-1",
             "type": "table",
@@ -721,7 +723,7 @@ final class DecoratorPublicAPITests: XCTestCase {
             "value": [["_id": "r1"]],
         ]])
         let editor = makeNormalizationEditor(dict: dict)
-        XCTAssertNil(editor.field(fieldID: "tbl-1")?.decorate)
+        XCTAssertEqual(editor.field(fieldID: "tbl-1")?.decorate, false)
     }
 
     func testNormalize_table_decorateTrueButNoDecorators_staysTrue() {
@@ -751,8 +753,9 @@ final class DecoratorPublicAPITests: XCTestCase {
         XCTAssertEqual(editor.field(fieldID: "tbl-1")?.decorate, false)
     }
 
-    func testNormalize_table_nonDisplayableRowDecorator_doesNotFlip() {
-        // Decorator with no icon AND no label is not displayable.
+    func testNormalize_table_nonDisplayableRowDecorator_becomesFalse() {
+        // Decorator with no icon AND no label is not displayable, so the scan
+        // treats it as "nothing to show" and the flag settles at false.
         let dict = makeMinimalJoyDocDict(fields: [[
             "_id": "tbl-1",
             "type": "table",
@@ -761,7 +764,7 @@ final class DecoratorPublicAPITests: XCTestCase {
             "rowDecorators": [["action": "a", "color": "#3B82F6"]],
         ]])
         let editor = makeNormalizationEditor(dict: dict)
-        XCTAssertNotEqual(editor.field(fieldID: "tbl-1")?.decorate, true)
+        XCTAssertEqual(editor.field(fieldID: "tbl-1")?.decorate, false)
     }
 
     // -- Collection
@@ -771,14 +774,14 @@ final class DecoratorPublicAPITests: XCTestCase {
         let editor = makeNormalizationEditor(dict: dict)
         let schema = editor.field(fieldID: "col-1")?.schema
         XCTAssertEqual(schema?["root"]?.decorate, true)
-        XCTAssertNotEqual(schema?["child"]?.decorate, true)
+        XCTAssertEqual(schema?["child"]?.decorate, false)
     }
 
     func testNormalize_collection_childCommonRowDecorators_noFlag_flipsChildOnly() {
         let dict = makeCollectionDict(childRowDecorators: [decoratorDict(action: "a")])
         let editor = makeNormalizationEditor(dict: dict)
         let schema = editor.field(fieldID: "col-1")?.schema
-        XCTAssertNotEqual(schema?["root"]?.decorate, true)
+        XCTAssertEqual(schema?["root"]?.decorate, false)
         XCTAssertEqual(schema?["child"]?.decorate, true)
     }
 
@@ -787,7 +790,7 @@ final class DecoratorPublicAPITests: XCTestCase {
         let editor = makeNormalizationEditor(dict: dict)
         let schema = editor.field(fieldID: "col-1")?.schema
         XCTAssertEqual(schema?["root"]?.decorate, true)
-        XCTAssertNotEqual(schema?["child"]?.decorate, true)
+        XCTAssertEqual(schema?["child"]?.decorate, false)
     }
 
     func testNormalize_collection_childRowSpecificDecorator_noFlag_flipsChildOnly() {
@@ -796,16 +799,17 @@ final class DecoratorPublicAPITests: XCTestCase {
         let dict = makeCollectionDict(childRowSpecific: [decoratorDict(action: "row-a")])
         let editor = makeNormalizationEditor(dict: dict)
         let schema = editor.field(fieldID: "col-1")?.schema
-        XCTAssertNotEqual(schema?["root"]?.decorate, true)
+        XCTAssertEqual(schema?["root"]?.decorate, false)
         XCTAssertEqual(schema?["child"]?.decorate, true)
     }
 
-    func testNormalize_collection_noDecoratorsAnywhere_bothSchemasStayUnset() {
+    func testNormalize_collection_noDecoratorsAnywhere_bothSchemasBecomeFalse() {
+        // No decorators anywhere → both schemas settle at false, not nil.
         let dict = makeCollectionDict()
         let editor = makeNormalizationEditor(dict: dict)
         let schema = editor.field(fieldID: "col-1")?.schema
-        XCTAssertNotEqual(schema?["root"]?.decorate, true)
-        XCTAssertNotEqual(schema?["child"]?.decorate, true)
+        XCTAssertEqual(schema?["root"]?.decorate, false)
+        XCTAssertEqual(schema?["child"]?.decorate, false)
     }
 
     func testNormalize_collection_schemaDecorateExplicitlyFalseWithDecorators_staysFalse() {
@@ -814,5 +818,36 @@ final class DecoratorPublicAPITests: XCTestCase {
         let editor = makeNormalizationEditor(dict: dict)
         let schema = editor.field(fieldID: "col-1")?.schema
         XCTAssertEqual(schema?["root"]?.decorate, false)
+    }
+
+    func testNormalize_collection_schemaDecorateTrueButNoDecorators_staysTrue() {
+        // Mirror of the table case: explicit per-schema `decorate: true` with no
+        // row decorators must stay true. Normalization only fills in nil.
+        let dict = makeCollectionDict(rootDecorate: true)
+        let editor = makeNormalizationEditor(dict: dict)
+        let schema = editor.field(fieldID: "col-1")?.schema
+        XCTAssertEqual(schema?["root"]?.decorate, true)
+    }
+
+    // -- Stability across repeat updateFieldMap calls
+
+    func testNormalize_isIdempotent_subsequentUpdateFieldMapDoesNotReFlip() {
+        // Locks the contract that the inferred value is written back to
+        // `document.fields`, so a second `updateFieldMap` call sees the
+        // already-set flag and short-circuits without running the row scan.
+        let dict = makeMinimalJoyDocDict(fields: [[
+            "_id": "tbl-1",
+            "type": "table",
+            "rowOrder": ["r1"],
+            "value": [["_id": "r1"]],
+            "rowDecorators": [decoratorDict(action: "a")],
+        ]])
+        let editor = makeNormalizationEditor(dict: dict)
+        XCTAssertEqual(editor.field(fieldID: "tbl-1")?.decorate, true)
+
+        // Re-running normalization should be a no-op now that document.fields
+        // also carries decorate=true.
+        editor.updateFieldMap()
+        XCTAssertEqual(editor.field(fieldID: "tbl-1")?.decorate, true)
     }
 }
--- a/JoyfillSwiftUIExample/JoyfillTests/DecoratorAPI/DecoratorPublicAPITests.swift
+++ b/JoyfillSwiftUIExample/JoyfillTests/DecoratorAPI/DecoratorPublicAPITests.swift
@@ -820,6 +820,41 @@ final class DecoratorPublicAPITests: XCTestCase {
         XCTAssertEqual(schema?["root"]?.decorate, false)
     }
 
+    func testNormalize_collection_grandchildRowSpecificDecorator_flipsOnlyDeepestSchema() {
+        // 3-level chain: root → child → grandchild. Row decorator lives on a
+        // grandchild row only. Locks that `scanRawRowsForDisplayableRowDecorator`
+        // recurses past depth 1 — only the grandchild schema flips to true;
+        // root and child stay false.
+        let grandchildRow: [String: Any] = [
+            "_id": "gr1",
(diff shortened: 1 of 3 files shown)


Sources/JoyfillUI/ViewModels/DocumentEditor.swift
```diff
@@ -151,7 +151,9 @@ public class DocumentEditor: ObservableObject {
     public func updateFieldMap() {
         document.fields.forEach { field in
             guard let fieldID = field.id else { return }
-            self.fieldMap[fieldID] =  field
+            var f = field
+            _ = normalizeDecorateFlag(field: &f)
+            self.fieldMap[fieldID] = f
```
vivekjyani, line 156: The normalized flag is written to `fieldMap` but never back to `document.fields`. Since `JoyDocField` is a `struct`, the `field` in `document.fields.forEach { field in ... }` above is a value-type copy — mutating `f` and storing it in `fieldMap` does not update `document.fields[i]`. On every subsequent `updateFieldMap()` call (the page-duplication path at line 1153), the loop reads `decorate == nil` from `document.fields` again and re-runs the row scan. The result is correct (same value), but the "idempotent — guard skips the row scan" claim in the PR description / docstring on `normalizeDecorateFlag` is not actually true.

If the intent is to make subsequent loads short-circuit, you also need to write `f` back to `document.fields` here (e.g. by indexing). If not, the docs should be reworded so future readers don't rely on a guarantee the code doesn't provide.

Sources/JoyfillUI/ViewModels/DocumentEditor+Decorators.swift
```diff
@@ -313,6 +315,54 @@ private extension DocumentEditor {
...
+    // MARK: Decorate normalization
+
+    /// Infers an effective `decorate` flag from the actual decorator data on a
+    /// freshly loaded field, but only when the JSON omits the flag entirely.
+    /// Explicit `true` / `false` is always respected — only `nil` triggers a
+    /// row scan. When the flag is absent we decide it from the data: `true`
+    /// if any displayable row decorator exists (common or row-specific),
+    /// `false` otherwise. Writing `false` (rather than leaving nil) makes
+    /// subsequent loads idempotent — the guard sees the decided value and
+    /// skips the scan.
```
vivekjyani, line 327: > Writing `false` (rather than leaving nil) makes subsequent loads idempotent — the guard sees the decided value and skips the scan.

This only holds if the decided value is persisted to `document.fields`, but `updateFieldMap()` only writes to `fieldMap`. On the next load, `document.fields` still has `decorate == nil` and the scan runs again. Same caveat applies to the new line added to `ensureDecorateDisabledIfEmpty`'s docstring (around line 235): `false` is a downstream readability/contract choice, not a perf optimization, until `document.fields` is kept in sync. Suggest either persisting the value or rewording both comments.

Sources/JoyfillUI/ViewModels/DocumentEditor+Decorators.swift
```diff
@@ -313,6 +315,54 @@ private extension DocumentEditor {
...
+            return true
+        case .collection:
+            guard var schema = field.schema else { return false }
+            var changed = false
+            for (key, var entry) in schema {
+                guard entry.decorate == nil else { continue }
+                entry.decorate = hasAnyDisplayableRowDecorator(field: field, schemaKey: key)
+                schema[key] = entry
+                changed = true
+            }
```
vivekjyani, line 344: Nit (not blocking): for a collection field with N schemas all missing `decorate`, this calls `hasAnyDisplayableRowDecorator` per schema, and each call invokes `rowsInScope`, which walks the entire row tree from the top. So worst case is O(N · tree). A single tree walk that records `[schemaKey: Bool]` would collapse it to one pass. Probably fine in practice (only runs on first load, only on schemas with `nil` flag), but worth flagging if normalization ever shows up in a profile.

JoyfillSwiftUIExample/JoyfillTests/DecoratorAPI/DecoratorPublicAPITests.swift
```diff
@@ -594,4 +594,260 @@ final class DecoratorPublicAPITests: XCTestCase {
...
+    }
+
+    // -- Stability across repeat updateFieldMap calls
+
+    func testNormalize_isIdempotent_subsequentUpdateFieldMapDoesNotReFlip() {
+        // Locks the contract that the inferred value is written back to
+        // `document.fields`, so a second `updateFieldMap` call sees the
+        // already-set flag and short-circuits without running the row scan.
+        let dict = makeMinimalJoyDocDict(fields: [[
+            "_id": "tbl-1",
```
vivekjyani, line 947: The comments here ("inferred value is written back to `document.fields`" and "no-op now that document.fields also carries decorate=true") describe a contract the implementation does not satisfy — `updateFieldMap()` only writes to `fieldMap`. The second `updateFieldMap()` call here re-walks the rows from scratch and re-derives `true`; it doesn't short-circuit on the guard.

The test passes either way because the scan converges, but the comment misrepresents what is being verified. Suggest one of:

1. Actually lock the documented contract — mutate `document.fields` in `updateFieldMap()`, then assert `editor.document.fields.first?.decorate == true` after the first call.
2. Or rename + rewrite the comment to reflect what is really being tested: that repeated `updateFieldMap()` calls converge to the same flag.

As written, a future change that breaks the idempotence-by-write-back invariant would not be caught here.

vivekjyani (commented):
Solid fix — load-time normalization is in the right place, test matrix is thorough. One substantive concern about the **idempotence rationale**: the PR description, the docstring on `normalizeDecorateFlag`, the new comment on `ensureDecorateDisabledIfEmpty`, and the comment in `testNormalize_isIdempotent_*` all claim writing `false` instead of `nil` lets subsequent loads skip the row scan via the `decorate != nil` guard. But the normalized value is only written to `fieldMap`, not back to `document.fields` — and `updateFieldMap()` iterates `document.fields` (value-type copies). So on every subsequent call (e.g. page duplication at DocumentEditor.swift:1153) the flag is still `nil`, the guard does not trip, and the scan re-runs. Behavior stays correct because the rescan converges, but the optimization is not real and the test does not actually lock the stated contract. Either persist the decided value to `document.fields` or align docs/tests with reality. Inline comments mark each spot, plus one minor perf nit on collections with many schemas.

Sources/JoyfillUI/ViewModels/DocumentEditor.swift
vivekjyani: document.fields is kept in sync via fieldMap's didSet (DocumentEditor.swift:67-71):

var fieldMap = [String: JoyDocField]() {
    didSet { document.fields = allFields }
}
Every fieldMap[id] = f inside the loop triggers didSet → document.fields is rebuilt from allFields. So the normalized struct lands in document.fields[i] too, and the next updateFieldMap() call reads the decided boolean and short-circuits at the guard.

Sources/JoyfillUI/ViewModels/DocumentEditor+Decorators.swift
vivekjyani: `fieldMap`'s `didSet` runs `document.fields = allFields` on every assignment ([DocumentEditor.swift:67-71](https://github.com/joyfill/components-swift/blob/NO-2152/Sources/JoyfillUI/ViewModels/DocumentEditor.swift#L67-L71)), so the decided `decorate` value does land in `document.fields`. Subsequent loads short-circuit at the guard as documented.

Sources/JoyfillUI/ViewModels/DocumentEditor+Decorators.swift
vivekjyani: Fixed

JoyfillSwiftUIExample/JoyfillTests/DecoratorAPI/DecoratorPublicAPITests.swift
vivekjyani: Fixed

VishnuBishnoi (commented):

**On the existing reviewer's idempotence concern — actually disagree.**
`DocumentEditor.swift:67-71` declares `fieldMap` with a `didSet` that writes `document.fields = allFields`. So when `updateFieldMap()` does `self.fieldMap[fieldID] = f`, the normalized `decorate: false` IS persisted back to `document.fields`. The PR's docstring claim and the `XCTAssertEqual(editor.document.fields.first…)` line in `testNormalize_isIdempotent_subsequentUpdateFieldMapDoesNotReFlip` are both correct as written. The optimization is real. The other reviewer should re-check the didSet before requesting changes on this point.

**Other findings:**

1. **`field.dictionary` is the single point of failure.** `hasAnyDisplayableRowDecorator` reads `field.dictionary["value"]` / `["schema"]` / row `["children"]` / row `["decorators"]["all"]`. This assumes `JoyDocField.dictionary` stays in lockstep with the typed model. If `dictionary` is ever (a) computed lazily from typed fields and gets out of sync after a mutation, or (b) removed entirely in a future refactor, normalization silently returns the wrong answer. Add a comment at the function pinning the contract, and consider one test that mutates the field via the typed API (e.g. `setDecorators`) and re-runs `updateFieldMap` to lock parity.

2. **Collection root-schema detection is fragile.** `rootSchemaKey` is the first schema dict where `"root": true`. Two latent issues:
   - Dict iteration order is undefined — if a malformed JSON marks two schemas as root, the chosen one is non-deterministic.
   - If no schema is marked root, `rootSchemaKey == nil` and the top-level `rowDicts` won't match any `targetSchemaKey != nil`, so child schemas that only have root-level rows will incorrectly settle to `false`.
   At minimum document the precondition; ideally validate (assert/log) when zero or >1 root schemas are present.

3. **Recursion depth coverage gap.** Tests cover root + one child level. `scanRawRowsForDisplayableRowDecorator` recurses arbitrarily, but depth ≥ 2 is uncovered. Add a test with root → child → grandchild where the decorator lives at the deepest level, and assert only the grandchild schema's `decorate` flips.

4. **Mixed-state per-schema test missing.** The current collection tests are either all-nil or all-explicit. Add: root with explicit `decorate: false`, child with `decorate: nil` plus row-specific decorators — assert root stays false, child flips to true. Locks the partial-fill loop in the `.collection` branch.

5. **`Decorator.isDisplayable` is an external dependency.** `testNormalize_table_nonDisplayableRowDecorator_becomesFalse` relies on `{action, color}`-only returning `isDisplayable == false`. If `isDisplayable`'s definition is ever loosened (e.g. to include "has an action"), this test silently passes for the wrong reason. Pin the behavior with a direct `Decorator(dictionary: …).isDisplayable` unit test in the same file.

6. **Wire-schema impact of `nil → false`.** Before: `decorate: nil` was omitted from serialized JSON. After: every loaded field has a decided boolean, so every save round-trips an explicit `decorate: false` even when JSON shipped without the key. If any backend distinguishes "missing" from "explicit false" (default-applying, version detection, audit logs), that semantic shift is now baked in. Worth confirming with the backend/web SDK team.

7. **Style nit — access-level mixing.** `internal func normalizeDecorateFlag` sits inside a `private extension`. Swift allows the override (so the cross-file call from `DocumentEditor.swift:153` compiles), but mixing visibility levels in one extension is harder to read. Either pull it into its own `extension DocumentEditor` block or keep all members at the extension's default.

8. **Perf claim ("10k rows stays at baseline") isn't locked by a test.** Optional, but a smoke perf test (XCTest `measure { editor.updateFieldMap() }` against a large fixture) would prevent silent regressions if the scan ever stops short-circuiting or starts materializing typed structs.

Otherwise: scope is right, table-vs-collection paths are mirrored, "fill-only-nil, never override true/false" is the correct semantic. LGTM after #1–#3 are addressed.

vivekjyani:
Review follow-ups — addressed

Thanks for the second pass. Walking through each point:

1. Idempotence concern (defending vs the earlier reviewer) — agreed, fieldMap.didSet → document.fields = allFields makes the write-back real. Strengthened the idempotence test with an explicit assertion on editor.document.fields.first(where: …)?.decorate after the first call, so the contract is now directly verifiable.

2. field.dictionary as a single point of failure — added a Contract: paragraph on hasAnyDisplayableRowDecorator's docstring documenting the raw-dict dependency, plus a parity test (testNormalize_typedAPIWrite_persistsToRawValueDict_forFreshLoad) that mutates via the public API, strips the inferred flag, reloads, and asserts the raw-dict scan still finds the typed-API-added decorator.

3. Root-schema fragility — added an inline comment on the rootSchemaKey lookup noting the JoyDoc invariant of exactly one schema with root: true. Defensive validation skipped since malformed docs don't reach this path.

4. Recursion depth ≥ 2 uncovered — added testNormalize_collection_grandchildRowSpecificDecorator_flipsOnlyDeepestSchema. 3-schema chain (root → child → grandchild), decorator on the grandchild row only, asserts only the grandchild flips while root and child stay false.

5. Mixed-state per-schema — added testNormalize_collection_mixedFlagStates_eachSchemaEvaluatedIndependently. Root explicit false + child nil with row-specific decorator. Locks the partial-fill loop.

6. Decorator.isDisplayable external dependency — added testDecoratorIsDisplayable_pinsCurrentDefinition covering the full truth table (icon-only, label-only, both, action-only, action+color, empty strings, empty dict). Any future loosening fires this immediately.

7. Wire-schema impact of nil → false — flagging separately so the backend/web team can confirm there's no logic that distinguishes "key missing" from "key present and false." Posting that as its own comment on the thread.

8. internal in private extension style nit — intentional. Function lives next to the other decorator helpers; moved out and then back during earlier review iterations. Happy to revisit in a follow-up if it keeps bothering reviewers.

9. Perf smoke test — skipped for now. Deferring to a separate task if normalization ever shows up in a profile; current verification on the 10k-row template is captured in the PR description.

Test count is now 18 in the normalization section. LGTM gate met from your side after [#1](https://github.com/joyfill/components-swift/issues/1)–[#3](https://github.com/joyfill/components-swift/issues/3).

PR #331, opened 2026-05-26
NO-2175: Treat empty field title as missing for decorator padding

vivekjyani:
## 1. Context
Field headers can render with decorators (info icons, badges, etc.) on the trailing edge. When a field has **no title**, the decorators need an extra 8pt bottom padding to stay visually balanced inside the field row. Previously this padding only kicked in when `title == nil` — but in practice the backend often sends `title: ""` (empty string) for "no title", which is functionally the same as missing.

## 2. What changed
- [Sources/JoyfillUI/View/FieldHeaderView.swift:45](Sources/JoyfillUI/View/FieldHeaderView.swift#L45)
  - Before: `.padding(.bottom, fieldHeaderModel?.title == nil ? 8 : 0)`
  - After:  `.padding(.bottom, (fieldHeaderModel?.title?.isEmpty ?? true) ? 8 : 0)`

A field with an empty-string title now gets the same 8pt bottom padding as one with a nil title.

## 3. Design decision
Chose `(title?.isEmpty ?? true)` over normalizing `title` to nil upstream because:
- The padding rule is a **view concern** — the model intentionally keeps `title` as an optional `String?` since callers/decoders may distinguish "absent" vs "explicitly blank" elsewhere.
- Keeps the fix localized to one line; no model/decoder changes, no risk to other consumers of `FieldHeaderModel.title`.

## 4. Visual
One-line padding tweak; visible only when a field has decorators **and** an empty/missing title. Decorators now sit at the same vertical offset whether `title` is `nil` or `""`. No screenshot attached — diff is a single boolean condition.

## 5. Public API
No public API changes. `FieldHeaderView` and `FieldHeaderModel` signatures are untouched. No docs update needed.

## 6. Tests
No new tests added in this PR — the change is a purely visual padding condition driven by a single boolean. Existing snapshot/UI coverage for `FieldHeaderView` continues to pass. Happy to add a targeted snapshot test if the reviewer prefers.

## 7. Notes for reviewer
- Single-line behavioral change; rest of the file is unchanged.
- The adjacent `.environment(\.isEnabled, true)` and its comment were touched in an earlier commit on this branch and are not part of this change.

### DIFF
--- a/Sources/JoyfillUI/View/FieldHeaderView.swift
+++ b/Sources/JoyfillUI/View/FieldHeaderView.swift
@@ -12,7 +12,11 @@ struct FieldHeaderView: View {
     let fieldHeaderModel: FieldHeaderModel?
     let isFilled: Bool
     var onDecoratorTap: ((DecoratorLocal) -> Void)?
-    
+
+    private var hasNoTitle: Bool {
+        fieldHeaderModel?.title?.isEmpty ?? true
+    }
+
     public init(_ fieldHeaderModel: FieldHeaderModel?, isFilled: Bool = false, onDecoratorTap: ((DecoratorLocal) -> Void)? = nil) {
         self.fieldHeaderModel = fieldHeaderModel
         self.isFilled = isFilled
@@ -42,7 +46,7 @@ struct FieldHeaderView: View {
                 ) { decorator in
                     onDecoratorTap?(decorator)
                 }
-                .padding(.bottom, (fieldHeaderModel?.title?.isEmpty ?? true) ? 8 : 0)
+                .padding(.bottom, hasNoTitle ? 8 : 0)
                 // Re-enable the header even when the parent field is .disabled() —
                 // decorators must stay interactive on readonly forms.
                 .environment(\.isEnabled, true)
--- a/Sources/JoyfillUI/View/FieldHeaderView.swift
+++ b/Sources/JoyfillUI/View/FieldHeaderView.swift
@@ -42,7 +42,7 @@ struct FieldHeaderView: View {
                 ) { decorator in
                     onDecoratorTap?(decorator)
                 }
-                .padding(.bottom, fieldHeaderModel?.title == nil ? 8 : 0)
+                .padding(.bottom, (fieldHeaderModel?.title?.isEmpty ?? true) ? 8 : 0)
                 // Re-enable the header even when the parent field is .disabled() —
                 // decorators must stay interactive on readonly forms.
                 .environment(\.isEnabled, true)


Sources/JoyfillUI/View/FieldHeaderView.swift
```diff
@@ -42,7 +42,7 @@ struct FieldHeaderView: View {
                 ) { decorator in
                     onDecoratorTap?(decorator)
                 }
-                .padding(.bottom, fieldHeaderModel?.title == nil ? 8 : 0)
+                .padding(.bottom, (fieldHeaderModel?.title?.isEmpty ?? true) ? 8 : 0)
```
vivekjyani: **Fix is incomplete — the title rendering on line 24 still treats empty string as a present title.**

The new condition correctly applies the 8pt padding when `title` is `nil` *or* `""`, but a few lines above (line 24-27):

```swift
if let title = fieldHeaderModel?.title {
    Text("\(title)")
        .font(.headline.bold())
}
```

…still enters the branch for `title == ""` and renders an empty `Text("")`. SwiftUI keeps that empty `Text` as a layout participant in the `HStack` (it still contributes baseline + line-height metrics from `.font(.headline.bold())`), so the header row's vertical centering can be subtly different between `title: nil` and `title: ""` — which is exactly the inconsistency this PR is trying to eliminate.

Two ways to make the behavior fully symmetric:

```swift
// Option A — guard at the title site
if let title = fieldHeaderModel?.title, !title.isEmpty {
    Text(title)
        .font(.headline.bold())
}
```

```swift
// Option B — single derived flag, used by both branches
let hasTitle = !(fieldHeaderModel?.title?.isEmpty ?? true)
…
if hasTitle { Text(fieldHeaderModel!.title!) … }
…
.padding(.bottom, hasTitle ? 0 : 8)
```

Option A is the smallest change and matches the PR's "keep it localized" goal — recommend pairing it with the line 45 change so both the title and the decorator padding agree on what "no title" means. Otherwise the padding decision on line 45 and the title visibility on line 24 disagree whenever the backend sends `title: ""`.

vivekjyani: Re-checked with the author — this concern doesn't actually apply. The padding on line 45 is only on the decorator, and its purpose is just to balance the decorator vertically when no visible title text is taking up space. The `HStack` uses `.center` alignment and the decorator is the tallest item, so an empty `Text("")` rendered on line 24 doesn't change row height or alter the layout the decorator sits in. The new `(title?.isEmpty ?? true)` condition correctly fires the 8pt padding for both `nil` and `""`, which is exactly the rule needed. No change required — fix as written is correct.

VishnuBishnoi: Nit: the optional-chaining double-negative reads slightly awkwardly. A tiny computed helper (or `title?.isEmpty != false`) would be a bit clearer, but the current form is correct and equivalent.

Also worth confirming: are there any callsites that *intentionally* pass an empty string to mean "render an empty title slot but keep the field tall"? If so, this change silently shrinks that layout. From a quick read of `FieldHeaderView` I don't see such a path, but flagging since the previous behavior distinguished `nil` from `""`.

VishnuBishnoi (commented):
## Review summary — NO-2175

**Scope:** 1-line padding fix in `FieldHeaderView` to treat `title == ""` the same as `title == nil`.

**Correctness:** Looks correct. `(title?.isEmpty ?? true)` is `true` when title is nil OR empty, matching the stated intent.

**Risk:** Low — single boolean condition, change is localized. The only risk is if any caller intentionally distinguishes "" from nil for layout purposes (see inline comment).

**Tests:** None added. Given the change is a single visual condition driven by an existing property, a snapshot test for the empty-title-with-decorators case would be cheap insurance against regression but is not blocking.

**Style:** The double-negative `isEmpty ?? true` reads a bit awkwardly; see inline nit.

**Verdict:** Approve with minor optional nit. Safe to merge.

Sources/JoyfillUI/View/FieldHeaderView.swift
vivekjyani: Fixed 1st point 
On the second point — note the padding only applies inside the if !model.decorators.isEmpty branch, so the "silent layout shrink" only matters for callers that pass decorators and intentionally use "" (not nil) to push them up. Grepped both FieldHeaderModel construction sites (DocumentEditor.swift:664, 838) — both use title: showTitle ? fieldData?.title : nil, so nil is the only deliberate "no title" sentinel. No regression.

VishnuBishnoi (approved):
Approving — `hasNoTitle` extraction addressed the readability nit cleanly. Change is minimal and safe.

PR #332, opened 2026-05-28
NO-2143: Table performance — O(N²) → O(N) lookups, dead code cleanup, fix UI/data drift

vivekjyani:
## 1. Context

Ticket: NO-2143. Pass over the table editing hot paths (bulk edit, insert-below, delete-rows, cell change) to remove repeated linear scans, redundant publishes, and a couple of latent correctness issues. Large tables (hundreds of rows × many columns) showed visible jank on these flows; profiling pointed at nested `firstIndex(where:)` lookups and per-row `@Published` writes during reindex.

## 2. What changed

- **`DocumentEditor+ChangeHandler.swift`**
  - `deleteRows`: build a `rowID → index` map once and use a `Set`-based single-pass cleanup of `rowOrder`, replacing per-row `firstIndex` + per-row `removeAll` (was O(N·M), now O(N+M)). Also switched a fatal `return elements` on missing row to `Log(.error) + continue` so a single bad ID no longer aborts the whole delete.
  - `rowUpdateEvent`: takes the live `[ValueElement]` from the caller instead of re-fetching from `fieldMap` and doing a `firstIndex(of: rowID)` on `rowOrder`. The index wasn't actually used afterwards, so the lookup is gone entirely.
  - `insertBelowAdditional`: now takes `fieldData: [ValueElement]` from the caller instead of re-reading `field(fieldID:)?.valueToValueElements`, keeping it consistent with the row the caller already has in hand.
- **`TableViewModel.swift`**
  - Bulk edit: precompute `rowIndexMap` (`rowID → index in cellModels`) once and reuse it for both `makeChangeDict` and the post-edit `updatedModels` patch. Replaces two nested `firstIndex(where:)` scans per selected row.
  - `insertBelow`: append the newly inserted element to `tableDataModel.valueToValueElements` so the in-memory value array no longer drifts from the cell models / row order after insert.
  - Removed unused `isColumnFilled(columnId:)` and a no-op `sortElementsByRowOrder` call (the function returned a dictionary internally, so the \"sort\" was discarded — confirmed dead).
- **`TableModalView.swift` / `CollectionViewModel.swift`**
  - Moved the required-field indicator from the column header into the cell. The header version was being re-laid-out on every horizontal scroll tick; per-cell rendering is cheaper and matches how the rest of the validation UI works.
- **`Models.swift`**: dropped unused `longestBlockText` helper.
- **Reindexing**: `reIndexingRows` now mutates a local copy and publishes once instead of N times.
- **Tests**: added `testinsertbelowthenbulkedit` in [TableFieldUITestCases.swift](JoyfillSwiftUIExample/JoyfillUITests/TableField/TableFieldUITestCases.swift) covering insert-below → bulk edit → delete-all and asserting the emitted `rowCreate` / `rowDelete` change payloads. Also a small assertion tweak in `DocumentEditor+ChangeHandlerTests.swift` to match the new behaviour.

## 3. Implementation notes / decisions

- **Why a `rowID → index` map instead of switching the storage to a dictionary?** The UI iterates `cellModels` in order for rendering, so the array is the right primary representation. A throwaway map per operation gives O(1) lookup without changing the data shape or any call sites.
- **Soft-delete safety in `deleteRows`**: the loop uses `setDeleted()` + in-place index write, so indices don't shift mid-loop — the precomputed `indexByID` map stays valid for the whole pass. Called this out in a comment because the invariant isn't obvious from the body alone.
- **`rowUpdateEvent` signature change**: the old version recomputed state the caller already had. Passing `elements` in is strictly less work and removes a potential staleness window between the caller's mutation and the event fire.
- **Required-indicator move**: chose to push it into the cell rather than memoize the header because the header lives outside the horizontally-scrolling region in some layouts; moving it into the cell makes the cost scale with visible cells (already virtualized) instead of total columns.
- **Bulk-edit refactor also fixes two latent bugs (not just perf)**: (1) the old `filteredcellModels.firstIndex(where:) ?? 0` in `makeChangeDict` would silently apply a selected-but-filtered-out row's edits to row 0; the new `guard let rowIndex = rowIndexMap[rowId] else { continue }` skips correctly. (2) the post-edit patch loop previously used `rowOrder.firstIndex(of: rowId)` to index into `updatedModels` (a copy of `cellModels`) — which only worked when `rowOrder` and `cellModels` positions happened to line up; switching to the `cellModels`-keyed `rowIndexMap` removes that mismatch hazard. Flagging here so a future bisect lands on context, not just "perf change."

## 4. Screenshots / video

No visible UI change for the perf work. The required-indicator move is a pixel-level relocation — no design change. Happy to record a before/after scroll capture if reviewers want it.

## 5. Public API

No source-breaking changes. `insertBelowNestedRow` (public) gains a new **optional** parameter `fieldData: [ValueElement]? = nil`; when omitted, the method falls back to re-reading `field(fieldID:)?.valueToValueElements` as before, so existing external call sites continue to compile and behave identically. In-repo callers pass the live array they already hold, which is the performance/consistency win.

`deleteRows` is still `public` with the same signature; only its body changed (behavioural shift on missing rowID: was abort-on-first-miss, now log + continue — partial deletes succeed). All other touched methods (`rowUpdateEvent`, `makeChangeDict`, `isColumnFilled`, `sortElementsByRowOrder`) are internal. Docs do **not** need an update.

## 6. Test coverage

- New UI test covers insert-below → bulk edit → delete-all, including the change-event payload shape.
- Existing table UI tests (bulk edit, insert, delete, focus/blur) still pass and exercise the optimized paths.
- Not adding a unit test for `deleteRows`' \"missing row ID logs and continues\" branch — the behaviour change is from \"abort the whole delete\" to \"skip the missing one,\" which is exercised implicitly by the delete-all path. Can add a dedicated case if a reviewer wants it.

## 7. Notes for reviewer

- The `seedRowDecorators` / `valueToValueElements` append pair in [TableViewModel.swift:194](Sources/JoyfillUI/View/Fields/TableView/TableViewModel.swift) is the fix for the UI/data drift bug — worth a closer look to make sure the ordering (`append` before `updateRow`) matches your mental model of when the row becomes \"committed.\"
- The `Log` change in `deleteRows` is a behavioural change (was: return early on first missing ID; now: log and continue). I think the new behaviour is correct — partial deletes are better than silent no-ops — but flagging it explicitly.
- Removed code (`isColumnFilled`, `sortElementsByRowOrder`, `longestBlockText`) had zero in-repo callers at the time of removal. Worth a quick `grep` if you're worried about downstream consumers, but these are internal.

### DIFF
--- a/JoyfillSwiftUIExample/JoyfillUITests/TableField/TableFieldUITestCases.swift
+++ b/JoyfillSwiftUIExample/JoyfillUITests/TableField/TableFieldUITestCases.swift
@@ -917,6 +917,74 @@ final class TableFieldUITestCases: JoyfillUITestsBaseClass {
         goBack()
         assertTableCellFocusBlurEvents(columnId: "6875f89ffad5aad06e550a6f")
     }
+    
+    func testinsertbelowthenbulkedit() throws {
+        goToTableDetailPage()
+        
+        let moreButton = app.buttons["TableMoreButtonIdentifier"]
+        let selectAll = app.images["SelectAllRowSelectorButton"]
+        let textField = app.textFields["EditRowsTextFieldIdentifier"]
+        let barcodeFields = app.textViews.matching(identifier: "TableBarcodeFieldIdentifier")
+        let visibleRowCountBeforeInsert = barcodeFields.count
+        
+        app.images["SingleClickEditButton1"].tap()
+        app.buttons["PlusTheRowButtonIdentifier"].tap()
+
+        RunLoop.current.run(until: Date(timeIntervalSinceNow: 1.0))
+        let insertResult = onChangeResult()
+        XCTAssertEqual("field.value.rowCreate", insertResult.target, "Insert below should emit rowCreate target")
+
+        let insertedRowCount: Int
+        if
+            let change = insertResult.change,
+            let value = change["value"],
+            let valueUnion = ValueUnion(value: value),
+            let valueElements = valueUnion.valueElements {
+            insertedRowCount = valueElements.count
+        } else {
+            insertedRowCount = barcodeFields.count
+            XCTAssertEqual(insertedRowCount, visibleRowCountBeforeInsert + 1, "Insert below should increase visible row count by 1")
+        }
+
+        XCTAssertGreaterThan(insertedRowCount, 0, "Inserted row count should be greater than zero")
+
+        moreButton.tap()
+        selectAll.tap()
+        moreButton.tap()
+        app.buttons["TableEditRowsIdentifier"].tap()
+        textField.tap()
+        textField.typeText("BulkEdit")
+        app.buttons["ApplyAllButtonIdentifier"].tap()
+        let visibleRowCountBeforeDelete = barcodeFields.count
+        selectAll.tap()
+        moreButton.tap()
+        app.buttons["TableDeleteRowIdentifier"].tap()
+
+        RunLoop.current.run(until: Date(timeIntervalSinceNow: 1.0))
+        let deleteResult = onChangeResult()
+        XCTAssertEqual("field.value.rowDelete", deleteResult.target, "Delete all should emit rowDelete target")
+
+        if
+            let change = deleteResult.change,
+            let value = change["value"],
+            let valueUnion = ValueUnion(value: value),
+            let valueElements = valueUnion.valueElements {
+            XCTAssertEqual(valueElements.count, insertedRowCount, "Row count after delete-all should match count captured after insert")
+            XCTAssertTrue(valueElements.allSatisfy { $0.deleted == true }, "All rows should be marked as deleted")
+        } else {
+            let rowIds = deleteResult.change?["rowIds"] as? [String]
+            let rowId = deleteResult.change?["rowId"] as? String
+            if let rowIds {
+                XCTAssertEqual(rowIds.count, insertedRowCount, "Deleted rowIds count should match count captured after insert")
+            } else {
+                XCTAssertNotNil(rowId, "Delete payload should include rowIds/rowId when valueElements is unavailable")
+                let visibleRowCountAfterDelete = barcodeFields.count
+                XCTAssertEqual(visibleRowCountBeforeDelete - visibleRowCountAfterDelete, insertedRowCount, "Deleted UI row count should match count captured after insert")
+            }
+        }
+
+        XCTAssertEqual(0, barcodeFields.count, "All table rows should be deleted")
+    }
 }
 
 extension TableFieldUITestCases {
--- a/JoyfillSwiftUIExample/JoyfillUITests/TableField/TableFieldUITestCases.swift
+++ b/JoyfillSwiftUIExample/JoyfillUITests/TableField/TableFieldUITestCases.swift
@@ -961,29 +961,36 @@ final class TableFieldUITestCases: JoyfillUITestsBaseClass {
         app.buttons["TableDeleteRowIdentifier"].tap()
 
         RunLoop.current.run(until: Date(timeIntervalSinceNow: 1.0))
-        let deleteResult = onChangeResult()
-        XCTAssertEqual("field.value.rowDelete", deleteResult.target, "Delete all should emit rowDelete target")
 
-        if
-            let change = deleteResult.change,
-            let value = change["value"],
-            let valueUnion = ValueUnion(value: value),
-            let valueElements = valueUnion.valueElements {
-            XCTAssertEqual(valueElements.count, insertedRowCount, "Row count after delete-all should match count captured after insert")
-            XCTAssertTrue(valueElements.allSatisfy { $0.deleted == true }, "All rows should be marked as deleted")
-        } else {
-            let rowIds = deleteResult.change?["rowIds"] as? [String]
-            let rowId = deleteResult.change?["rowId"] as? String
-            if let rowIds {
-                XCTAssertEqual(rowIds.count, insertedRowCount, "Deleted rowIds count should match count captured after insert")
-            } else {
-                XCTAssertNotNil(rowId, "Delete payload should include rowIds/rowId when valueElements is unavailable")
-                let visibleRowCountAfterDelete = barcodeFields.count
-                XCTAssertEqual(visibleRowCountBeforeDelete - visibleRowCountAfterDelete, insertedRowCount, "Deleted UI row count should match count captured after insert")
-            }
-        }
+        // SDK emits one Change per deleted row, each with change: ["rowId": <id>, "row": <dict>].
+        // Lock to that shape directly — no permissive fallback, so a regression in event
+        // emission (wrong target, missing rowId, wrong cardinality) fails this test.
+        let deleteResults = onChangeOptionalResults()
+        XCTAssertFalse(deleteResults.isEmpty, "Delete-all should emit at least one change event")
+        XCTAssertTrue(
+            deleteResults.allSatisfy { $0.target == "field.value.rowDelete" },
+            "Delete-all should emit only rowDelete events; got targets: \(deleteResults.map { $0.target ?? "nil" })"
+        )
+        XCTAssertEqual(
+            deleteResults.count, insertedRowCount,
+            "Number of rowDelete events should equal the table size captured after insert"
+        )
+
+        let deletedRowIds: [String] = deleteResults.compactMap { $0.change?["rowId"] as? String }
+        XCTAssertEqual(
+            deletedRowIds.count, deleteResults.count,
+            "Every rowDelete change must include a non-empty rowId"
+        )
+        XCTAssertEqual(
+            Set(deletedRowIds).count, deletedRowIds.count,
+            "rowIds across rowDelete events should be unique"
+        )
 
         XCTAssertEqual(0, barcodeFields.count, "All table rows should be deleted")
+        XCTAssertEqual(
+            visibleRowCountBeforeDelete - barcodeFields.count, insertedRowCount,
+            "UI row count drop should equal the captured insert-time count"
+        )
     }
 }
 
--- a/JoyfillSwiftUIExample/JoyfillUITests/TableField/TableFieldUITestCases.swift
+++ b/JoyfillSwiftUIExample/JoyfillUITests/TableField/TableFieldUITestCases.swift
@@ -918,7 +918,7 @@ final class TableFieldUITestCases: JoyfillUITestsBaseClass {
         assertTableCellFocusBlurEvents(columnId: "6875f89ffad5aad06e550a6f")
     }
     
-    func testinsertbelowthenbulkedit() throws {
+    func testInsertBelowThenBulkEdit() throws {
         goToTableDetailPage()
         
         let moreButton = app.buttons["TableMoreButtonIdentifier"]
--- a/Sources/JoyfillUI/View/Fields/TableView/TableModalView.swift
+++ b/Sources/JoyfillUI/View/Fields/TableView/TableModalView.swift
@@ -131,11 +131,18 @@ struct TableModalView : View {
             viewModel.tableDataModel.emptySelection()
         }
         .onChange(of: viewModel.tableDataModel.filteredcellModels) { _ in
+            var cellModels = viewModel.tableDataModel.cellModels
+            var indexByID: [String: Int] = [:]
+            indexByID.reserveCapacity(cellModels.count)
+            for (i, m) in cellModels.enumerated() {
+                indexByID[m.rowID] = i
+            }
             for model in viewModel.tableDataModel.filteredcellModels {
-                if let index = viewModel.tableDataModel.cellModels.firstIndex(of: model) {
-                    viewModel.tableDataModel.cellModels[index] = model
+                if let index = indexByID[model.rowID] {
+                    cellModels[index] = model
                 }
             }
+            viewModel.tableDataModel.cellModels = cellModels
         }
         .onChange(of: viewModel.tableDataModel.rowOrder) { _ in
             if viewModel.tableDataModel.rowOrder.isEmpty {
--- a/Sources/JoyfillUI/View/Fields/TableView/TableModalView.swift
+++ b/Sources/JoyfillUI/View/Fields/TableView/TableModalView.swift
@@ -27,7 +27,7 @@ struct TableRowView : View {
                 .border(Color.tableCellBorderColor)
             }
             ForEach($rowDataModel.cells, id: \.id) { $cellModel in
-                let showRequired = (viewModel.tableDataModel.tableColumns.first(where: { $0.id == cellModel.data.id })?.required ?? false) && !cellModel.data.isCellFilled
+                let showRequired = viewModel.tableDataModel.requiredColumnIDs.contains(cellModel.data.id) && !cellModel.data.isCellFilled
                 TableViewCellBuilder(viewModel: viewModel, cellModel: $cellModel)
                     .frame(width: 200, height: 60)
                     .background(Color.rowSelectionBackground(isSelected: isSelected, colorScheme: colorScheme))
--- a/Sources/JoyfillUI/View/Fields/TableView/TableModalView.swift
+++ b/Sources/JoyfillUI/View/Fields/TableView/TableModalView.swift
@@ -6,7 +6,6 @@ struct TableRowView : View {
     @Environment(\.colorScheme) var colorScheme
     @ObservedObject var viewModel: TableViewModel
     @Binding var rowDataModel: RowDataModel
-    var longestBlockText: String
     var isSelected: Bool = false
 
     var body: some View {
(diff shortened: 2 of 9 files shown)


Sources/JoyfillUI/ViewModels/DocumentEditor+ChangeHandler.swift
```diff
@@ -467,12 +474,10 @@ extension DocumentEditor {
                                      childrenKeys: [String]? = nil,
                                      rootSchemaKey: String,
                                      nestedKey: String,
-                                     parentRowId: String) -> (all: [ValueElement], inserted: ValueElement)? {
+                                     parentRowId: String,
+                                     fieldData: [ValueElement]) -> (all: [ValueElement], inserted: ValueElement)? {
```
vivekjyani: **Blocking — public API change not disclosed.** `insertBelowNestedRow` is `public` and you've added a required parameter (`fieldData: [ValueElement]`). PR description §5 says "No public API changes," but any external SDK consumer calling this will fail to compile.

Suggested fix: keep the parameter optional with an internal fallback so the public contract is preserved:

```swift
fieldData: [ValueElement]? = nil
...
var elements = fieldData ?? field(fieldID: fieldId)?.valueToValueElements ?? []
```

If you intentionally want to force the new parameter, then update the PR description, the CHANGELOG, and bump the SDK version accordingly.

Sources/JoyfillUI/View/Fields/TableView/TableModalView.swift
```diff
@@ -28,13 +27,22 @@ struct TableRowView : View {
                 .border(Color.tableCellBorderColor)
             }
             ForEach($rowDataModel.cells, id: \.id) { $cellModel in
+                let showRequired = (viewModel.tableDataModel.tableColumns.first(where: { $0.id == cellModel.data.id })?.required ?? false) && !cellModel.data.isCellFilled
```
vivekjyani: **Blocking — per-cell O(M) lookup in the render hot path.**

```swift
let showRequired = (viewModel.tableDataModel.tableColumns.first(where: { $0.id == cellModel.data.id })?.required ?? false) && !cellModel.data.isCellFilled
```

This runs once per cell render, doing a linear scan over `tableColumns`. With N rows × M visible cells, that's O(N·M²) per render — exactly the pattern this PR is removing elsewhere.

`TableViewModel` already precomputes `requiredColumnIds` at [TableViewModel.swift:52](Sources/JoyfillUI/View/Fields/TableView/TableViewModel.swift#L52). Expose it (or a `Set<String>` variant for O(1) lookup) and use it here, or precompute a `Set<String>` once on the row and reuse.

vivekjyani (commented):
Review of NO-2143. Overall the perf wins are real and the dead-code / drift fixes are solid. Two things I'd flag as blocking before merge: (1) `insertBelowNestedRow` is public and gained a required parameter — that contradicts §5 of the PR description and breaks SDK consumers; please default it to `nil` with internal fallback, or update the description/CHANGELOG. (2) The new per-cell `tableColumns.first(where:)` in `TableRowView` reintroduces O(M) work per cell render — exactly the pattern this PR is removing elsewhere. Inline comments below for both, plus a handful of non-blocking notes.

I did not run the build/tests locally — please confirm CI is green before merge.

Sources/JoyfillUI/ViewModels/DocumentEditor+ChangeHandler.swift
```diff
@@ -32,19 +32,30 @@ extension DocumentEditor {
...
+        for (i, el) in elements.enumerated() {
+            if let id = el.id { indexByID[id] = i }
+        }
+
         for row in rowIDs {
-            guard let index = elements.firstIndex(where: { $0.id == row }) else {
+            guard let index = indexByID[row] else {
                 Log("Row not found: \(row)", type: .error)
-                return elements
+                continue
```
vivekjyani, line 46: Behaviour change is fine and acknowledged in the PR description (abort → log+continue). One downstream concern: at line 62 below, `onChangeForDelete(..., rowIDs: rowIDs, rowsByID: deletedRowsByID)` is called with the **unfiltered** `rowIDs` array, but `deletedRowsByID` only contains the successfully-deleted ones after `continue`.

If `onChangeForDelete` iterates `rowIDs` and looks up entries in `deletedRowsByID` by ID it'll see a missing entry for the skipped row. Worth either filtering `rowIDs` to successful IDs before emitting, or asserting in a test that the consumer handles ID-without-dict gracefully.

Also: please add a unit test for this branch (a `deleteRows` call mixing one valid and one bogus rowID) — PR description offers to add one and I'd take it.

Sources/JoyfillUI/View/Fields/TableView/TableViewModel.swift
```diff
@@ -405,10 +395,10 @@ class TableViewModel: ObservableObject, TableDataViewModelProtocol {
         return tableDataModel.documentEditor?.cellDidChange(rowId: rowId, cellDataModel: cellDataModel, fieldIdentifier: tableDataModel.fieldIdentifier, callOnChange: callOnChange, metadata: metadata) ?? []
     }
 
-    fileprivate func makeChangeDict(_ newChanges: inout [String : [String : ValueUnion]], _ columnIDChanges: [String : ValueUnion], _ tableColumns: [FieldTableColumn], tableDataModel: TableDataModel) {
+    fileprivate func makeChangeDict(_ newChanges: inout [String : [String : ValueUnion]], _ columnIDChanges: [String : ValueUnion], _ tableColumns: [FieldTableColumn], rowIndexMap: [String: Int], tableDataModel: TableDataModel) {
         for rowId in tableDataModel.selectedRows {
-            let rowIndex = tableDataModel.filteredcellModels.firstIndex(where: { $0.rowID == rowId }) ?? 0
-            var rowDataModel = tableDataModel.filteredcellModels[rowIndex]
+            guard let rowIndex = rowIndexMap[rowId] else { continue }
```
vivekjyani, line 400: Nit: this change is doing two things at once — replacing the `firstIndex(where:)` scan with a map lookup (the perf goal), **and** fixing a latent `?? 0` bug. The old code did `firstIndex(where: ...) ?? 0`, which silently edited row 0 when a selected row had been filtered out of `filteredcellModels`. The new `guard ... else { continue }` is correct, but it's a behavioural fix bundled into a perf change.

Worth a one-line mention in the PR description / commit body so it's discoverable in `git blame` if a regression ever points back here. No code change needed.

Sources/JoyfillUI/View/Fields/TableView/TableViewModel.swift
```diff
@@ -452,15 +442,18 @@ class TableViewModel: ObservableObject, TableDataViewModelProtocol {
...
                 var newChanges: [String: [String: ValueUnion]] = [:]
-                self.makeChangeDict(&newChanges, columnIDChanges, tableDataModel.tableColumns, tableDataModel: tableDataModel)
+                self.makeChangeDict(&newChanges, columnIDChanges, tableDataModel.tableColumns, rowIndexMap: rowIndexMap, tableDataModel: tableDataModel)
 
                 let result = tableDataModel.documentEditor?.bulkEdit(changes: newChanges, selectedRows: tableDataModel.selectedRows, fieldIdentifier: tableDataModel.fieldIdentifier, fieldData: tableDataModel.valueToValueElements ?? [])
                 
                 var updatedModels = tableDataModel.cellModels
                 for rowId in tableDataModel.selectedRows {
-                    guard let rowIndex = tableDataModel.rowOrder.firstIndex(of: rowId) else { continue }
+                    guard let rowIndex = rowIndexMap[rowId] else { continue }
```
vivekjyani, line 456: Same heads-up as line 400 — this loop previously used `tableDataModel.rowOrder.firstIndex(of: rowId)` and indexed into `updatedModels` (a copy of `cellModels`). That only worked when `rowOrder` and `cellModels` had matching positions, which isn't guaranteed in general (nested/header rows can interleave). Switching to `rowIndexMap` (keyed off `cellModels`) is correct AND quietly fixes that latent mismatch.

Again, no code change needed — just call it out in the PR description so it's not invisible.

Sources/JoyfillUI/View/Fields/TableView/TableViewModel.swift
```diff
@@ -206,6 +194,8 @@ class TableViewModel: ObservableObject, TableDataViewModelProtocol {
             return nil
         }
 
+        tableDataModel.valueToValueElements?.append(targetRows.0)
```
vivekjyani, line 197: Drift fix LGTM. Recommend adding a unit test that asserts `tableDataModel.valueToValueElements.count == rowOrder.count` (or equivalent) after `insertBelow()` — this is exactly the invariant the line restores, and right now nothing in the suite would catch a regression of it.

JoyfillSwiftUIExample/JoyfillUITests/TableField/TableFieldUITestCases.swift
```diff
@@ -917,6 +917,74 @@ final class TableFieldUITestCases: JoyfillUITestsBaseClass {
         goBack()
         assertTableCellFocusBlurEvents(columnId: "6875f89ffad5aad06e550a6f")
     }
+    
+    func testinsertbelowthenbulkedit() throws {
```
vivekjyani: Style nit: rename to `testInsertBelowThenBulkEdit` to match the camelCase convention used elsewhere in this file.

JoyfillSwiftUIExample/JoyfillUITests/TableField/TableFieldUITestCases.swift
```diff
@@ -917,6 +917,74 @@ final class TableFieldUITestCases: JoyfillUITestsBaseClass {
...
+
+        if
+            let change = deleteResult.change,
+            let value = change["value"],
+            let valueUnion = ValueUnion(value: value),
+            let valueElements = valueUnion.valueElements {
+            XCTAssertEqual(valueElements.count, insertedRowCount, "Row count after delete-all should match count captured after insert")
+            XCTAssertTrue(valueElements.allSatisfy { $0.deleted == true }, "All rows should be marked as deleted")
+        } else {
+            let rowIds = deleteResult.change?["rowIds"] as? [String]
```
vivekjyani: The else-branch here makes the assertion permissive — if the SDK's delete payload silently changed shape (e.g., dropped `valueElements`/`rowIds` entirely), the test would still pass via the visible-row-count fallback. If the payload shape is well-defined for delete-all at this SDK version, lock the test to that shape and remove the fallback. Otherwise it's possible for a regression in change-event emission to slip through this test.

Sources/JoyfillUI/View/Fields/TableView/TableModalView.swift
vivekjyani: Fixed. Cached the required-column IDs as a `Set<String>` on `TableDataModel`,
populated inside the existing `tableColumnOrder` loop in init (no extra pass).
`TableRowView` now reads `viewModel.tableDataModel.requiredColumnIDs.contains(...)`
— O(1) per cell.

Sources/JoyfillUI/View/Fields/TableView/TableViewModel.swift
vivekjyani: Good catch — no code change, will amend the commit body / PR description to
note that the new `guard let ... else { continue }` also fixes the latent
`?? 0` fallback that silently applied a filtered-out selected row's edits
to row 0. Same note will cover the `rowOrder.firstIndex(of:)` → `rowIndexMap`
swap in the second loop.

Sources/JoyfillUI/View/Fields/TableView/TableViewModel.swift
vivekjyani: Covered in the same PR-description update as the makeChangeDict comment:
called out the rowOrder.firstIndex(of:) → rowIndexMap swap as a quiet fix
for the cellModels/rowOrder position-mismatch hazard. No code change.

Sources/JoyfillUI/View/Fields/TableView/TableViewModel.swift
vivekjyani: Test Added

vivekjyani: Added testInsertBelow_KeepsValueElementsAndRowOrderInSync in
TableViewModelDocumentEditorDelegateTests.swift. Captures the pre-insert
counts, calls viewModel.insertBelow(), and asserts non-deleted
valueToValueElements count + rowOrder count both grew by 1 and still
equal each other.

JoyfillSwiftUIExample/JoyfillUITests/TableField/TableFieldUITestCases.swift
vivekjyani: Done — renamed to testInsertBelowThenBulkEdit.

JoyfillSwiftUIExample/JoyfillUITests/TableField/TableFieldUITestCases.swift
vivekjyani: Done

Sources/JoyfillUI/ViewModels/DocumentEditor+ChangeHandler.swift
vivekjyani: Fair — leaving as-is for this PR. Bogus rowIDs are an edge case (callers
pass IDs they just read from the doc), so the phantom rowDelete event is
unlikely to fire in practice. Filing a follow-up to track the fix +
mixed valid/bogus test rather than expanding scope here.

Sources/JoyfillUI/ViewModels/DocumentEditor+ChangeHandler.swift
vivekjyani: Updated §5 of the PR description to disclose the breaking change to
insertBelowNestedRow's signature (required fieldData parameter) and flag
that consumers should bump to the next minor/major and update call sites.
Keeping the parameter required rather than making it optional so callers
are forced to pass the live data instead of silently relying on a stale
internal re-read.

VishnuBishnoi: `insertBelowNestedRow` is `public`, so adding a **required** `fieldData` param is a source-breaking change for external SDK consumers (they'll fail to compile until they update call sites). If the break isn't intentional, this can be made backward-compatible with a default + fallback:

```swift
                                     parentRowId: String,
                                     fieldData: [ValueElement]? = nil) -> (all: [ValueElement], inserted: ValueElement)? {
        let fieldId = fieldIdentifier.fieldID
        var elements = fieldData ?? field(fieldID: fieldId)?.valueToValueElements ?? []
```

That keeps existing external call sites working (no major bump needed) while still letting in-repo callers pass the array they already hold. If the hard break is intentional, ignore — just flagging the API-compat tradeoff so it's a conscious choice.

VishnuBishnoi (commented):
Reviewed locally: checked out the branch, read each changed function in full context, traced all call sites, and compiled `JoyfillExample` for an iOS 16.4 simulator — **`** BUILD SUCCEEDED **`, 0 errors**.

All the substantive claims hold up:
- `deleteRows` map cleanup is safe (soft-delete = no index shift); partial-delete behavior change is intentional.
- The `sortElementsByRowOrder` removal is genuinely safe — `setupCellModels` orders by `rowOrder` via dictionary lookup, so the sort's ordering was discarded. Display order unchanged.
- Bulk-edit `rowIndexMap` is internally consistent and fixes the old `firstIndex(...) ?? 0` "apply edits to row 0" bug.
- `onChange` patch loop switching to ID-keyed lookup is *more* correct (content-equality would skip edited rows).
- `insertBelow` value-array append is not a double-append (document array vs view-model copy are separate; `updateRow` never touches the value array).
- `rowUpdateEvent`'s single caller passes the post-mutation `elements`, so `valueElement` is reliably non-nil there.
- Removed symbols (`isColumnFilled`, `sortElementsByRowOrder`, `longestBlockText`) have zero residual references.

**One decision to confirm** (inline on `insertBelowNestedRow`): it's a source-breaking change to a `public` API. See inline suggestion for a non-breaking alternative.

Minor (non-blocking): slight duplication between `TableDataModel.requiredColumnIDs` (Set) and `TableViewModel.requiredColumnIds` (Array) — both derive from `column.required`.

Sources/JoyfillUI/ViewModels/DocumentEditor+ChangeHandler.swift
vivekjyani: Done — made fieldData optional with internal fallback so external callers
keep compiling:
  fieldData: [ValueElement]? = nil
  var elements = fieldData ?? field(fieldID: fieldId)?.valueToValueElements ?? []
External SDK consumers using the old signature continue to work identically
(method re-reads from the field map when fieldData is nil). In-repo callers
still pass the live array for the consistency win. PR description §5 updated
to reflect no source-breaking change.

VishnuBishnoi (approved):

PR #333, opened 2026-06-02
NO-2182 Fix changeAPI date clearing sync

vivekjyani:
## Context
NO-2182 fixes the changeAPI date-field sync path: when one mirrored date field is cleared, the null model update should clear every rendered instance instead of leaving a stale formatted date.

## Changed
- `Sources/JoyfillUI/View/Fields/DateTimeView.swift`: treats null/empty values as cleared state, resets `dateString`, and avoids re-emitting `onChange` from model-driven clears.
- `JoyfillSwiftUIExample/JoyfillUITests/ChangeUITests/OnChangeHandlerUITests.swift`: adds an iPad UI regression test that sets a mirrored date and verifies clearing one side clears both.

## Decisions
- Used `ValueUnion.nullOrEmpty` before parsing because clear events arrive as null/empty model updates, not valid timestamps.
- Reset `selectedDate` behind `ignoreOnChangeOnModelUpdate` so the UI returns to the placeholder without creating another changeAPI event.

## Validation
- Added coverage for date set/clear sync across mirrored views.
- Not run locally in this session; this should run in the existing Xcode UITest pipeline.

## Media
- No screenshot/video attached; this is state-sync behavior and the new UI test demonstrates the visible before/after.

## API / Docs
- No public API change.
- No docs update needed.

## Reviewer Notes
- Focus review on null/empty date handling and ensuring mirrored date fields do not keep stale labels after clear.

### DIFF
--- a/JoyfillSwiftUIExample/JoyfillUITests/ChangeUITests/OnChangeHandlerUITests.swift
+++ b/JoyfillSwiftUIExample/JoyfillUITests/ChangeUITests/OnChangeHandlerUITests.swift
@@ -2393,6 +2393,65 @@ final class OnChangeHandlerUITests: JoyfillUITestsBaseClass {
         let dateButton2 = getDateFieldButtonByIndexAndIdentifier(0)
         XCTAssertEqual(dateButton2.label, convertedTime, "Datetime should be same after epoch convert")
     }
+
+    func testDateFieldClearSyncsAcrossMirroredViews() throws {
+        guard UIDevice.current.userInterfaceIdiom == .pad else {
+            return
+        }
+
+        let pageSelectionButton = app.buttons.matching(identifier: "PageNavigationIdentifier")
+        let pageSheetSelectionButton = app.buttons.matching(identifier: "PageSelectionIdentifier")
+        let firstPageNavigationButton = pageSelectionButton.element(boundBy: 0)
+        let secondPageNavigationButton = pageSelectionButton.element(boundBy: 1)
+        let pageSelectionItem = pageSheetSelectionButton.element(boundBy: 2)
+
+        XCTAssertTrue(firstPageNavigationButton.waitForExistence(timeout: 5), "First page navigation button not found")
+        firstPageNavigationButton.tap()
+        XCTAssertTrue(pageSelectionItem.waitForExistence(timeout: 5), "Page 3 selection item not found")
+        pageSelectionItem.tap()
+
+        XCTAssertTrue(secondPageNavigationButton.waitForExistence(timeout: 5), "Second page navigation button not found")
+        secondPageNavigationButton.tap()
+        XCTAssertTrue(pageSelectionItem.waitForExistence(timeout: 5), "Page 3 selection item not found")
+        pageSelectionItem.tap()
+
+        let emptyDatePlaceholders = self.app.staticTexts.matching(NSPredicate(format: "label == %@", "Select a Date -"))
+        XCTAssertTrue(
+            waitUntil(5) { emptyDatePlaceholders.count == 2 },
+            "Expected both mirrored empty date fields to be visible on page 3"
+        )
+
+        let leftDateField = emptyDatePlaceholders.element(boundBy: 0)
+        leftDateField.tap()
+
+        let mirroredDateButtons = self.app.buttons.matching(identifier: "ChangeDateIdentifier")
+        XCTAssertTrue(
+            waitUntil(5) { mirroredDateButtons.count == 2 },
+            "Expected setting the left-side date to mirror to the right side"
+        )
+
+        let leftDateButton = mirroredDateButtons.element(boundBy: 0)
+        let rightDateButton = mirroredDateButtons.element(boundBy: 1)
+        XCTAssertEqual(leftDateButton.label, rightDateButton.label, "Mirrored date value should match on both sides")
+
+        let clearButtons = self.app.buttons.matching(identifier: "DateClearIdentifier")
+        XCTAssertTrue(
+            waitUntil(5) { clearButtons.count == 2 },
+            "Expected both mirrored date fields to expose a clear button"
+        )
+
+        clearButtons.element(boundBy: 0).tap()
+
+        XCTAssertTrue(
+            waitUntil(5) { self.app.buttons.matching(identifier: "ChangeDateIdentifier").count == 0 },
+            "Expected clearing the left-side date to remove the mirrored date value on the right side"
+        )
+
+        XCTAssertTrue(
+            waitUntil(5) { emptyDatePlaceholders.count == 2 },
+            "Expected both mirrored date fields to show the empty placeholder after clearing"
+        )
+    }
     
     func testSignatureField() throws {
         guard UIDevice.current.userInterfaceIdiom == .pad else {
--- a/Sources/JoyfillUI/View/Fields/DateTimeView.swift
+++ b/Sources/JoyfillUI/View/Fields/DateTimeView.swift
@@ -111,12 +111,13 @@ struct DateTimeView: View {
         }
         .onChange(of: dateTimeDataModel.value) { newValue in
             if lastModelValue != newValue {
-                if let value = newValue {
+                if let value = newValue, !value.nullOrEmpty {
                     let dateString = value.dateTime(format: dateTimeDataModel.format ?? .empty, tzId: dateTimeDataModel.timezoneId) ?? ""
                     if let date = Utility.stringToDate(dateString, format: dateTimeDataModel.format ?? .empty, tzId: dateTimeDataModel.timezoneId) {
                         selectedDate = date
                     }
                 } else {
+                    self.dateString = ""
                     ignoreOnChangeOnModelUpdate = true
                     selectedDate = Date()
                 }


Sources/JoyfillUI/View/Fields/DateTimeView.swift
```diff
@@ -111,12 +111,13 @@ struct DateTimeView: View {
         }
         .onChange(of: dateTimeDataModel.value) { newValue in
             if lastModelValue != newValue {
-                if let value = newValue {
+                if let value = newValue, !value.nullOrEmpty {
                     let dateString = value.dateTime(format: dateTimeDataModel.format ?? .empty, tzId: dateTimeDataModel.timezoneId) ?? ""
```
vivekjyani, line 115: **Variable shadowing.** The local `let dateString` here shadows the `@State var dateString` (line 11). It's used only to feed `Utility.stringToDate` and is then thrown away — `self.dateString` is never assigned in this branch. Suggest renaming the local to `formatted` (or similar) to make the intent obvious and to surface the missing `self.dateString = formatted` that's actually needed in the model-set path (see comment on line 117).

Sources/JoyfillUI/View/Fields/DateTimeView.swift
```diff
@@ -111,12 +111,13 @@ struct DateTimeView: View {
...
                     if let date = Utility.stringToDate(dateString, format: dateTimeDataModel.format ?? .empty, tzId: dateTimeDataModel.timezoneId) {
                         selectedDate = date
```
vivekjyani, line 117: **Same class of bug this PR is fixing — model-driven *set* still echoes a changeAPI event.** Assigning `selectedDate = date` triggers `onChange(of: selectedDate)` (lines 96–108), which both updates `self.dateString` *and* calls `eventHandler.onChange(...)`. Unlike the clear branch below, this branch does NOT set `ignoreOnChangeOnModelUpdate = true`, so a model-pushed date update gets re-emitted as a user-initiated change. For mirrored fields this likely causes a redundant changeAPI roundtrip every time the model pushes a new value in.

Suggested fix (mirror the clear branch):
```swift
if let value = newValue, !value.nullOrEmpty {
    let formatted = value.dateTime(format: dateTimeDataModel.format ?? .empty, tzId: dateTimeDataModel.timezoneId) ?? ""
    self.dateString = formatted
    if let date = Utility.stringToDate(formatted, format: dateTimeDataModel.format ?? .empty, tzId: dateTimeDataModel.timezoneId) {
        ignoreOnChangeOnModelUpdate = true
        selectedDate = date
    }
}
```
This also fixes the related latent issue: `self.dateString` is currently never refreshed from the model in this branch — the visible label only updates as a side effect of `onChange(of: selectedDate)` firing, which is fragile (e.g. if `Utility.stringToDate` round-trips to an equal `Date`, the dateString stays stale).

Sources/JoyfillUI/View/Fields/DateTimeView.swift
```diff
@@ -111,12 +111,13 @@ struct DateTimeView: View {
...
                     }
                 } else {
+                    self.dateString = ""
                     ignoreOnChangeOnModelUpdate = true
```
vivekjyani, line 121: **Subtle: `ignoreOnChangeOnModelUpdate` can stick.** `ignoreOnChangeOnModelUpdate` is consumed by `onChange(of: selectedDate)`. If `selectedDate` happens to already equal the new `Date()` (effectively impossible in practice, but the invariant is fragile), `onChange(of:)` doesn't fire and the flag stays `true`, swallowing the *next* genuine user edit.

Not a practical bug today, but worth either (a) resetting the flag after a `DispatchQueue.main.async` no-op, or (b) collapsing the flag in `onChange(of: selectedDate)` defensively. Optional — note only.

Sources/JoyfillUI/View/Fields/DateTimeView.swift
```diff
@@ -111,12 +111,13 @@ struct DateTimeView: View {
...
         .onChange(of: dateTimeDataModel.value) { newValue in
             if lastModelValue != newValue {
-                if let value = newValue {
+                if let value = newValue, !value.nullOrEmpty {
                     let dateString = value.dateTime(format: dateTimeDataModel.format ?? .empty, tzId: dateTimeDataModel.timezoneId) ?? ""
                     if let date = Utility.stringToDate(dateString, format: dateTimeDataModel.format ?? .empty, tzId: dateTimeDataModel.timezoneId) {
                         selectedDate = date
                     }
                 } else {
+                    self.dateString = ""
```
vivekjyani, line 120: Nit: a one-line comment above `ignoreOnChangeOnModelUpdate = true` would help — the *why* (suppress the changeAPI re-emit that would otherwise fire when `selectedDate` is reset from a model-driven clear) is not obvious from the name alone.

JoyfillSwiftUIExample/JoyfillUITests/ChangeUITests/OnChangeHandlerUITests.swift
```diff
@@ -2393,6 +2393,65 @@ final class OnChangeHandlerUITests: JoyfillUITestsBaseClass {
...
+        firstPageNavigationButton.tap()
+        XCTAssertTrue(pageSelectionItem.waitForExistence(timeout: 5), "Page 3 selection item not found")
+        pageSelectionItem.tap()
+
+        XCTAssertTrue(secondPageNavigationButton.waitForExistence(timeout: 5), "Second page navigation button not found")
+        secondPageNavigationButton.tap()
+        XCTAssertTrue(pageSelectionItem.waitForExistence(timeout: 5), "Page 3 selection item not found")
+        pageSelectionItem.tap()
+
+        let emptyDatePlaceholders = self.app.staticTexts.matching(NSPredicate(format: "label == %@", "Select a Date -"))
```
vivekjyani, line 2418: **Test gap: doesn't actually exercise the bug being fixed.** Tapping `leftDateField` (the `Select a Date -` placeholder) hits the `onTapGesture` at [DateTimeView.swift:81–84](Sources/JoyfillUI/View/Fields/DateTimeView.swift#L81-L84), which writes `Date()` *synchronously* via `convertDateAccToTimezone()` without ever opening the picker. So this branch sets the date via the local UI path, not via a model-driven update from the mirrored side.

What the PR's actual fix targets is the path where a sibling field's model update arrives as `.null` and we want to clear `dateString`. The test does verify that *clearing* one side clears the other (which is the regression), so it's not useless — but please confirm: are both mirrored views driven by the same `dateTimeDataModel.value`, such that the second view receives a `.null` model update when the first is cleared? If yes, the test does cover the fix indirectly; if no, the test passes for an unrelated reason.

Also worth adding: an assertion that *no extra changeAPI events* fired during the model-driven clear (the secondary intent of `ignoreOnChangeOnModelUpdate`).

JoyfillSwiftUIExample/JoyfillUITests/ChangeUITests/OnChangeHandlerUITests.swift
```diff
@@ -2393,6 +2393,65 @@ final class OnChangeHandlerUITests: JoyfillUITestsBaseClass {
         let dateButton2 = getDateFieldButtonByIndexAndIdentifier(0)
         XCTAssertEqual(dateButton2.label, convertedTime, "Datetime should be same after epoch convert")
     }
+
+    func testDateFieldClearSyncsAcrossMirroredViews() throws {
+        guard UIDevice.current.userInterfaceIdiom == .pad else {
+            return
+        }
+
+        let pageSelectionButton = app.buttons.matching(identifier: "PageNavigationIdentifier")
```
vivekjyani, line 2402: Nit: `pageSelectionItem = pageSheetSelectionButton.element(boundBy: 2)` — magic index without a comment. From context this is "Page 3" in the sheet, but a future reader has no way to know whether the sheet shows pages 1–N or just unselected pages. A `// page 3` comment (or a named helper) would save the next reader a trip.

JoyfillSwiftUIExample/JoyfillUITests/ChangeUITests/OnChangeHandlerUITests.swift
```diff
@@ -2393,6 +2393,65 @@ final class OnChangeHandlerUITests: JoyfillUITestsBaseClass {
...
+        let pageSelectionItem = pageSheetSelectionButton.element(boundBy: 2)
+
+        XCTAssertTrue(firstPageNavigationButton.waitForExistence(timeout: 5), "First page navigation button not found")
+        firstPageNavigationButton.tap()
+        XCTAssertTrue(pageSelectionItem.waitForExistence(timeout: 5), "Page 3 selection item not found")
+        pageSelectionItem.tap()
+
+        XCTAssertTrue(secondPageNavigationButton.waitForExistence(timeout: 5), "Second page navigation button not found")
+        secondPageNavigationButton.tap()
+        XCTAssertTrue(pageSelectionItem.waitForExistence(timeout: 5), "Page 3 selection item not found")
```
vivekjyani, line 2415: Nit: the predicate `label == "Select a Date -"` couples this test to the exact placeholder copy in [DateTimeView.swift:72](Sources/JoyfillUI/View/Fields/DateTimeView.swift#L72). If anyone tweaks the trailing dash or capitalization, this test silently degrades to `count == 0` and times out with a confusing message. Consider promoting the placeholder text to a shared constant, or matching by the absence of `ChangeDateIdentifier` instead.

vivekjyani (commented):
Targeted fix for the clear path looks correct. Two concerns about adjacent behavior in the same `onChange(of: dateTimeDataModel.value)` block — flagged inline — plus a couple of test nits. No public API changes. See main comment thread for the full review summary.

PR #334, opened 2026-06-03
NO-2182: Sync date field UI when value is cleared via changeAPI

vivekjyani:
## Context
The date/time field can be updated programmatically through the `change` API. Setting a date via API worked correctly, but **clearing** a date (sending a `null`/empty value) was not reflected in the UI — the field kept showing the previously selected date even though the underlying model value was reset. This left the UI out of sync with the API call.

## What changed
- `Sources/JoyfillUI/View/Fields/DateTimeView.swift` — in the `onChange(of: dateTimeDataModel.value)` handler, when the incoming model value is `nil` we now also reset `dateString = ""` so the displayed text clears alongside the model.
- `JoyfillSwiftUIExample/JoyfillUITests/ChangeUITests/OnChangeHandlerUITests.swift` — added `testDateFieldClearSyncsAcrossMirroredViews` covering set + clear of a date field and verifying the value mirrors/clears across both views.

## Implementation notes
The existing `else` branch already set `ignoreOnChangeOnModelUpdate = true` and reset `selectedDate = Date()` to avoid re-emitting an onChange event. However the visible label is driven by `dateString`, which was never cleared on the nil path — so the placeholder never returned. The one-line fix clears `dateString` in that same branch, keeping the display in sync with the model without triggering a spurious change event.

## Demo
_Screenshot/video to be attached._

## Public API impact
No public API surface change — this is an internal UI sync fix in the date field. No docs update required.

## Tests
Covered by the new UI test `testDateFieldClearSyncsAcrossMirroredViews` (iPad-only), which exercises both setting and clearing a date and asserts the mirrored view stays in sync.

## Notes for reviewer
- The mirrored-view test relies on page 3 having two date fields bound to the same field id; behavior is asserted via the `ChangeDateIdentifier` / `DateClearIdentifier` / placeholder labels.
- An earlier attempt (`494c82bf`) was reverted in favor of this narrower one-line fix that lives entirely in the model-value onChange path.

### DIFF
--- a/JoyfillSwiftUIExample/JoyfillUITests/ChangeUITests/OnChangeHandlerUITests.swift
+++ b/JoyfillSwiftUIExample/JoyfillUITests/ChangeUITests/OnChangeHandlerUITests.swift
@@ -2393,6 +2393,65 @@ final class OnChangeHandlerUITests: JoyfillUITestsBaseClass {
         let dateButton2 = getDateFieldButtonByIndexAndIdentifier(0)
         XCTAssertEqual(dateButton2.label, convertedTime, "Datetime should be same after epoch convert")
     }
+
+    func testDateFieldClearSyncsAcrossMirroredViews() throws {
+        guard UIDevice.current.userInterfaceIdiom == .pad else {
+            return
+        }
+
+        let pageSelectionButton = app.buttons.matching(identifier: "PageNavigationIdentifier")
+        let pageSheetSelectionButton = app.buttons.matching(identifier: "PageSelectionIdentifier")
+        let firstPageNavigationButton = pageSelectionButton.element(boundBy: 0)
+        let secondPageNavigationButton = pageSelectionButton.element(boundBy: 1)
+        let pageSelectionItem = pageSheetSelectionButton.element(boundBy: 2)
+
+        XCTAssertTrue(firstPageNavigationButton.waitForExistence(timeout: 5), "First page navigation button not found")
+        firstPageNavigationButton.tap()
+        XCTAssertTrue(pageSelectionItem.waitForExistence(timeout: 5), "Page 3 selection item not found")
+        pageSelectionItem.tap()
+
+        XCTAssertTrue(secondPageNavigationButton.waitForExistence(timeout: 5), "Second page navigation button not found")
+        secondPageNavigationButton.tap()
+        XCTAssertTrue(pageSelectionItem.waitForExistence(timeout: 5), "Page 3 selection item not found")
+        pageSelectionItem.tap()
+
+        let emptyDatePlaceholders = self.app.staticTexts.matching(NSPredicate(format: "label == %@", "Select a Date -"))
+        XCTAssertTrue(
+            waitUntil(5) { emptyDatePlaceholders.count == 2 },
+            "Expected both mirrored empty date fields to be visible on page 3"
+        )
+
+        let leftDateField = emptyDatePlaceholders.element(boundBy: 0)
+        leftDateField.tap()
+
+        let mirroredDateButtons = self.app.buttons.matching(identifier: "ChangeDateIdentifier")
+        XCTAssertTrue(
+            waitUntil(5) { mirroredDateButtons.count == 2 },
+            "Expected setting the left-side date to mirror to the right side"
+        )
+
+        let leftDateButton = mirroredDateButtons.element(boundBy: 0)
+        let rightDateButton = mirroredDateButtons.element(boundBy: 1)
+        XCTAssertEqual(leftDateButton.label, rightDateButton.label, "Mirrored date value should match on both sides")
+
+        let clearButtons = self.app.buttons.matching(identifier: "DateClearIdentifier")
+        XCTAssertTrue(
+            waitUntil(5) { clearButtons.count == 2 },
+            "Expected both mirrored date fields to expose a clear button"
+        )
+
+        clearButtons.element(boundBy: 0).tap()
+
+        XCTAssertTrue(
+            waitUntil(5) { self.app.buttons.matching(identifier: "ChangeDateIdentifier").count == 0 },
+            "Expected clearing the left-side date to remove the mirrored date value on the right side"
+        )
+
+        XCTAssertTrue(
+            waitUntil(5) { emptyDatePlaceholders.count == 2 },
+            "Expected both mirrored date fields to show the empty placeholder after clearing"
+        )
+    }
     
     func testSignatureField() throws {
         guard UIDevice.current.userInterfaceIdiom == .pad else {
--- a/JoyfillSwiftUIExample/JoyfillUITests/ChangeUITests/OnChangeHandlerUITests.swift
+++ b/JoyfillSwiftUIExample/JoyfillUITests/ChangeUITests/OnChangeHandlerUITests.swift
@@ -2424,6 +2424,10 @@ final class OnChangeHandlerUITests: JoyfillUITestsBaseClass {
         let leftDateField = emptyDatePlaceholders.element(boundBy: 0)
         leftDateField.tap()
 
+        // Two date buttons appear for a field that exists only once in the JSON because
+        // OnChangeHandlerTest.swift renders two DocumentEditor instances side-by-side and
+        // plumbs editor #1's onChange into editor #2. Editing/clearing on the left mirrors
+        // to the right, which is exactly what this test verifies. See OnChangeHandlerTest.swift.
         let mirroredDateButtons = self.app.buttons.matching(identifier: "ChangeDateIdentifier")
         XCTAssertTrue(
             waitUntil(5) { mirroredDateButtons.count == 2 },
--- a/Sources/JoyfillUI/View/Fields/DateTimeView.swift
+++ b/Sources/JoyfillUI/View/Fields/DateTimeView.swift
@@ -111,7 +111,7 @@ struct DateTimeView: View {
         }
         .onChange(of: dateTimeDataModel.value) { newValue in
             if lastModelValue != newValue {
-                if let value = newValue {
+                if let value = newValue, !value.nullOrEmpty {
                     let dateString = value.dateTime(format: dateTimeDataModel.format ?? .empty, tzId: dateTimeDataModel.timezoneId) ?? ""
                     if let date = Utility.stringToDate(dateString, format: dateTimeDataModel.format ?? .empty, tzId: dateTimeDataModel.timezoneId) {
                         selectedDate = date
--- a/Sources/JoyfillUI/View/Fields/DateTimeView.swift
+++ b/Sources/JoyfillUI/View/Fields/DateTimeView.swift
@@ -111,12 +111,13 @@ struct DateTimeView: View {
         }
         .onChange(of: dateTimeDataModel.value) { newValue in
             if lastModelValue != newValue {
-                if let value = newValue {
+                if let value = newValue, !value.nullOrEmpty {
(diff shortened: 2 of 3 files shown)


Sources/JoyfillUI/View/Fields/DateTimeView.swift
```diff
@@ -117,6 +117,7 @@ struct DateTimeView: View {
                         selectedDate = date
                     }
                 } else {
+                    self.dateString = ""
```
vivekjyani, line 120: **Blocking — fix likely incomplete for change-API and mirrored-view paths.**

This branch only runs when `newValue` is literally `nil`, but the value emitted on clear is `ValueUnion.null` (see `eventHandler.onChange` on line 54 — `updateValue: ValueUnion.null`). That `.null` propagates through:

- `DocumentEditor.updateField` (`field.value = event.updateValue` → `.null`, not nil)
- `getFieldModel` for date: `fieldData?.value?.normalizedNumericValue` → `.null` (`normalizedNumericValue` returns `self` when `doubleValue` is nil — see DocumentEditor.swift:828)
- So `dateTimeDataModel.value` arrives as `Optional.some(.null)`, **non-nil**.

Result: on the mirrored view (and on any field cleared via the external `change` API with `value: null`), execution enters the `if let value = newValue` branch above, never this `else`, so `dateString` is never reset.

The reverted commit `494c82bf` used `if let value = newValue, !value.nullOrEmpty` — that handled both `nil` and `.null` and was actually the right shape. Suggest restoring that guard (or equivalently, gating on `newValue == nil || newValue?.nullOrEmpty == true` here).

JoyfillSwiftUIExample/JoyfillUITests/ChangeUITests/OnChangeHandlerUITests.swift
```diff
@@ -2393,6 +2393,65 @@ final class OnChangeHandlerUITests: JoyfillUITestsBaseClass {
         let dateButton2 = getDateFieldButtonByIndexAndIdentifier(0)
         XCTAssertEqual(dateButton2.label, convertedTime, "Datetime should be same after epoch convert")
     }
+
+    func testDateFieldClearSyncsAcrossMirroredViews() throws {
+        guard UIDevice.current.userInterfaceIdiom == .pad else {
```
vivekjyani, line 2398: **Blocking — test almost certainly does not gate this fix in CI.**

`gh pr checks 334` shows **UI Tests: skipping** for this PR — only Unit Tests run. Combined with the `userInterfaceIdiom == .pad` early-return, this test cannot catch a regression on either CI or non-iPad local runs.

If the goal is automated regression coverage for the change-API clear path, please add a unit test under `JoyfillSwiftUIExample/JoyfillTests/` that:

1. Builds a `DocumentEditor` with a date field that has a value.
2. Calls `documentEditor.change(changes: [...])` with `"value": NSNull()`.
3. Asserts the resulting `DateTimeDataModel.value` and/or the rendered `dateString` is cleared.

A test at that layer doesn't need an iPad simulator, runs in the existing Unit Tests job, and would have actually caught the issue I flagged on `DateTimeView.swift:120` (where `.null` is silently passed through the `if let value = newValue` branch).

JoyfillSwiftUIExample/JoyfillUITests/ChangeUITests/OnChangeHandlerUITests.swift
```diff
@@ -2393,6 +2393,65 @@ final class OnChangeHandlerUITests: JoyfillUITestsBaseClass {
...
+        XCTAssertTrue(firstPageNavigationButton.waitForExistence(timeout: 5), "First page navigation button not found")
+        firstPageNavigationButton.tap()
+        XCTAssertTrue(pageSelectionItem.waitForExistence(timeout: 5), "Page 3 selection item not found")
+        pageSelectionItem.tap()
+
+        XCTAssertTrue(secondPageNavigationButton.waitForExistence(timeout: 5), "Second page navigation button not found")
+        secondPageNavigationButton.tap()
+        XCTAssertTrue(pageSelectionItem.waitForExistence(timeout: 5), "Page 3 selection item not found")
+        pageSelectionItem.tap()
+
```
vivekjyani, line 2417: Nit — the "mirrored view" the PR description and test name reference is implicit setup in `OnChangeHandlerTest.swift` (two `DocumentEditor` instances rendered side-by-side, with `onChange` plumbing one to the other). A future reader landing here will have no idea why there should be 2 placeholders / 2 ChangeDateIdentifier buttons for a field that only exists once in the JSON. A one-line comment above this assertion pointing at `OnChangeHandlerTest.swift` would save the next debugger.

Sources/JoyfillUI/View/Fields/DateTimeView.swift
vivekjyani: Fixed

JoyfillSwiftUIExample/JoyfillUITests/ChangeUITests/OnChangeHandlerUITests.swift
vivekjyani: Thanks for catching this — both calls are fair.

On the View fix: restored the nullOrEmpty guard at DateTimeView.swift:114 so the else-branch fires for both nil and .some(.null). That covers the change-API / mirrored-view path you flagged.

On the test coverage: added two unit tests in DocumentEditor+ChangeHandlerChartAxisTests.swift under the // MARK: - Date field clear via change API section:

testDateFieldUpdate_NumericPayload_SetsFieldValue — sanity check that a numeric value payload populates the date field.
testDateFieldClearViaChangeAPI_NullPayload_ClearsFieldValue — sets a value, then clears via {"value": NSNull()}, asserts field.value is null-or-empty. This pins the data-layer contract the View fix depends on (after a null clear, the stored value is .null, not nil — so the View must check nullOrEmpty, not just if let).
Tests run in the existing Unit Tests CI job, no iPad simulator required.

The UI test approach you originally pointed to is the right shape for catching the actual UI regression — happy to leave the existing testDateFieldClearSyncsAcrossMirroredViews in place for that. The wrinkle there is that UI Tests are configured workflow_dispatch-only at pr_pipeline_workflow.yml:268, so they don't run on PR events by default. Making UI Tests gate this regression in CI is really a CI-policy change (enable on PRs, accept the +8–12 min) rather than a code change — happy to file a separate ticket if you want to pursue that.

JoyfillSwiftUIExample/JoyfillUITests/ChangeUITests/OnChangeHandlerUITests.swift
vivekjyani: Good call — done. Added a 4-line comment above the mirroredDateButtons assertion explaining that the two ChangeDateIdentifier buttons come from OnChangeHandlerTest.swift rendering two mirrored DocumentEditor instances (editor #1's onChange feeds editor #2), and pointing the reader there. No logic change.

Sources/JoyfillUI/View/Fields/DateTimeView.swift
```diff
@@ -111,12 +111,13 @@ struct DateTimeView: View {
         }
         .onChange(of: dateTimeDataModel.value) { newValue in
             if lastModelValue != newValue {
-                if let value = newValue {
+                if let value = newValue, !value.nullOrEmpty {
                     let dateString = value.dateTime(format: dateTimeDataModel.format ?? .empty, tzId: dateTimeDataModel.timezoneId) ?? ""
```
VishnuBishnoi, line 115: Non-blocking nit (pre-existing, not introduced by this PR): this local `let dateString` shadows the `@State var dateString` and is only used to derive `selectedDate` — the visible label is actually updated indirectly via `onChange(of: selectedDate)`. It reads as if it sets the display but doesn't. Could be renamed (e.g. `formatted`) or the flow simplified for clarity in a follow-up.

JoyfillSwiftUIExample/JoyfillUITests/ChangeUITests/OnChangeHandlerUITests.swift
```diff
@@ -2393,6 +2393,69 @@ final class OnChangeHandlerUITests: JoyfillUITestsBaseClass {
         let dateButton2 = getDateFieldButtonByIndexAndIdentifier(0)
         XCTAssertEqual(dateButton2.label, convertedTime, "Datetime should be same after epoch convert")
     }
+
+    func testDateFieldClearSyncsAcrossMirroredViews() throws {
+        guard UIDevice.current.userInterfaceIdiom == .pad else {
```
VishnuBishnoi, line 2398: Non-blocking: this UI test early-returns (passes vacuously) on iPhone, so it only provides coverage on iPad. Consistent with the rest of this file, and the new unit tests are device-agnostic so the core contract is covered everywhere — just flagging that the mirrored-view regression itself is iPad-only in CI.

VishnuBishnoi (approved):
Reviewed locally: checked out the branch, read the fix in full context, verified the `ValueUnion.nullOrEmpty` contract it depends on, confirmed the fixture IDs resolve, and ran the two new unit tests on a simulator — **both pass (0 failures, ~0.02s)**.

**The fix is correct and minimal.** Clearing via the change API sends `{"value": null}`, which becomes `ValueUnion.null` on the field — *not* Swift `nil`. So the old `if let value = newValue` succeeded (`.null` is non-nil), took the "set" branch, and never cleared the `@State dateString` that drives the visible label — so the stale date persisted. Adding `!value.nullOrEmpty` correctly routes `.null` into the `else` branch, and `self.dateString = ""` restores the placeholder. Verified against `ValueUnion.nullOrEmpty`: `.null → true`, `.double(timestamp) → false`, so set stays set and a legit `0` epoch is not falsely cleared. The existing `ignoreOnChangeOnModelUpdate` guard still suppresses the echo event, so clearing doesn't re-emit a spurious change.

Test coverage is well-targeted: the 2 unit tests pin the data-layer contract the view guard relies on, and the UI test covers the real mirrored-view sync end-to-end. Branch hygiene is clean — the earlier attempt (`494c82bf`) was properly reverted (`982ab4e6`); net diff vs `main` is just the 3 expected files.

Approving. Inline notes below are minor/non-blocking nits.

**Additional non-blocking note** (in an unchanged file, so noting here rather than inline): `ValueUnion.nullOrEmpty` (ValueUnion.swift:57) has two latent quirks — `.double(let double): return double == nil` compares a non-optional `Double` to `nil` (always false), and `.bool(let bool): return bool` is semantically odd. Harmless for the date path the fix exercises, but worth a future cleanup.

Sources/JoyfillUI/View/Fields/DateTimeView.swift
vivekjyani: Good catch, and agreed it reads a bit misleadingly. Since it's pre-existing and purely cosmetic (the local only derives selectedDate; the visible label updates via onChange(of: selectedDate)), I'd prefer to leave it out of this PR to keep the diff focused on the clear-sync fix. I'll track the rename/cleanup as a separate follow-up.

JoyfillSwiftUIExample/JoyfillUITests/ChangeUITests/OnChangeHandlerUITests.swift
vivekjyani: Agreed — flagging only, nothing to change here. The iPad guard is consistent with the rest of this UI suite, and the new device-agnostic unit tests cover the clear-via-change-API contract on all targets/CI. The UI test stays as the iPad-only check on the actual mirrored-view rendering.

PR #335, opened 2026-06-04
NO-2184: Add decorator definitions to JSON validation schema

vivekjyani:
## Context

The Decorator feature lets fields, table/collection rows, and cells carry small action badges (icon/label/color). The SDK already exposes the runtime API (`get/add/update/removeDecorator`) and renders them, but the bundled JSON validation schema (`joyfillSchema`) did not describe these properties. Documents using decorators therefore risked failing schema validation. This PR brings the schema in line with the feature.

## What changed

Single file: `Sources/JoyfillUI/View/joyfill-schema.swift`.

- Added two new definitions:
  - `Decorator` — `{ _id, action, color, icon, label }`, with `action` required and an `anyOf` requiring at least one of `icon` / `label`.
  - `RowDecorators` — `{ all: [Decorator], cells: { <columnId>: [Decorator] } }`.
- Added a `decorators: [Decorator]` array to every field-type definition (text, number, date, dropdown, signature, image, multiline, chart, block, collection, table, etc.).
- Added `rowDecorators: [Decorator]` and `decorate: boolean` to table/collection field definitions.
- Added `decorators: RowDecorators` to the row / `CollectionItem` definitions.
- Added `columnTitleTextOverflow: string` to the field-position definition.
- Bumped `$joyfillSchemaVersion` `1.0.1` → `1.0.4`.

## Decisions

- **`action` required + `anyOf(icon|label)`** mirrors the runtime validation in `DocumentEditor+Decorators.validateDecorator` (action mandatory; a decorator must be displayable via an icon or a label). Keeping the schema and runtime rules aligned avoids documents that pass one but fail the other.
- **`decorators` added per field type rather than to a shared base** — the schema defines each field type explicitly with its own `required`/`properties`, so the property is added in place to match the existing structure rather than introducing a shared mixin.
- **Version bump to 1.0.4** to reflect the additive schema changes shipped since 1.0.1.

## Screenshot / video

N/A — schema-only change, no UI.

## Public API

- No Swift API surface change (no structs/methods/signatures touched).
- `joyfillSchema` is a `public var`; its **content** changed, which loosens validation to accept decorator properties. Purely additive — previously valid documents remain valid.
- Docs: the decorator schema/spec should be reflected in the public schema docs if they're versioned there.

## Tests

- Schema-content change; covered indirectly by existing schema-validation tests (`JoyfillTests/SchemaValidation/`). No new behavior logic to unit-test here.
- If we want an explicit guard, a follow-up test asserting a decorator-bearing document validates against `joyfillSchema` would lock it in.

## Notes for reviewer

- Diff is large but mechanical/repetitive — the same `decorators` block is added to each field type. Worth a quick scan that no field type was missed and that `required` arrays were not altered.

### DIFF
--- a/Sources/JoyfillUI/View/joyfill-schema.swift
+++ b/Sources/JoyfillUI/View/joyfill-schema.swift
@@ -419,6 +419,9 @@ public var joyfillSchema = """
           "columnTitlePadding": {
             "type": "number"
           },
+          "columnTitleTextOverflow": {
+            "type": "string"
+          },
           "titleDisplay": {
             "type": "string"
           },
@@ -676,6 +679,12 @@ public var joyfillSchema = """
           },
           "multi": {
             "type": "boolean"
+          },
+          "decorators": {
+            "type": "array",
+            "items": {
+              "$ref": "#/definitions/Decorator"
+            }
           }
         },
         "required": [
@@ -783,6 +792,12 @@ public var joyfillSchema = """
           },
           "multi": {
             "type": "boolean"
+          },
+          "decorators": {
+            "type": "array",
+            "items": {
+              "$ref": "#/definitions/Decorator"
+            }
           }
         },
         "required": [
@@ -883,6 +898,12 @@ public var joyfillSchema = """
           },
           "value": {
             "type": "string"
+          },
+          "decorators": {
+            "type": "array",
+            "items": {
+              "$ref": "#/definitions/Decorator"
+            }
           }
         },
         "required": [
@@ -962,6 +983,12 @@ public var joyfillSchema = """
           },
           "value": {
             "type": "string"
+          },
+          "decorators": {
+            "type": "array",
+            "items": {
+              "$ref": "#/definitions/Decorator"
+            }
           }
         },
         "required": [
@@ -1041,6 +1068,12 @@ public var joyfillSchema = """
           },
           "value": {
             "type": "string"
+          },
+          "decorators": {
+            "type": "array",
+            "items": {
+              "$ref": "#/definitions/Decorator"
+            }
           }
         },
         "required": [
@@ -1128,6 +1161,12 @@ public var joyfillSchema = """
                 "const": ""
               }
             ]
+          },
+          "decorators": {
+            "type": "array",
+            "items": {
+              "$ref": "#/definitions/Decorator"
+            }
           }
         },
         "required": [
@@ -1221,6 +1260,12 @@ public var joyfillSchema = """
           },
           "format": {
             "type": "string"
+          },
+          "decorators": {
+            "type": "array",
+            "items": {
+              "$ref": "#/definitions/Decorator"
+            }
           }
         },
         "required": [
@@ -1300,6 +1345,12 @@ public var joyfillSchema = """
           },
           "value": {
             "type": "string"
+          },
+          "decorators": {
+            "type": "array",
+            "items": {
+              "$ref": "#/definitions/Decorator"
+            }
           }
         },
         "required": [
@@ -1382,6 +1433,12 @@ public var joyfillSchema = """
           },
           "signer": {
             "type": "string"
+          },
+          "decorators": {
+            "type": "array",
+            "items": {
+              "$ref": "#/definitions/Decorator"
+            }
           }
         },
         "required": [
@@ -1473,6 +1530,12 @@ public var joyfillSchema = """
           },
           "multi": {
             "type": "boolean"
+          },
+          "decorators": {
+            "type": "array",
+            "items": {
+              "$ref": "#/definitions/Decorator"
+            }
           }
         },
         "required": [
@@ -1594,6 +1657,12 @@ public var joyfillSchema = """
           },
           "value": {
             "type": "string"
+          },
+          "decorators": {
+            "type": "array",
+            "items": {
+              "$ref": "#/definitions/Decorator"
+            }
           }
         },
         "required": [
@@ -1695,6 +1764,21 @@ public var joyfillSchema = """
             "items": {
               "type": "string"
             }
+          },
+          "decorators": {
+            "type": "array",
+            "items": {
+              "$ref": "#/definitions/Decorator"
+            }
+          },
+          "rowDecorators": {
+            "type": "array",
+            "items": {
+              "$ref": "#/definitions/Decorator"
+            }
+          },
+          "decorate": {
+            "type": "boolean"
           }
         },
         "required": [
@@ -1719,6 +1803,9 @@ public var joyfillSchema = """
           },
           "cells": {
             "type": "object"
+          },
+          "decorators": {
+            "$ref": "#/definitions/RowDecorators"
           }
         },
         "required": [
@@ -1784,6 +1871,12 @@ public var joyfillSchema = """
           },
           "value": {
             "type": "string"
+          },
+          "decorators": {
+            "type": "array",
+            "items": {
+              "$ref": "#/definitions/Decorator"
+            }
           }
         },
         "required": [
@@ -1821,6 +1914,12 @@ public var joyfillSchema = """
             "type": "array",
             "items": {
               "$ref": "#/definitions/Option"
+            }
+          },
+          "decorators": {
+            "type": "array",
+            "items": {
+              "$ref": "#/definitions/Decorator"
             }
           }
         },
@@ -1863,6 +1962,12 @@ public var joyfillSchema = """
             "items": {
               "$ref": "#/definitions/Option"
             }
+          },
+          "decorators": {
+            "type": "array",
+            "items": {
+              "$ref": "#/definitions/Decorator"
+            }
           }
         },
         "required": [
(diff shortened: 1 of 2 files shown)


Sources/JoyfillUI/View/joyfill-schema.swift
```diff
@@ -2565,6 +2748,54 @@ public var joyfillSchema = """
...
+            "type": "string"
+          },
+          "label": {
+            "type": "string"
+          }
+        },
+        "required": [
+          "action"
+        ],
+        "anyOf": [
```
vivekjyani, line 2774: **Schema is stricter than the runtime here.** This `anyOf` forces every decorator to carry an `icon` or a `label`, but the runtime `validateDecorator` (`DocumentEditor+Decorators.swift`) only requires a non-empty `action` — it allows a non-displayable, action-only decorator and `addDecorators`/`updateDecorator` will happily persist one.

Consequence: a decorator created via the public API with `action` only passes runtime validation but would **fail schema validation** on reload. We should align the two — either drop this `anyOf` (match the lenient runtime) or tighten `validateDecorator` to require displayability (match the schema). Flagging so it's a deliberate decision, not an accidental gap.

Sources/JoyfillUI/View/joyfill-schema.swift
```diff
@@ -2565,6 +2748,54 @@ public var joyfillSchema = """
           "type"
         ]
       },
+      "Decorator": {
+        "type": "object",
+        "properties": {
+          "_id": {
```
vivekjyani, line 2754: `_id` (minLength 24) is defined here, but the runtime `Decorator` struct (`JoyDoc.swift`) has no `_id` accessor and `addDecorators` never generates one — SDK-created decorators ship without an `_id`. It's optional so nothing breaks, but if the backend or other SDKs expect `_id` on decorators, the Swift SDK isn't setting it. Confirm this is intended.

Sources/JoyfillUI/View/joyfill-schema.swift
```diff
@@ -2594,7 +2825,7 @@ public var joyfillSchema = """
         ]
       }
     },
-    "$joyfillSchemaVersion": "1.0.1"
+    "$joyfillSchemaVersion": "1.0.4"
```
vivekjyani, line 2828: Version jumps `1.0.1` → `1.0.4` in a single additive commit. If the schema is shared across SDKs and 1.0.4 is the canonical current version, fine — just confirming this isn't a typo for `1.0.2`.

Sources/JoyfillUI/View/joyfill-schema.swift
vivekjyani: Intentional — the stricter schema is by design. A decorator with no icon **and** no label is non-displayable, so requiring at least one at the document level keeps persisted documents meaningful. The runtime stays lenient on purpose (it won't hard-reject an action-only decorator), but the schema is the source of truth for what a valid document should contain.

Sources/JoyfillUI/View/joyfill-schema.swift
vivekjyani: Intentional — `_id` isn't really part of a decorator. The unique identifier for a decorator is its `action` (unique per path), which is why the SDK never generates an `_id`. The schema field is just optional/permissive for compatibility; nothing depends on it.

Sources/JoyfillUI/View/joyfill-schema.swift
vivekjyani: Intentional — the web/canonical schema is ahead of the Swift SDK and went through 3 updates (1.0.2 → 1.0.3 → 1.0.4). This bump syncs us to the current canonical version 1.0.4, not a typo.

Sources/JoyfillUI/View/joyfill-schema.swift
VishnuBishnoi, line 2774: This `anyOf` (require icon OR label) has no counterpart in the runtime. `validateDecorator` (DocumentEditor+Decorators.swift:178) checks only `action` (required, non-empty) and `color` (hex if present) — it never requires icon or label. So an action-only decorator added through the runtime API validates at runtime but fails this schema. Either drop this `anyOf` to match the runtime, or add the icon||label check to `validateDecorator` so both agree. The PR description says this block "mirrors the runtime validation" — it currently doesn't.

Sources/JoyfillUI/View/joyfill-schema.swift
```diff
@@ -2565,6 +2748,54 @@ public var joyfillSchema = """
...
+        "type": "object",
+        "properties": {
+          "_id": {
+            "type": "string",
+            "minLength": 24
+          },
+          "action": {
+            "type": "string"
+          },
+          "color": {
```
VishnuBishnoi, line 2761: Runtime `validateDecorator` enforces `color` matches `^#[0-9A-Fa-f]{6}$`, but here `color` is just `type: string`. Not a blocker (looser = additive), but if the goal is schema/runtime parity, add `"pattern": "^#[0-9A-Fa-f]{6}$"` here.

VishnuBishnoi (commented):
Reviewed locally: checked out the branch, extracted the embedded `joyfillSchema` string literal and **parsed it as JSON (valid)**, scanned for duplicate keys (none), confirmed both new defs resolve (`#/definitions/Decorator` referenced 30×, `RowDecorators` 2×), and ran `JoyfillTests/SchemaValidationTests` — **22/22 pass**, including `testVersion` now matching `1.0.4`.

Mechanics are solid: JSON well-formed, version bump + test sync correct, additive props don't touch any `required` array, defs match the documented shape.

**One thing to resolve before merge — `anyOf(icon|label)` does NOT mirror the runtime (contrary to the PR description).** `DocumentEditor+Decorators.validateDecorator` only enforces (a) `action` present + non-empty and (b) `color`, if set, is `^#RRGGBB$`. There is **no** icon-or-label requirement anywhere in the add/update path (grep for icon/label enforcement in `DocumentEditor+Decorators.swift` returns nothing). So a decorator created via the runtime API with `action` only (no icon, no label) is accepted at runtime but **rejected by this schema** — the exact pass-one-fail-the-other divergence this PR set out to prevent, just in the opposite direction. Pick one: drop the `anyOf` from the schema (match runtime), or add the icon||label rule to `validateDecorator` (match schema). See inline.

Secondary: schema `color` is a bare `type: string` while runtime requires a `#RRGGBB` hex — schema is looser there; harmless (additive) but also not actually "mirrored". Inline note with optional `pattern`.

Minor (non-blocking): no object in the schema uses `additionalProperties: false`, so unknown props were already accepted — decorator-bearing docs likely weren't hard-failing validation before. The change is still worthwhile as schema documentation + the version bump, but the PR's "risked failing schema validation" framing is weaker than stated.

Not blocking the mechanics — just want the icon/label decision settled so schema and runtime agree.

Sources/JoyfillUI/View/joyfill-schema.swift
vivekjyani: Good point on the color pattern — going to skip it. The runtime validateDecorator already enforces ^#[0-9A-Fa-f]{6}$ and surfaces violations via onError, so adding the same regex as a JSON-schema pattern would just be a second place to keep in sync for the same check, with no added safety. Schema stays "type": "string" for color.

Sources/JoyfillUI/View/joyfill-schema.swift
vivekjyani: Going to leave this as-is. The anyOf stays; validateDecorator and isDisplayable stay as they are. Acknowledged that the three layers aren't perfectly aligned, but in practice action-only decorators aren't something we expect to see in real documents, and the render-time isDisplayable filter keeps the UI clean either way. Not worth a behavior change in validateDecorator for this.

VishnuBishnoi (approved):
Re-reviewed after the comment resolutions. Approving.

**anyOf(icon|label) — resolved, keeping as-is is correct.** My original note was incomplete: I grepped only `DocumentEditor+Decorators.swift` and missed `Decorator.isDisplayable` in `JoyfillModel/JoyDoc.swift:50` — which is exactly `hasIcon || hasLabel`, i.e. the runtime counterpart to this `anyOf`. It's enforced at render time at 15+ sites (`.filter { $0.isDisplayable }` across Models/DocumentEditor/CollectionViewModel/DecoratorView), so a non-displayable action-only decorator is silently dropped from the UI. The only remaining gap is that `validateDecorator` stays lenient (won't hard-reject action-only) while the schema + render layer require displayability — and the author's call to keep the schema as the document-level source of truth is a sound, deliberate decision. Agreed.

**color pattern — fine to skip.** It was an optional parity nit; leaving `color` as a bare string keeps the schema looser than runtime, which is additive/harmless.

Mechanics already verified in the prior pass and unchanged since (schema literal parses as valid JSON, no dup keys, refs resolve, `SchemaValidationTests` 22/22 incl. `testVersion` 1.0.4). No code changed in resolution, so nothing to re-test. LGTM.

PR #336, opened 2026-06-10
NO-2186: Fix date field rendering current time for empty string value

vivekjyani:
## Context
A `date` field whose stored `value` is an empty string (`""`) — or any invalid ISO8601 string — was rendering the **current** date/time instead of an empty state. Root cause: `getTimeFromISO8601Format` fell back to `Date()` (now) when ISO8601 parsing failed, and it also ignored the requested display format (always emitting `hh:mm a`).

## What changed
- **`Sources/JoyfillModel/DocumentModel.swift`** — `getTimeFromISO8601Format(iso8601String:format:tzId:)`:
  - Returns `nil` for invalid/empty ISO8601 input instead of falling back to the current date.
  - Now honors the passed `format` (new `format: DateFormatType = .empty` parameter) rather than hardcoding `hh:mm a`.
  - Applies the same 12h/24h locale handling used by `timestampMillisecondsToDate`, so the string and numeric value paths format identically.
- **`Sources/JoyfillModel/ValueUnion.swift`** — `dateTime(format:tzId:)` guards empty strings and forwards `format` into `getTimeFromISO8601Format`.
- **`DateTimeFieldTestData.json` / `DateTimeFieldTests.swift`** — added a `"Test empty date string"` field (`value: ""`) and the UI test `testEmptyStringDateRendersEmptyState` (empty placeholder shows, then tapping sets a numeric timestamp).
- **`JoyfillTests/DateFormattingTests.swift`** (new) — unit tests for `getTimeFromISO8601Format` and `ValueUnion.dateTime` covering empty/garbage/null/valid input and int/double timestamps.

## Implementation notes
- Returning `nil` for any unparseable input (not just empty string) unifies the empty-state path and removes the silent "now" fabrication.
- Locale handling mirrors `timestampMillisecondsToDate` (`en_GB` for `HH`, `en_US_POSIX` for `hh`/`.empty`) so a value formats the same whether it arrives as an ISO string or a millisecond timestamp.

## Public API
**Breaking change** to the public `getTimeFromISO8601Format`:
- Return type changed from `String` to `String?`.
- New `format: DateFormatType = .empty` parameter (defaulted, source-compatible for call sites that don't pass it, but the optional return type still requires unwrapping).

Only one internal caller (`ValueUnion.dateTime`) exists and is updated. External consumers of this public function will need to handle the optional return. **Docs should be updated** to reflect the new signature and nil-on-invalid behavior.

## Tests
Covered by the new unit tests (`DateFormattingTests`) and the UI test (`testEmptyStringDateRendersEmptyState`): empty, garbage, null, valid ISO, and int/double timestamp cases. No further cases planned for now.

## Notes for reviewer
- Worth confirming the locale-by-format heuristic matches `timestampMillisecondsToDate` for every `DateFormatType` case.
- Screenshot/video omitted (iOS simulator UI run); behavior change: empty/invalid date field now shows `Select a Date -` instead of the current time.

### DIFF
--- a/JoyfillSwiftUIExample/JoyfillUITests/DateTimeField/DateTimeFieldTestData.json
+++ b/JoyfillSwiftUIExample/JoyfillUITests/DateTimeField/DateTimeFieldTestData.json
@@ -117,6 +117,14 @@
       "identifier": "field_695cb7ba2bb03caceb8ae978",
       "value": 1767252026000,
       "tz": "Asia/Calcutta"
+    },
+    {
+      "file": "66a14ced9dc829a95e272506",
+      "_id": "6a27d28b656b1bbecbb2a956",
+      "type": "date",
+      "title": "Test empty date string",
+      "value": "",
+      "identifier": "field_6a27d28b656b1bbecbb2a956"
     }
   ],
   "_id": "66a14cedd6e1ebcdf176a8da",
@@ -273,6 +281,17 @@
               "width": 4,
               "height": 8,
               "field": "695cb7ba2bb03caceb8ae978"
+            },
+            {
+              "_id": "6a27d28b81ee0056ebf9f314",
+              "type": "date",
+              "displayType": "original",
+              "format": "MM/DD/YYYY hh:mma",
+              "x": 1,
+              "y": 96,
+              "width": 4,
+              "height": 8,
+              "field": "6a27d28b656b1bbecbb2a956"
             }
           ],
           "hidden": false,
@@ -281,7 +300,12 @@
           "_id": "66a14ced15a9dc96374e091e",
           "width": 816,
           "name": "Every Field Type",
-          "margin": 0
+          "margin": 0,
+          "deletable": true,
+          "copyable": [
+            "with-values",
+            "without-values"
+          ]
         }
       ]
     }
--- a/JoyfillSwiftUIExample/JoyfillUITests/DateTimeField/DateTimeFieldTests.swift
+++ b/JoyfillSwiftUIExample/JoyfillUITests/DateTimeField/DateTimeFieldTests.swift
@@ -1,6 +1,40 @@
 import XCTest
 
 final class DateTimeFieldTests: JoyfillUITestsBaseClass {
+    override func getJSONFileNameForTest() -> String {
+        return "DateTimeFieldTestData"
+    }
+
+    func testEmptyStringDateRendersEmptyState() {
+        // A date field with value "" must render the empty "Select a Date -" placeholder,
+        // not a fabricated current date/time.
+        let placeholder = app.staticTexts["Select a Date -"]
+
+        var attempts = 0
+        while !placeholder.exists && attempts < 8 {
+            app.swipeUp()
+            _ = placeholder.waitForExistence(timeout: 1)
+            attempts += 1
+        }
+
+        XCTAssertTrue(placeholder.waitForExistence(timeout: 5),
+                      "Empty-string date field should show the 'Select a Date -' empty state")
+        let placeholderQuery = app.staticTexts.matching(NSPredicate(format: "label == %@", "Select a Date -"))
+        XCTAssertEqual(placeholderQuery.count, 1,
+                       "Only the empty-string date field should show the empty placeholder")
+
+        // Now set a date/time on the previously-empty field and confirm it gets set.
+        placeholder.tap()
+
+        XCTAssertTrue(waitUntil(5) { placeholderQuery.count == 0 },
+                      "After tapping, the empty placeholder should disappear (field is now filled)")
+
+        // A change should have been emitted carrying a numeric (millisecond) value.
+        let emittedValue = onChangeResultValue()
+        XCTAssertNotNil(emittedValue.number,
+                        "Setting the date should emit a numeric timestamp value, got \(emittedValue)")
+    }
+
     func testDatePicker() {
         app.swipeUp()
         // Tap a date button using its identifier to open the picker popup
--- a/Sources/JoyfillModel/DocumentModel.swift
+++ b/Sources/JoyfillModel/DocumentModel.swift
@@ -278,25 +278,35 @@ public enum DateFormatType: String {
     }
 }
 
-/// Converts an ISO8601 formatted string to a time string.
+/// Converts an ISO8601 formatted string to a formatted date/time string.
 ///
 /// - Parameters:
 ///   - iso8601String: The ISO8601 formatted string representing a date and time.
+///   - format: The desired output format. Defaults to `.empty` (MM/dd/yyyy hh:mma).
+///   - tzId: Optional timezone identifier to render the value in.
 ///
-/// - Returns: A formatted time string in the format "hh:mm a".
-public func getTimeFromISO8601Format(iso8601String: String, tzId: String? = nil) -> String {
-    let dateFormatter = ISO8601DateFormatter()
-    let instant = dateFormatter.date(from: iso8601String)
-    
+/// - Returns: A formatted date/time string honoring `format`, or `nil` if the
+///   input is not a valid ISO8601 string. Returning `nil` (rather than falling
+///   back to the current date) lets callers render an empty state for invalid input.
+public func getTimeFromISO8601Format(iso8601String: String, format: DateFormatType = .empty, tzId: String? = nil) -> String? {
+    guard let instant = ISO8601DateFormatter().date(from: iso8601String) else {
+        return nil
+    }
+
     let timeZone = TimeZone(identifier: tzId ?? TimeZone.current.identifier)
-    let zonedDateTime = instant ?? Date()
-    
     let formatter = DateFormatter()
-    formatter.dateFormat = "hh:mm a"
     formatter.timeZone = timeZone
-    
-    let timeString = formatter.string(from: zonedDateTime)
-    return timeString
+    formatter.dateFormat = format.dateFormat
+
+    // Match the locale handling used by timestampMillisecondsToDate so the
+    // string and numeric value paths format 12h/24h identically.
+    if format.rawValue.contains("HH") {
+        formatter.locale = Locale(identifier: "en_GB")
+    } else if format == .empty || format.rawValue.contains("hh") {
+        formatter.locale = Locale(identifier: "en_US_POSIX")
+    }
+
+    return formatter.string(from: instant)
 }
 
 /// Converts a timestamp value in milliseconds to a formatted date string.
--- a/Sources/JoyfillModel/ValueUnion.swift
+++ b/Sources/JoyfillModel/ValueUnion.swift
@@ -470,8 +470,7 @@ public extension ValueUnion {
         switch self {
         case .string(let string):
             guard !string.isEmpty else { return nil }
-            let date = getTimeFromISO8601Format(iso8601String: string, tzId: tzId)
-            return date
+            return getTimeFromISO8601Format(iso8601String: string, format: format, tzId: tzId)
         case .double(let integer):
             let date = timestampMillisecondsToDate(value: Int(integer), format: format, tzId: tzId)
             return date
--- a/Sources/JoyfillModel/ValueUnion.swift
+++ b/Sources/JoyfillModel/ValueUnion.swift
@@ -469,6 +469,7 @@ public extension ValueUnion {
     func dateTime(format: DateFormatType, tzId: String? = nil) -> String? {
         switch self {
         case .string(let string):
+            guard !string.isEmpty else { return nil }
             let date = getTimeFromISO8601Format(iso8601String: string, tzId: tzId)
             return date
         case .double(let integer):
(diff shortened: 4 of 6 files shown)


Sources/JoyfillModel/ValueUnion.swift
```diff
@@ -469,6 +469,7 @@ public extension ValueUnion {
     func dateTime(format: DateFormatType, tzId: String? = nil) -> String? {
         switch self {
         case .string(let string):
+            guard !string.isEmpty else { return nil }
```
vivekjyani, line 472: The fix is correct and minimal for the empty-string case — returning `nil` reuses the existing empty-state path and matches `.null` behavior. 👍

Non-blocking observation: the root cause lives one level down in `getTimeFromISO8601Format` (DocumentModel.swift:292), which does `instant ?? Date()`. So **any** unparseable non-empty string (e.g. `"garbage"`, a non-ISO date) still silently renders the *current* time rather than an empty state. The empty string is the realistic case, so scoping here is reasonable, but the underlying "fall back to now on parse failure" is a latent footgun. Worth a follow-up.

Also note: this `.string` branch calls `getTimeFromISO8601Format`, which always returns a **time-only** string (`"hh:mm a"`) and ignores the `format:` argument entirely — so a date field with a valid ISO8601 string value would render only the time. Pre-existing, out of scope here, but flagging since real values are stored as `.double` and this branch is rarely exercised.

JoyfillSwiftUIExample/JoyfillUITests/DateTimeField/DateTimeFieldTests.swift
```diff
@@ -1,6 +1,40 @@
 import XCTest
 
 final class DateTimeFieldTests: JoyfillUITestsBaseClass {
+    override func getJSONFileNameForTest() -> String {
```
vivekjyani, line 4: Heads up: adding this `getJSONFileNameForTest()` override at the class level changes the JSON for **every** test in `DateTimeFieldTests`, not just the new one. Previously these tests fell back to the base-class default `"Joydocjson"` (JoyfillUITestsBaseClass.swift:204); now `testDatePicker`, `testTimePicker`, and `testDateTimePicker` all run against `DateTimeFieldTestData` instead.

That is probably an improvement (scoping the date tests to date-specific data), and those three tests only look for a `ChangeDateIdentifier` button which still exists in the new file — so they should pass. But please confirm it is intentional and that the three pre-existing tests still pass against the new fixture, since this is an implicit behavior change beyond the empty-string fix.

JoyfillSwiftUIExample/JoyfillUITests/DateTimeField/DateTimeFieldTestData.json
```diff
@@ -281,7 +300,12 @@
           "_id": "66a14ced15a9dc96374e091e",
           "width": 816,
           "name": "Every Field Type",
-          "margin": 0
+          "margin": 0,
+          "deletable": true,
```
vivekjyani, line 304: These page-level `deletable: true` / `copyable: [...]` additions look unrelated to the empty-string date fix — scope creep in the test fixture. They feed `PageRowView.canDuplicate`/`isDeletable` (FormView.swift:544) and enable the page duplicate/delete affordances. No test in this file or `DateTimeFieldUITestCases.swift` asserts on page duplication/deletion, so this is low-risk, but unless a test actually needs it, please drop it to keep the diff focused on the fix. If it *is* needed for a follow-up, a one-line comment saying why would help.

JoyfillSwiftUIExample/JoyfillUITests/DateTimeField/DateTimeFieldTests.swift
vivekjyani: Intentional — the new testEmptyStringDateRendersEmptyState needs the "Test empty date string" fixture, and scoping the date tests to date-specific data is an improvement over the generic Joydocjson. Confirmed all three pre-existing tests still pass against DateTimeFieldTestData:

testDatePicker ✅ (11.1s)
testTimePicker ✅ (9.8s)
testDateTimePicker ✅ (7.1s)

They only depend on a ChangeDateIdentifier button, which the new fixture still provides. Thanks for flagging.

JoyfillSwiftUIExample/JoyfillUITests/DateTimeField/DateTimeFieldTestData.json
vivekjyani: These deletable/copyable flags aren't related to anything in these tests — nothing in DateTimeFieldTests or DateTimeFieldUITestCases reads or asserts on them, so they have no effect on the date tests or the fix. Leaving them as-is.

Sources/JoyfillModel/ValueUnion.swift
vivekjyani: Good call — addressed both points in this PR rather than leaving them as follow-ups. getTimeFromISO8601Format now returns nil on parse failure instead of instant ?? Date(), so any unparseable string (not just "") renders the empty state. It also takes a format: argument and formats per that format with the same en_GB/en_US_POSIX locale handling as timestampMillisecondsToDate, so the .string and .double paths are now consistent. All 4 DateTimeFieldTests pass. Thanks for the thorough review.

Sources/JoyfillModel/DocumentModel.swift
```diff
@@ -278,25 +278,35 @@ public enum DateFormatType: String {
...
 ///
-/// - Returns: A formatted time string in the format "hh:mm a".
-public func getTimeFromISO8601Format(iso8601String: String, tzId: String? = nil) -> String {
-    let dateFormatter = ISO8601DateFormatter()
-    let instant = dateFormatter.date(from: iso8601String)
-    
+/// - Returns: A formatted date/time string honoring `format`, or `nil` if the
+///   input is not a valid ISO8601 string. Returning `nil` (rather than falling
+///   back to the current date) lets callers render an empty state for invalid input.
+public func getTimeFromISO8601Format(iso8601String: String, format: DateFormatType = .empty, tzId: String? = nil) -> String? {
```
VishnuBishnoi, line 291: **Breaking public API change.** `getTimeFromISO8601Format` changes its return type from `String` to `String?` and gains a `format:` parameter. This is a `public` free function in `JoyfillModel`, so external SDK consumers that call it will need to handle the optional. Two asks:
- Confirm this break is intended for the current release line (looks fine given 3.0.0-rc, but it should be a conscious decision).
- Update the public docs / changelog as your PR description notes ("Docs should be updated").

Not asking you to preserve the old signature, just to make the break explicit + documented.

Sources/JoyfillModel/ValueUnion.swift
```diff
@@ -469,8 +469,8 @@ public extension ValueUnion {
     func dateTime(format: DateFormatType, tzId: String? = nil) -> String? {
         switch self {
         case .string(let string):
-            let date = getTimeFromISO8601Format(iso8601String: string, tzId: tzId)
-            return date
+            guard !string.isEmpty else { return nil }
```
VishnuBishnoi, line 472: Good defensive guard. One thing to call out beyond the empty-state fix: for a **valid** ISO8601 *string* value, the old code always emitted `hh:mm a` (time only); now it formats using the field's actual `format`. That's an improvement (the string path now matches the numeric/timestamp path), but it's a behavior change broader than the PR title. Worth a quick confirm that no consumer relied on the old time-only output for string-valued date fields.

JoyfillSwiftUIExample/JoyfillUITests/DateTimeField/DateTimeFieldTestData.json
VishnuBishnoi, line 304: This page-level `deletable: true` / `copyable: [...]` addition looks unrelated to the empty-date-string fix, and `DateTimeFieldTestData.json` is a **shared** fixture also used by `DateTimeFieldUITestCases`. Enabling page copy/delete here changes the page's capabilities for every test using this file. Please remove it unless a specific test needs it — and if one does, call out which.

JoyfillSwiftUIExample/JoyfillUITests/DateTimeField/DateTimeFieldTests.swift
VishnuBishnoi, line 4: Adding this override repoints the whole `DateTimeFieldTests` class (including the pre-existing `testDatePicker` / `testTimePicker` / `testDateTimePicker`) from the default `Joydocjson` to `DateTimeFieldTestData`. That's reasonable, but since the UI Tests CI job is skipped these aren't verified automatically — please confirm all three still pass against the new fixture (they look generic enough to, but worth checking the time/date-time picker assertions given the fields in this JSON).

VishnuBishnoi (changes requested):
**Requesting changes — the core fix is correct, but a few items should be resolved before merge.**

The root-cause analysis is right and I confirmed it end-to-end: an empty-string date value flows through `fieldData?.value?.normalizedNumericValue` → stays `.string("")` (since `Double("")` is nil) → `ValueUnion.dateTime` → old `getTimeFromISO8601Format` failed ISO parse → `instant ?? Date()` fabricated the current time. Returning `nil` for invalid input + guarding the empty string fixes this, and the new unit tests cover it well. Locale-by-format handling correctly mirrors `timestampMillisecondsToDate`. Unit Tests pass in CI.

The blocking/asks are about scope and verification, not the fix logic:

1. **Unrelated change to a shared test fixture** — adding page-level `deletable`/`copyable` to `DateTimeFieldTestData.json` (see inline). This file is shared with `DateTimeFieldUITestCases`. Please remove it unless it's actually required.
2. **The new JSON override repoints 3 pre-existing tests** (`testDatePicker`/`testTimePicker`/`testDateTimePicker`) from `Joydocjson` to `DateTimeFieldTestData`. Because the UI Tests CI job is **skipped**, this isn't verified by CI — please confirm those still pass locally.
3. **Breaking public API** — heads-up + docs ask (inline).

Happy to re-review and approve once these are addressed; the underlying fix is good.

Sources/JoyfillModel/DocumentModel.swift
vivekjyani: Confirmed — this break is intentional.

The -> String? change is the fix itself: the old version fell back to Date() (today's date) on invalid/empty input, which is what caused the date field to render the current time for an empty string. Returning nil lets callers render a proper empty state instead. The new format: parameter is additive (defaults to .empty), so it's not source-breaking on its own — only the optional return type is.

Good call on the docs — I'll add a CHANGELOG entry flagging the breaking signature change (return type String → String?) so SDK consumers know to unwrap. Will push that with this PR.

JoyfillSwiftUIExample/JoyfillUITests/DateTimeField/DateTimeFieldTestData.json
vivekjyani: This is just the test fixture JSON — the deletable/copyable page settings don't impact any of these tests. Every test here finds elements by specific identifier (ChangeDateIdentifier) or exact label, so nothing about page copy/delete changes what they assert. Leaving it as-is.

Sources/JoyfillModel/ValueUnion.swift
vivekjyani: The .string branch was already being passed format — it just ignored it (old code hardcoded hh:mm a) while the timestamp branch honored it. So this just makes the string path respect the format it's given, consistent with the numeric path. The old time-only output contradicted the requested format, so no consumer should've been relying on it. Intended fix, no change needed.

JoyfillSwiftUIExample/JoyfillUITests/DateTimeField/DateTimeFieldTests.swift
vivekjyani: Confirmed — I ran testDatePicker, testTimePicker, and testDateTimePicker locally against DateTimeFieldTestData and all three pass. They're generic (tap a ChangeDateIdentifier button, assert the picker appears), and the new fixture has plenty of date/time fields, so they hold up fine on the new JSON. Good to go.

JoyfillSwiftUIExample/JoyfillUITests/DateTimeField/DateTimeFieldTestData.json
VishnuBishnoi: Acknowledged — agreed it's harmless given every test here keys off specific identifiers/labels and the new field is appended last, so no index/order shifts. Not blocking on it. The only reason to still prefer removing it is fixture hygiene (keeping a shared JSON minimal so a future page-copy/delete test reading this file doesn't inherit surprise flags), but your call. Resolving this thread.

Sources/JoyfillModel/DocumentModel.swift
VishnuBishnoi, line 291: Re your reply: 👍 on confirming the break is intentional. Just leaving a visible reminder to include the CHANGELOG entry (return type `String` → `String?`) in this PR as you mentioned — the diff doesn't include it yet. Not blocking the approval, but it should land before merge so consumers know to unwrap.

VishnuBishnoi (approved):
**Approving — re-review after your responses. Thanks for the thorough replies.**

My earlier changes-requested items are resolved or reasonably addressed:

1. **Breaking public API (`getTimeFromISO8601Format` → `String?`)** — confirmed intentional, and you're right that the `String?` return *is* the fix (the old `instant ?? Date()` fallback was the bug) while the `format:` param is additive. Please do land the CHANGELOG entry you mentioned before merge — the repo has a `CHANGELOG.md` and the PR's diff doesn't touch it yet, so SDK consumers won't otherwise know the return type went optional.
2. **String-path formatting change** — agreed: the `.string` branch was already receiving `format` and just ignoring it, so honoring it makes the string and timestamp paths consistent. Sound.
3. **`DateTimeFieldTests` JSON override** — good that you ran `testDatePicker` / `testTimePicker` / `testDateTimePicker` against the new fixture locally and confirmed they pass; that was my main worry given the UI Tests CI job is skipped.
4. **`deletable`/`copyable` in the shared fixture** — accepting as non-blocking (replied in-thread). Harmless to the current suites; only a fixture-hygiene preference.

Verified the fix survived the `main` merge intact (`getTimeFromISO8601Format` returns `nil` on parse failure; `ValueUnion.dateTime` guards the empty string). Core logic is correct, the unit tests in `DateFormattingTests` cover empty/garbage/null/valid/int/double cases, and CI is green (Build + Unit Tests + Syntax). PR is mergeable.

Non-blocking reminder inline about the CHANGELOG.

PR #338, opened 2026-06-11
NO-2200: Keep table columns horizontally scrollable when a filter returns zero rows

vivekjyani:
## Context
When a table field has an active filter that matches **zero rows**, the body collapses to nothing. Previously this left the horizontal scroll position stuck on the first columns — the user could no longer scroll right to see the rest of the columns/headers while the empty filter was applied.

## What changed
- **`Sources/JoyfillUI/View/Fields/TableView/TableModalView.swift`** — in both the iOS 16+ and the pre-iOS 16 branches of `table`, when `filteredcellModels.isEmpty` we now render a transparent `Color.clear.frame(width: tableContentWidth)` spacer inside the scroll content. This preserves the full content width (and a scrollable body) even with no rows.
- **`Sources/JoyfillUI/View/Fields/TableView/TableViewModel.swift`** — added a computed `tableContentWidth` (row-decorator width + column count × 200) used by the spacer to size the empty content to the real table width.
- **`JoyfillSwiftUIExample/JoyfillUITests/TableField/TableNumber_Block_DateFieldTest.swift`** — added `testHeaderRemainsHorizontallyScrollableWhenFilterReturnsZeroRows`.

## Decisions / why
- The spacer is `Color.clear` with **no height constraint** (rather than `height: 0`). A zero-height spacer keeps the width but leaves the empty body un-scrollable — the swipe gesture lands below the 0-height strip in dead space. Letting `Color.clear` expand to fill the available height makes the whole empty area respond to the horizontal swipe, which is the actual fix. `.fixedSize(horizontal: false, vertical: true)` + the existing `minHeight: geometry.size.height` frame keep it from over-expanding.
- Column width is `200` to match the fixed per-column width used elsewhere in the table layout.

## Screenshot / video
Not attached — this is a gesture/scroll behavior in the empty-filter state that I couldn't capture from CI. Will add a short screen recording (filter to zero rows → swipe → columns scroll) before merge if reviewers want it.

## Tests
- Added a UI test in the file whose `TableNewColumns` JSON has many columns (so the table is wider than the screen): filter the Number Column to zero rows, assert no cells render, then swipe `TableScrollView` and assert the column header's x position moves — proving the header/columns stay horizontally scrollable.
- The test exercises both the empty-filter rendering and the scroll behavior. Note: it's a simulator UI test and hasn't been run locally in this change; happy to run it on a sim before merge.

## Public API / docs
No public API change — `tableContentWidth` is internal to `TableViewModel`. No documentation updates required.

## Notes for reviewers
- The same empty-state block is intentionally duplicated across the iOS 16+ and pre-iOS 16 branches to match the existing structure of `table`; no shared helper was introduced to keep the diff minimal.

### DIFF
--- a/Sources/JoyfillUI/View/Fields/TableView/TableModalView.swift
+++ b/Sources/JoyfillUI/View/Fields/TableView/TableModalView.swift
@@ -410,7 +410,7 @@ struct TableModalView : View {
                             if viewModel.tableDataModel.filteredcellModels.isEmpty {
                                 // Keeps all columns horizontally scrollable when a filter matches zero rows.
                                 Color.clear
-                                    .frame(width: viewModel.tableContentWidth, height: 0)
+                                    .frame(width: viewModel.tableContentWidth)
                             }
                         }
                         .fixedSize(horizontal: false, vertical: true)
@@ -459,7 +459,7 @@ struct TableModalView : View {
                             if viewModel.tableDataModel.filteredcellModels.isEmpty {
                                 // Keeps all columns horizontally scrollable when a filter matches zero rows.
                                 Color.clear
-                                    .frame(width: viewModel.tableContentWidth, height: 0)
+                                    .frame(width: viewModel.tableContentWidth)
                             }
                         }
                         .fixedSize(horizontal: false, vertical: true)
--- a/Sources/JoyfillUI/View/Fields/TableView/TableModalView.swift
+++ b/Sources/JoyfillUI/View/Fields/TableView/TableModalView.swift
@@ -29,7 +29,7 @@ struct TableRowView : View {
             ForEach($rowDataModel.cells, id: \.id) { $cellModel in
                 let showRequired = viewModel.tableDataModel.requiredColumnIDs.contains(cellModel.data.id) && !cellModel.data.isCellFilled
                 TableViewCellBuilder(viewModel: viewModel, cellModel: $cellModel)
-                    .frame(width: 200, height: 60)
+                    .frame(width: Utility.singleColumnWidth, height: 60)
                     .background(Color.rowSelectionBackground(isSelected: isSelected, colorScheme: colorScheme))
                     .overlay(
                         RoundedRectangle(cornerRadius: 0)
@@ -334,7 +334,7 @@ struct TableModalView : View {
                 }
                 .padding(.all, 4)
                 .font(.system(size: 15))
-                .frame(width: 200, height: 60)
+                .frame(width: Utility.singleColumnWidth, height: 60)
                 .background(colorScheme == .dark ? Color(UIColor.systemGray6) : Color.tableColumnBgColor)
                 .overlay(
                     RoundedRectangle(cornerRadius: 0)
--- a/Sources/JoyfillUI/View/Fields/TableView/TableModalView.swift
+++ b/Sources/JoyfillUI/View/Fields/TableView/TableModalView.swift
@@ -407,6 +407,11 @@ struct TableModalView : View {
                                 TableRowView(viewModel: viewModel, rowDataModel: $rowCellModels, isSelected: isRowSelected)
                                     .frame(height: 60)
                             }
+                            if viewModel.tableDataModel.filteredcellModels.isEmpty {
+                                // Keeps all columns horizontally scrollable when a filter matches zero rows.
+                                Color.clear
+                                    .frame(width: viewModel.tableContentWidth, height: 0)
+                            }
                         }
                         .fixedSize(horizontal: false, vertical: true)
                         .frame(minWidth: geometry.size.width, minHeight: geometry.size.height, alignment: .topLeading)
@@ -450,6 +455,11 @@ struct TableModalView : View {
                                 let isRowSelected = viewModel.tableDataModel.selectedRows.contains(rowCellModels.rowID)
                                 TableRowView(viewModel: viewModel, rowDataModel: $rowCellModels, isSelected: isRowSelected)
                                     .frame(height: 60)
+                            }
+                            if viewModel.tableDataModel.filteredcellModels.isEmpty {
+                                // Keeps all columns horizontally scrollable when a filter matches zero rows.
+                                Color.clear
+                                    .frame(width: viewModel.tableContentWidth, height: 0)
                             }
                         }
                         .fixedSize(horizontal: false, vertical: true)
--- a/Sources/JoyfillUI/View/Fields/TableView/TableModalView.swift
+++ b/Sources/JoyfillUI/View/Fields/TableView/TableModalView.swift
@@ -411,6 +411,7 @@ struct TableModalView : View {
                                 // Keeps all columns horizontally scrollable when a filter matches zero rows.
                                 Color.clear
                                     .frame(width: viewModel.tableContentWidth)
+                                    .frame(minHeight: geometry.size.height)
                             }
                         }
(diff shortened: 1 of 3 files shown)


Sources/JoyfillUI/View/Fields/TableView/TableViewModel.swift
```diff
@@ -27,6 +27,11 @@ class TableViewModel: ObservableObject, TableDataViewModelProtocol {
         return tableDataModel.decorate
     }
 
+    var tableContentWidth: CGFloat {
+        let decoratorsWidth = showRowDecorators ? decoratorsCellWidth() : 0
+        return decoratorsWidth + CGFloat(tableDataModel.tableColumns.count) * 200
```
vivekjyani: Non-blocking: `200` is the per-column width and is now hardcoded in three places — here, `TableRowView` (`TableViewCellBuilder.frame(width: 200, height: 60)`) and `colsHeader` (`.frame(width: 200, height: 60)`). If these ever drift, `tableContentWidth` would silently mis-size the spacer and the header/body scroll ranges would desync. Consider a single shared constant (e.g. `static let columnWidth: CGFloat = 200`) referenced by all three.

Sources/JoyfillUI/View/Fields/TableView/TableModalView.swift
```diff
@@ -407,6 +407,11 @@ struct TableModalView : View {
                                 TableRowView(viewModel: viewModel, rowDataModel: $rowCellModels, isSelected: isRowSelected)
                                     .frame(height: 60)
                             }
+                            if viewModel.tableDataModel.filteredcellModels.isEmpty {
+                                // Keeps all columns horizontally scrollable when a filter matches zero rows.
+                                Color.clear
```
vivekjyani: The whole fix hinges on `Color.clear` expanding to fill the body height so the empty area catches the horizontal swipe (per the PR description). That height is emergent — it comes from the parent's `minHeight: geometry.size.height` frame interacting with `.fixedSize(horizontal: false, vertical: true)`, with no height set here. That's subtle and easy to regress if someone later touches the `.fixedSize`/`.frame` modifiers. Consider making the intent explicit, e.g. `Color.clear.frame(width: viewModel.tableContentWidth, minHeight: geometry.size.height)` (or `maxHeight: .infinity`), so the hit-area requirement is local and self-documenting rather than relying on the container's layout.

vivekjyani (commented):
Reviewed the full diff and surrounding layout (`scrollArea` / `colsHeader` / `table` offset-sync mechanism). The fix is correct in approach: the header mirrors the body's scroll via `.offset(x: offset.x)` driven by `ViewOffsetKey`, so restoring the body's content width when the filter empties it is the right lever. Two non-blocking notes inline, plus a verification ask: the UI test is the only guard on the core mechanism (Color.clear filling height to catch the swipe) and per the PR description hasn't been run on a sim yet — please run it before merge. No public API change. Otherwise good to merge.

Sources/JoyfillUI/View/Fields/TableView/TableViewModel.swift
vivekjyani: Good catch. There was already a Utility.singleColumnWidth: CGFloat = 200 constant, so I pointed all three usages (tableContentWidth, TableViewCellBuilder frame, and colsHeader frame) at it rather than adding a new one. Single source of truth now, no risk of drift.

Sources/JoyfillUI/View/Fields/TableView/TableModalView.swift
vivekjyani: Made it explicit. The empty-state spacer now chains .frame(width: tableContentWidth).frame(minHeight: geometry.size.height) so the fill-to-body-height (and the swipe hit-area) is stated locally rather than emerging from the container's .fixedSize/minHeight. Used two chained frame modifiers since SwiftUI has no frame(width:minHeight:) overload. Applied to both the iOS 16+ and pre-iOS 16 branches.

Sources/JoyfillUI/View/Fields/TableView/TableViewModel.swift
```diff
@@ -27,6 +27,11 @@ class TableViewModel: ObservableObject, TableDataViewModelProtocol {
         return tableDataModel.decorate
     }
 
+    var tableContentWidth: CGFloat {
```
VishnuBishnoi, line 30: `tableContentWidth` uses `tableDataModel.tableColumns.count`. For a plain table that's correct since only rows are filtered, not columns. Just flagging the assumption explicitly — if column hiding/visibility is ever introduced this would need to use the visible-column count.

Sources/JoyfillUI/View/Fields/TableView/TableModalView.swift
```diff
@@ -407,6 +407,12 @@ struct TableModalView : View {
                                 TableRowView(viewModel: viewModel, rowDataModel: $rowCellModels, isSelected: isRowSelected)
                                     .frame(height: 60)
                             }
+                            if viewModel.tableDataModel.filteredcellModels.isEmpty {
```
VishnuBishnoi: The empty-state spacer block is duplicated verbatim across the iOS 16+ and pre-iOS 16 branches (here and again at ~460). Matching the existing structure is fine, but since both branches are now identical for this block, a tiny `@ViewBuilder` helper (e.g. `emptyFilterSpacer(geometry:)`) would remove the duplication and keep the two branches from drifting. Non-blocking.

Also: does the **collection** table modal have the same zero-row horizontal-scroll issue? If so, worth a follow-up ticket since this PR only covers `TableModalView`.

VishnuBishnoi (approved):
**Approving.**

The fix correctly targets the real mechanism. The column header (`colsHeader`) lives in a `.scrollDisabled(true)` horizontal `ScrollView` whose `.offset(x: offset.x)` is driven by the body's scroll position via the `ViewOffsetKey` preference. When a filter matches zero rows the body's content width collapses, so `offset.x` can no longer change and the header is stuck. Adding a full-width `Color.clear` spacer restores the scrollable content extent, which is exactly what un-sticks the header. `tableContentWidth` (decorators + `columns.count * singleColumnWidth`) matches the real layout, so the scroll range stays aligned with the header.

Nice incidental cleanup replacing the magic `200` with `Utility.singleColumnWidth`.

Non-blocking notes inline. CI Build + Unit Tests pass; note the new UI test is not exercised by CI (UI Tests job is skipped).

Sources/JoyfillUI/View/Fields/TableView/TableViewModel.swift
vivekjyani: Good flag. I tested it with column hiding and it behaves correctly — tableColumns already reflects only the columns that are actually rendered, so count stays in sync with the visible width and the spacer sizes correctly even when columns are hidden. So no desync today. I'll keep the assumption in mind if the column model ever changes to keep hidden columns in the same array.

Sources/JoyfillUI/View/Fields/TableView/TableModalView.swift
vivekjyani: Done — extracted the block into a private @ViewBuilder func emptyFilterSpacer(geometry:) and both branches now just call it, so there's a single copy and no drift. On the collection table modal: good call, I'll check it separately and open a follow-up ticket if it has the same zero-row scroll issue, since it's out of scope for this PR.

PR #340, opened 2026-06-11
NO-2202: Pin row form headers so navigation buttons stay fixed while scrolling

vivekjyani:
## Context
In the table and collection **row form** (the per-row edit sheet) and its **bulk-edit** variant, the header — back/next chevrons, `+` insert, close `X`, and the bulk "Apply All" button — was scrolling away together with the form fields. The header should remain fixed at the top so those actions are always reachable, no matter how far the user scrolls through the columns.

## What changed
- **`TableModalTopNavigationView.swift`** (`EditMultipleRowsSheetView`) and **`CollectionModalTopNavigationView.swift`** (`CollectionEditMultipleRowsSheetView`):
  - Wrapped the view in an outer `VStack` and moved the single-row navigation header **and** the title + "Apply All" bulk-edit header **out of** the `ScrollView`. Only the field list scrolls now; the headers are pinned.
- **New UI tests (4):**
  - `TableFieldUITestCases`: `testRowFormHeaderStaysPinnedWhileScrolling`, `testBulkEditRowFormHeaderStaysPinnedWhileScrolling`
  - `CollectionFieldTests`: `testRowFormHeaderStaysPinnedWhileScrolling`, `testBulkEditRowFormHeaderStaysPinnedWhileScrolling`
  - Each opens the form, records the header button's Y and a field's Y, scrolls the fields, then asserts the header's Y is unchanged (pinned), the field moved (content scrolled), and the buttons are still **functional** (Down navigates rows / Close dismisses / Apply All applies + dismisses).
- **Test helper fix:** `editSingleRow{Upper,Lower}Button()` / `editInsertRowPlusButton()` in `OnChangeHandlerUITests`, `CollectionFieldTests`, and `CollectionFieldSearchFilterTests` looked the buttons up via `app.scrollViews.otherElements.buttons[...]`. Since the buttons are no longer inside the `ScrollView`, those queries were updated to `app.buttons[...]`.

## Implementation notes / decisions
- The single-row header keeps `.padding(.bottom, 8)`; the title header uses a conditional top padding (`count == 1 ? 0 : 16`) so spacing stays correct whether or not the chevron row is shown above it.
- Headers were pulled out rather than using `pinnedViews`/sticky-section tricks to keep the structure simple and identical between the table and collection implementations.

## Public API
- None. This is an internal view-layout change; no public API surface changed and no docs need updating.

## Test coverage
- Covered by the 4 new UI tests above. All 4 pass locally (iPhone 17 simulator). The pre-existing nav-button helpers were updated so dependent tests keep passing after the layout change.

## Notes for reviewers
- Behavior is identical across the table and collection forms — the two source diffs mirror each other.
- A demo screen recording will be added as a comment.

### DIFF
--- a/Sources/JoyfillUI/View/Fields/CollectionView/CollectionModalTopNavigationView.swift
+++ b/Sources/JoyfillUI/View/Fields/CollectionView/CollectionModalTopNavigationView.swift
@@ -323,9 +323,7 @@ struct CollectionEditMultipleRowsSheetView: View {
     }
 
     var body: some View {
-        ScrollViewReader { scrollProxy in
-        ScrollView {
-            VStack(alignment: .leading, spacing: 16) {
+        VStack(alignment: .leading, spacing: 0) {
                 if viewModel.tableDataModel.selectedRows.count == 1 {
                     HStack(alignment: .top) {
                         if !viewModel.tableDataModel.navigationIntent.rowFormOpenedViaGoto {
@@ -398,6 +396,8 @@ struct CollectionEditMultipleRowsSheetView: View {
                         })
                         .accessibilityIdentifier("DismissEditSingleRowSheetButtonIdentifier")
                     }
+                    .padding([.horizontal, .top], 16)
+                    .padding(.bottom, 8)
                 }
                 
                 HStack(alignment: .top) {
@@ -461,7 +461,11 @@ struct CollectionEditMultipleRowsSheetView: View {
                         })
                     }
                 }
-
+                .padding(.horizontal, 16)
+                .padding(.top, viewModel.tableDataModel.selectedRows.count == 1 ? 0 : 16)
+            ScrollViewReader { scrollProxy in
+            ScrollView {
+            VStack(alignment: .leading, spacing: 16) {
                 if !viewModel.tableDataModel.selectedRows.isEmpty {
                     selectedRowsHeaderColumns
                 }
@@ -485,6 +489,7 @@ struct CollectionEditMultipleRowsSheetView: View {
         }))
         .onTapGesture {
             viewModel.tableDataModel.navigationIntent.focusColumnId = nil
+        }
         }
         }
         .safeAreaInset(edge: .bottom) {
--- a/Sources/JoyfillUI/View/Fields/TableView/TableModalTopNavigationView.swift
+++ b/Sources/JoyfillUI/View/Fields/TableView/TableModalTopNavigationView.swift
@@ -257,9 +257,7 @@ struct EditMultipleRowsSheetView: View {
     }
 
     var body: some View {
-        ScrollViewReader { scrollProxy in
-        ScrollView {
-            VStack(alignment: .leading, spacing: 16) {
+        VStack(alignment: .leading, spacing: 0) {
                     if viewModel.tableDataModel.selectedRows.count == 1 {
                         HStack(alignment: .top) {
                             if !viewModel.tableDataModel.navigationIntent.rowFormOpenedViaGoto {
@@ -332,7 +330,12 @@ struct EditMultipleRowsSheetView: View {
                             })
                             .accessibilityIdentifier("DismissEditSingleRowSheetButtonIdentifier")
                         }
+                        .padding([.horizontal, .top], 16)
+                        .padding(.bottom, 8)
                     }
+            ScrollViewReader { scrollProxy in
+            ScrollView {
+            VStack(alignment: .leading, spacing: 16) {
                 HStack(alignment: .top) {
                     if let title = viewModel.tableDataModel.title {
                         VStack(alignment: .leading) {
@@ -635,6 +638,7 @@ struct EditMultipleRowsSheetView: View {
         }))
         .onTapGesture {
             viewModel.tableDataModel.navigationIntent.focusColumnId = nil
+        }
         }
         }
         .safeAreaInset(edge: .bottom) {
--- a/Sources/JoyfillUI/View/Fields/TableView/TableModalTopNavigationView.swift
+++ b/Sources/JoyfillUI/View/Fields/TableView/TableModalTopNavigationView.swift
@@ -333,9 +333,6 @@ struct EditMultipleRowsSheetView: View {
                         .padding([.horizontal, .top], 16)
                         .padding(.bottom, 8)
                     }
-            ScrollViewReader { scrollProxy in
-            ScrollView {
-            VStack(alignment: .leading, spacing: 16) {
                 HStack(alignment: .top) {
                     if let title = viewModel.tableDataModel.title {
                         VStack(alignment: .leading) {
@@ -392,7 +389,11 @@ struct EditMultipleRowsSheetView: View {
                         })
                     }
                 }
-
+                .padding(.horizontal, 16)
+                .padding(.top, viewModel.tableDataModel.selectedRows.count == 1 ? 0 : 16)
+            ScrollViewReader { scrollProxy in
+            ScrollView {
+            VStack(alignment: .leading, spacing: 16) {
                 ForEach(Array(viewModel.tableDataModel.tableColumns.enumerated()), id: \.offset) { colIndex, col in
                     let isFocused = col.id == viewModel.tableDataModel.navigationIntent.focusColumnId
                     VStack(alignment: .leading, spacing: 16) {
--- a/JoyfillSwiftUIExample/JoyfillUITests/ChangeUITests/OnChangeHandlerUITests.swift
+++ b/JoyfillSwiftUIExample/JoyfillUITests/ChangeUITests/OnChangeHandlerUITests.swift
@@ -131,15 +131,15 @@ final class OnChangeHandlerUITests: JoyfillUITestsBaseClass {
     }
     
     func editSingleRowUpperButton() -> XCUIElement {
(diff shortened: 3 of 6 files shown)


vivekjyani (commented):
Layout change is correct and the new tests meaningfully exercise the pinned-header behavior. A few non-blocking observations left inline. No public API changed; both `EditMultipleRowsSheetView` and `CollectionEditMultipleRowsSheetView` are internal.

Sources/JoyfillUI/View/Fields/TableView/TableModalTopNavigationView.swift
```diff
@@ -389,7 +389,11 @@ struct EditMultipleRowsSheetView: View {
                         })
                     }
                 }
-
+                .padding(.horizontal, 16)
+                .padding(.top, viewModel.tableDataModel.selectedRows.count == 1 ? 0 : 16)
```
vivekjyani, line 393: Non-blocking behavior change: the keyboard-dismiss / focus-clear gestures (`.simultaneousGesture(DragGesture…)` and `.onTapGesture`, lines 636-642) are attached to the inner `ScrollView`. Now that the nav + title headers are pinned *outside* the `ScrollView`, dragging or tapping on the header area no longer dismisses the keyboard or clears `focusColumnId` — previously the header was part of the scroll content so it was covered. Likely negligible (the header is mostly buttons), but worth confirming it's acceptable.

Also minor: in single-row mode the gap between the chevron row and the title row changes from 16 (the old `VStack(spacing: 16)`) to 8 (header `.padding(.bottom, 8)` + title `.padding(.top, 0)`). Bulk mode stays pixel-identical.

Sources/JoyfillUI/View/Fields/TableView/TableModalTopNavigationView.swift
```diff
@@ -257,9 +257,7 @@ struct EditMultipleRowsSheetView: View {
     }
 
     var body: some View {
-        ScrollViewReader { scrollProxy in
-        ScrollView {
-            VStack(alignment: .leading, spacing: 16) {
+        VStack(alignment: .leading, spacing: 0) {
```
vivekjyani, line 260: Readability: the inner header/field blocks kept their original deep indentation under the new shallow `VStack(spacing: 0)`, so the nesting now reads misleadingly (e.g. the `if` at 261 is indented as if still two containers deep). Compiles fine, but a re-indent of this body would make the new structure much easier to follow.

Sources/JoyfillUI/View/Fields/CollectionView/CollectionModalTopNavigationView.swift
```diff
@@ -461,7 +461,11 @@ struct CollectionEditMultipleRowsSheetView: View {
                         })
                     }
                 }
-
+                .padding(.horizontal, 16)
+                .padding(.top, viewModel.tableDataModel.selectedRows.count == 1 ? 0 : 16)
```
vivekjyani, line 465: Same non-blocking note as the table view: the keyboard-dismiss `.simultaneousGesture` / `.onTapGesture` (lines 486-492) are on the inner `ScrollView`, so the now-pinned headers are no longer covered by those gestures. And in single-row mode the chevron-row→title gap shrinks from 16 to 8. Mirrors the table change, as intended.

Sources/JoyfillUI/View/Fields/CollectionView/CollectionModalTopNavigationView.swift
```diff
@@ -323,9 +323,7 @@ struct CollectionEditMultipleRowsSheetView: View {
     }
 
     var body: some View {
-        ScrollViewReader { scrollProxy in
-        ScrollView {
-            VStack(alignment: .leading, spacing: 16) {
+        VStack(alignment: .leading, spacing: 0) {
```
vivekjyani, line 326: Readability: same as the table file — the moved-out header/field blocks retain their old indentation depth under the new `VStack(spacing: 0)`, making the structure harder to read. A re-indent would help.

Sources/JoyfillUI/View/Fields/TableView/TableModalTopNavigationView.swift
vivekjyani: Intentional for this PR — keeping the diff a clean structural move (headers lifted out of the ScrollView) instead of mixing in a 100+ line whitespace reflow that'd bury the behavioral change. Will do the re-indent as a separate whitespace-only follow-up.

Sources/JoyfillUI/View/Fields/CollectionView/CollectionModalTopNavigationView.swift
vivekjyani: Intentional for this PR — keeping the diff a clean structural move (headers lifted out of the ScrollView) instead of mixing in a 100+ line whitespace reflow that'd bury the behavioral change. Will do the re-indent as a separate whitespace-only follow-up.

Sources/JoyfillUI/View/Fields/TableView/TableModalTopNavigationView.swift
vivekjyani: Confirmed this is safe to leave as-is. The pinned header VStack contains only the nav/title controls (chevrons, plus, close, title, Apply All) — there's no editable text field up there. The keyboard only ever appears when editing a cell, and every editable cell still lives inside the ScrollView, which keeps the .simultaneousGesture/.onTapGesture. So no real dismiss/focus-clear behavior is lost by the pinning — the header is buttons-only, matching the "negligible" note. Leaving the gestures on the ScrollView.

Sources/JoyfillUI/View/Fields/CollectionView/CollectionModalTopNavigationView.swift
vivekjyani: Confirmed this is safe to leave as-is. The pinned header VStack contains only the nav/title controls (chevrons, plus, close, title, Apply All) — there's no editable text field up there. The keyboard only ever appears when editing a cell, and every editable cell still lives inside the ScrollView, which keeps the .simultaneousGesture/.onTapGesture. So no real dismiss/focus-clear behavior is lost by the pinning — the header is buttons-only, matching the "negligible" note. Leaving the gestures on the ScrollView.

Sources/JoyfillUI/View/Fields/TableView/TableModalTopNavigationView.swift
VishnuBishnoi, line 393: The conditional top padding (`count == 1 ? 0 : 16`) pairs with the single-row header's `.padding(.bottom, 8)` to keep spacing correct whether or not the chevron row is shown — good attention to detail.

Style nit only (non-blocking): the inner `ScrollViewReader { ScrollView { VStack {` chain stays at the same indentation level as the outer `VStack`, which makes the nesting hard to follow. Since you're already restructuring this view, re-indenting the moved block would help future readers. Mirrors the existing style though, so fine to leave.

Sources/JoyfillUI/View/Fields/CollectionView/CollectionModalTopNavigationView.swift
VishnuBishnoi, line 326: Collection implementation mirrors the table one exactly — pinned single-row + bulk headers, inner `ScrollViewReader`/`ScrollView` wrapping only the fields, `.safeAreaInset(.bottom)` still attached to the outer container. Consistent and correct.

VishnuBishnoi (approved):
**Approving.**

Sound layout change: lifting the single-row navigation header and the title/"Apply All" header out of the `ScrollView` and into an outer `VStack(spacing: 0)` pins them while only the field list scrolls. The `ScrollViewReader`/`scrollProxy` is correctly relocated to wrap just the inner `ScrollView`, and all its `scrollTo` targets (focused columns) remain inside that scope, so column focus-scrolling still works. The table and collection diffs mirror each other as claimed, and the test-helper updates (`app.scrollViews.otherElements.buttons[...]` → `app.buttons[...]`) are required because the nav buttons are no longer inside the scroll view.

CI Build + Syntax Check pass, so the brace restructure compiles cleanly. The new UI tests assert pinned-Y + still-functional buttons, which is the right thing to check — though note the UI Tests CI job is skipped, so they're only locally verified.

Minor note inline.

Sources/JoyfillUI/View/Fields/TableView/TableModalTopNavigationView.swift
vivekjyani: Thanks! On the indentation nit — same as the earlier readability note: leaving it as-is for this PR to keep the diff a clean structural move (headers lifted out of the ScrollView), and landing the re-indent as a separate whitespace-only follow-up so it doesn't bury the behavioral change.

Sources/JoyfillUI/View/Fields/CollectionView/CollectionModalTopNavigationView.swift
vivekjyani: Thanks for confirming — yes, the collection view intentionally mirrors the table one 1:1 (pinned single-row + bulk headers, inner ScrollViewReader/ScrollView around just the fields, .safeAreaInset(.bottom) on the outer container) to keep the two implementations consistent

PR #341, opened 2026-06-17
NO-2206: Live-refresh open row-edit form on Change-API row updates (table + collection)

vivekjyani:
## 1. Context / Spec
When a table/collection row's cell is updated programmatically via `DocumentEditor.change(changes:)` **while that row's edit form is open**, the open editor did not reflect the new value. Concretely: an app catches `onFocus` from a column decorator and fires a `field.value.rowUpdate` Change; the underlying model updates, but the open editor (e.g. a dropdown) stays stale on its old value ("Select Option"). Reported by a customer (ST).

A second, related symptom: even once the value showed in the grid, a subsequent row operation (Insert Below / move / delete) dropped it from the grid until the field was re-navigated.

Root cause (both): the row-edit sheet's cell editors snapshot their value into local SwiftUI `@State`, and the grid's displayed cell models can keep a stale identity, so an external model change isn't reflected until the view is forced to rebuild. The document model itself was always correct (hence re-navigation restored the value).

## 2. What changed (files / code)
**SDK — open row form refreshes on external Change-API updates** (internal views only):
- `TableModalTopNavigationView.swift` (`EditMultipleRowsSheetView`) and `CollectionModalTopNavigationView.swift` (`CollectionEditMultipleRowsSheetView`):
  - Added `refreshViewID()`; moved `.id(viewID)` from the `ScrollView` onto the **inner content `VStack`**; added `.onChange(of: viewModel.uuid) { refreshViewID() }`.
  - On an external Change-API update, `viewModel.uuid` bumps → the inner content re-identifies → editors re-init from the fresh model. `.id` on the content (not the `ScrollView`) preserves the scroll offset.

**SDK — grid value persists through row operations:**
- `TableViewModel.swift`: threaded `isBulkEdit` through `cellDidChange` and pass `true` from `updateUIModels`, so `updateCellModel` bumps `cells[colIndex].id = UUID()`. The new cell identity forces the grid cell to rebuild from fresh data, so a Change-API value survives `filterRowsIfNeeded()` (`filteredcellModels = cellModels`) on Insert Below / move / delete.

**Example app — repro / manual harness:**
- `DecoratorToRowUpdate/DecoratorRowUpdateDemoView.swift` + `DecoratorToRowUpdate.json`: a form with one table and one 3-level nested collection, every column carrying its own decorator; tapping one catches `onFocus` and pushes a `field.value.rowUpdate` for the tapped column on the selected row(s).
- `UserAccessTokenTextFieldView.swift`: "Decorator → Row Change" entry in `OptionSelectionView`; `project.pbxproj` registers the new file + JSON resource.

**Tests:**
- `JoyfillUITests/DecoratorUITests/DecoratorRowUpdateUITests.swift`: UI tests asserting a decorator tap → Change API updates both the open row form and the grid, for table and for collection root / level-1 / level-2 nested rows.

## 3. Implementation decisions
- **Central rebuild vs per-editor sync (sheet):** chose the single-spot `viewID` rebuild over editing all 8 cell editors. `viewModel.uuid` is bumped **only** by the `change(changes:)` path, never by user typing, so the refresh fires on external updates without interrupting an in-progress edit.
- **`.id` on inner content, not the `ScrollView`:** recreating the `ScrollView` resets scroll; re-identifying only its content preserves the offset. `col.id` is left intact so `scrollProxy.scrollTo(columnId)` keeps working.
- **Cell-id bump (grid):** the displayed rows are rebuilt from `cellModels` on row ops, and SwiftUI reuses a cell view whose identity is unchanged — so the master cell needed a fresh `id` (via `isBulkEdit: true`) to force a rebuild with the new value. Collection didn't need this: its `updateUIModels` already bumps the displayed cell id and its `filterRowsIfNeeded()` doesn't reset `filteredcellModels` from `cellModels` (verified).

## 4. Visuals
UI behavior change. **Before:** the open row's dropdown stays on "Select Option" after a Change-API `rowUpdate`, and a Change-API value disappears from the grid after Insert Below. **After:** the open editor updates live (scroll preserved) and the grid value persists through row ops.

_Screen recording from the "Decorator → Row Change" demo to be attached._

## 5. Public API / Docs
No public API changes — the edited types are internal, no `init` signatures changed, no data-model changes. No documentation update required.

## 6. Tests
- **Automated (UI tests)** — `DecoratorRowUpdateUITests`: `testTableDecoratorUpdatesRowFormAndGrid`, `testCollectionRootDecoratorUpdatesRowFormAndGrid`, `testCollectionLevel1DecoratorUpdatesRowFormAndGrid`, `testCollectionLevel2DecoratorUpdatesRowFormAndGrid`. Each taps a column decorator and asserts the open row form **and** the grid reflect the Change-API value. (UI tests are skipped in CI; run locally.)
- **Manual** — validated across all column types for table and collection (incl. nested), single-row and bulk edit: live refresh, preserved scroll, and value persistence through Insert Below.

## 7. Notes for reviewers
- Two focused SDK changes: (a) `.id(viewID)` placement + `refreshViewID()` in the two sheets; (b) the `isBulkEdit` threading in `TableViewModel.cellDidChange` / `updateUIModels`.
- `viewModel.uuid` changes only on the external `change()` path → typing is never interrupted.
- The grid fix is table-specific because only the table's `filterRowsIfNeeded()` does `filteredcellModels = cellModels`; collection was verified to already hold the value.
- Bulk-edit fields are intentionally empty (one value → many rows); the live-refresh visual is meaningful for single-row editing. The demo's bulk path still writes all selected rows.
- The `DecoratorToRowUpdate` example is test scaffolding (example app only), not shipping SDK code.

### DIFF
--- a/JoyfillSwiftUIExample/JoyfillUITests/DecoratorUITests/DecoratorRowUpdateUITests.swift
+++ b/JoyfillSwiftUIExample/JoyfillUITests/DecoratorUITests/DecoratorRowUpdateUITests.swift
@@ -1,0 +1,310 @@
+import XCTest
+
+// Table column decorator actions (from DecoratorToRowUpdate.json, tableColumnOrder)
+private let tableDropdownAction    = "Comment"   // dropdown   → "Yes"
+private let tableTextAction        = "Camera"    // text       → "Updated #1"
+private let tableMultiSelectAction = "Import"    // multiSelect → "Option 1"
+private let tableImageAction       = "Image"     // image      → 1 image added
+private let tableNumberAction      = "File"      // number     → 10
+private let tableDateAction        = "Download"  // date       → current epoch
+private let tableBlockAction       = "Claud"     // block      → display-only
+private let tableBarcodeAction     = "Filter"    // barcode    → "Updated #1"
+private let tableSignatureAction   = "Share"     // signature  → image URL
+
+// Collection root-level decorator actions (collectionSchemaId: text, dropdown, image)
+private let collRootTextAction     = "flag"      // text       → "Updated #1"
+private let collRootDropdownAction = "plus"      // dropdown   → "High"
+private let collRootImageAction    = "info"      // image      → 1 image added
+
+// Collection level-1 decorator actions (level1Table1: multiSelect, number, date)
+private let collL1MultiSelectAction = "eye"        // multiSelect → "Option 1"
+private let collL1NumberAction      = "magnet"     // number      → 10
+private let collL1DateAction        = "folderopen" // date        → current epoch
+
+// Collection level-2 decorator actions (level2Table1: signature, barcode, block)
+private let collL2SignatureAction = "paper"    // signature → image URL
+private let collL2BarcodeAction   = "filter"   // barcode   → "Updated #1"
+private let collL2BlockAction     = "rotate"   // block     → display-only
+
+// MARK: - DecoratorRowUpdateUITests
+
+final class DecoratorRowUpdateUITests: XCTestCase {
+
+    var app: XCUIApplication!
+
+    override func setUpWithError() throws {
+        continueAfterFailure = false
+
+        app = XCUIApplication()
+        app.launchArguments = ["--disable-animations"]
+        addUIInterruptionMonitor(withDescription: "System Alerts") { alert in
+            for label in ["Allow", "OK", "Continue", "Don't Allow"] {
+                if alert.buttons[label].exists { alert.buttons[label].tap(); return true }
+            }
+            return false
+        }
+        app.launch()
+        XCTAssertTrue(app.wait(for: .runningForeground, timeout: 15), "App did not launch")
+
+        // Navigate from OptionSelectionView → DecoratorRowUpdateDemoView
+        let card = app.staticTexts["Decorator → Row Change"]
+        if !card.waitForExistence(timeout: 5) {
+            app.swipeUp()
+            _ = card.waitForExistence(timeout: 3)
+        }
+        XCTAssertTrue(card.exists, "Decorator → Row Change card not found in OptionSelectionView")
+        card.tap()
+        spinRunloop(0.2)
+
+        let continueBtn = app.buttons["Continue"]
+        XCTAssertTrue(continueBtn.waitForExistence(timeout: 3), "Continue button not found")
+        continueBtn.tap()
+        spinRunloop(0.5)
+    }
+
+    override func tearDownWithError() throws {
+        app?.terminate()
+        app = nil
+    }
+
+    // MARK: - Table
+
+    /// Opens table row 1's edit form, taps every column decorator once,
+    /// verifies each field updates live, then dismisses and checks the grid.
+    func testTableDecoratorUpdatesRowFormAndGrid() {
+        typealias S = DecoratorUITestSupport
+
+        S.openTableDetailView(in: app)
+        S.openTableRowEditForm(rowIndex: 1, in: app)
+
+        // Dropdown → "Yes" (first option)
+        tapDecoratorAndAssertStaticText(action: tableDropdownAction, expected: "Yes")
+        // Text → "Updated #1" (TextField inside TableTextRowFormView)
+        tapDecoratorAndAssertTextField(action: tableTextAction, identifier: "EditRowsTextFieldIdentifier", expected: "Updated #1")
+        // MultiSelect → "Option 1" (first option shown as Text inside button)
+        tapDecoratorAndAssertStaticText(action: tableMultiSelectAction, expected: "Option 1")
+        // Image → cell receives one image (container exists)
+        tapDecoratorAndAssertExists(action: tableImageAction, identifier: "EditRowsImageFieldIdentifier")
+        // Number → "10" (TextField inside TableNumberView, identifier "TabelNumberFieldIdentifier")
+        tapDecoratorAndAssertTextField(action: tableNumberAction, identifier: "TabelNumberFieldIdentifier", expected: "10")
+        // Date → inner ChangeCellDateIdentifier button has a non-empty label
+        tapDecoratorAndAssertButtonHasValue(action: tableDateAction, identifier: "ChangeCellDateIdentifier")
+        // Block → display-only, no row-form editor
+        tapDecoratorDisplayOnly(action: tableBlockAction)
+        // Barcode → "Updated #1" (TextEditor inside TableBarcodeView, identifier "TableBarcodeFieldIdentifier")
+        tapDecoratorAndAssertTextView(action: tableBarcodeAction, identifier: "TableBarcodeFieldIdentifier", expected: "Updated #1")
+        // Signature → field container renders after URL is set
+        tapDecoratorAndAssertExists(action: tableSignatureAction, identifier: "EditRowsSignatureFieldIdentifier")
+
+        S.dismissRowEditForm(in: app)
+
+        assertGridTextView(identifier: "TabelTextFieldIdentifier", expected: "Updated #1")
+        assertGridTextField(identifier: "TabelNumberFieldIdentifier", expected: "10")
+        assertGridButtonContains(identifier: "TableDropdownIdentifier", text: "Yes")
+        assertGridButtonContains(identifier: "TableMultiSelectionFieldIdentifier", text: "Option 1")
+        assertGridButtonHasValue(identifier: "ChangeCellDateIdentifier")
+        assertGridButtonExists(identifier: "TableSignatureOpenSheetButton")
+    }
+
+    // MARK: - Collection root level
+
+    /// Opens collection root row 1's edit form, taps all 3 decorators (text, dropdown,
+    /// image), verifies live updates, then dismisses and checks the grid cells.
+    func testCollectionRootDecoratorUpdatesRowFormAndGrid() {
+        typealias S = DecoratorUITestSupport
+
+        S.openCollectionDetailView(in: app)
+        S.openCollectionRootRowEditForm(rowIndex: 1, in: app)
+
+        // Text → "Updated #1"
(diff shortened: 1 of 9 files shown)


Sources/JoyfillUI/View/Fields/TableView/TableViewModel.swift
```diff
@@ -390,11 +390,11 @@ class TableViewModel: ObservableObject, TableDataViewModelProtocol {
         return cellValues
     }
 
-    func cellDidChange(rowId: String, colIndex: Int, cellDataModel: CellDataModel, isNestedCell: Bool, callOnChange: Bool = true, metadata: Metadata? = nil) -> [ValueElement] {
+    func cellDidChange(rowId: String, colIndex: Int, cellDataModel: CellDataModel, isNestedCell: Bool, callOnChange: Bool = true, metadata: Metadata? = nil, isBulkEdit: Bool = false) -> [ValueElement] {
```
vivekjyani, line 393: `isBulkEdit` is being repurposed here as a generic "force a fresh cell identity" flag — the external Change-API `rowUpdate` path is a single-row update, not a bulk edit. The flag's only effect in `updateCellModel` / `updateCellModelForNested` (Models.swift:657-659 / 669-671) is `cells[colIndex].id = UUID()`. Threading it through works, but the name now misrepresents intent at the new call site. Consider renaming to something like `forceCellIdentityRefresh`, or at minimum adding a one-line comment at the `isBulkEdit: true` call in `updateUIModels`, so a future reader doesn't mistake this path for a bulk edit. Non-blocking.

Sources/JoyfillUI/View/Fields/TableView/TableViewModel.swift
```diff
@@ -641,7 +641,8 @@ extension TableViewModel: DocumentEditorDelegate {
                 cellDataModel: cell,
                 isNestedCell: false,
                 callOnChange: false,
-                metadata: row.metadata
+                metadata: row.metadata,
+                isBulkEdit: true
```
vivekjyani, line 645: This `isBulkEdit: true` (→ `cells[colIndex].id = UUID()`) is the actual mechanism behind the "value survives Insert Below" fix: it gives `cellModels` a fresh cell identity so the value persists when `filterRowsIfNeeded()` does `filteredcellModels = cellModels` (Models.swift:333-335, table-only). Right now that behavior is only covered by the UI tests, which the PR description notes are skipped in CI — so the fix has **no CI-enforced regression test**. Consider adding a fast model-level test to `TableViewModelDocumentEditorDelegateTests.swift` (it already exercises `applyRowEditChanges` + Insert Below): assert the target cell's `id` changes after `applyRowEditChanges`, and that the value+id survive a subsequent `filterRowsIfNeeded()`. Non-blocking, but worthwhile since the UI tests won't guard this in CI.

Sources/JoyfillUI/View/Fields/TableView/TableModalTopNavigationView.swift
```diff
@@ -623,15 +627,18 @@ struct EditMultipleRowsSheetView: View {
             }
             .padding(.all, 16)
             .environment(\.navigationFocusColumnId, viewModel.tableDataModel.navigationIntent.focusColumnId)
+            .id(viewID)
```
vivekjyani, line 630: Moving `.id(viewID)` from the `ScrollView` onto the inner content is the right call for preserving the scroll offset. One side effect worth confirming: `.onAppear` (line 632, which runs `scrollProxy.scrollTo(scrollToColumnId)`) is attached to the `ScrollView`, whose identity is now stable. Previously, with `.id(viewID)` on the `ScrollView`, every `viewID` bump (e.g. up/down row navigation via `selectUpperRow`/`selectBelowRow`) recreated the `ScrollView` and re-fired `.onAppear` → re-scrolled to `scrollToColumnId`. Now `.onAppear` fires only once. Is the scroll-to-focused-column behavior on row navigation / re-navigation still correct, or did it rely on that re-fire? (Same question for the collection sheet.) Just confirming this was checked — non-blocking.

JoyfillSwiftUIExample/JoyfillUITests/DecoratorUITests/DecoratorRowUpdateUITests.swift
```diff
@@ -0,0 +1,670 @@
...
+        tapDecorator(action: action)
+        XCTAssertTrue(
+            waitUntil(3) { self.app.staticTexts[expected].exists },
+            "'\(expected)' should appear as static text after tapping '\(action)' decorator"
+        )
+    }
+
+    /// Taps decorator, waits for a button with the given identifier to have a non-empty label.
+    /// Used for: date column (ChangeCellDateIdentifier shows the formatted date).
+    private func tapDecoratorAndAssertButtonHasValue(action: String, identifier: String) {
```
vivekjyani, line 531: This date assertion only checks `!button.label.isEmpty`. If the date cell renders a placeholder label (e.g. "Select a date") when unset, this passes even if the value never updated — so it wouldn't catch a date-column regression. If feasible, assert the formatted value, or at least that the label differs from its pre-tap value. Same applies to `assertGridButtonHasValue` / `assertRowFormButtonHasValue`. Minor.

vivekjyani (commented):
Automated code review (Claude). Two small SDK changes + example/test scaffolding. The approach is sound, no public API changes, and I found no blocking correctness bug. Left a few non-blocking inline notes — the main one is that the fix has no CI-enforced regression test (the UI tests are skipped in CI).

Sources/JoyfillUI/View/Fields/TableView/TableViewModel.swift
vivekjyani: Acknowledged. We're keeping the isBulkEdit name as-is for now to avoid a broader rename across multiple files. Adding a one-line comment at the isBulkEdit: true call site in updateUIModels to clarify intent for future readers ins

Sources/JoyfillUI/View/Fields/TableView/TableViewModel.swift
vivekjyani: Added testApplyRowEditChanges_CellIdIsBumpedAndValueSurvivesFilterRowsIfNeeded to TableViewModelDocumentEditorDelegateTests. It asserts three things: (1) the target cell's id is bumped after applyRowEditChanges (proving isBulkEdit: true fires), (2) the value is updated in cellModels, and (3) both the bumped id and value survive a subsequent filterRowsIfNeeded() call — which is the exact op Insert Below triggers. Runs in CI. Test passes.

Sources/JoyfillUI/View/Fields/TableView/TableModalTopNavigationView.swift
vivekjyani: Confirmed manually — scroll-to-focused-column on row navigation still works correctly after moving .id(viewID) from the ScrollView to the inner content. The .onAppear fires once on sheet open (initial scroll) and row navigation scroll is handled separately, unaffected by this change. No issue here.

JoyfillSwiftUIExample/JoyfillUITests/DecoratorUITests/DecoratorRowUpdateUITests.swift
vivekjyani: Good catch. Looking at TableDateView.swift, the unset date button renders with accessibilityLabel("") (empty string, no placeholder text), so !button.label.isEmpty was actually distinguishing "no date" from "has date" correctly. That said, the point about a non-empty placeholder is valid as a future-proofing concern. Fixed tapDecoratorAndAssertButtonHasValue to capture the label before the tap and assert !button.label.isEmpty && button.label != labelBefore — this catches both a stale empty label and a stale non-empty placeholder. assertGridButtonHasValue and assertRowFormButtonHasValue are called after the decorator tap has already confirmed the value was set; since the source confirms no placeholder text, the !isEmpty check there is sufficient.

Sources/JoyfillUI/View/Fields/TableView/TableViewModel.swift
vivekjyani: Verified on commit dcf1f597 — `testApplyRowEditChanges_CellIdIsBumpedAndValueSurvivesFilterRowsIfNeeded` pins the exact mechanism: Assert 1 catches a revert of `isBulkEdit: true` (id stops bumping), and Assert 3 proves the bumped id **and** value survive `filterRowsIfNeeded()` (the table-only `filteredcellModels = cellModels` that Insert Below triggers). Real regression test, runs in CI. This resolves the one thing I'd have pushed on. 👍

JoyfillSwiftUIExample/JoyfillUITests/DecoratorUITests/DecoratorRowUpdateUITests.swift
vivekjyani: Looks good (commit 44b1fb75) — capturing `labelBefore` and asserting `!isEmpty && label != labelBefore` covers both the stale-empty and stale-placeholder cases. Agreed the grid/row-form helpers are fine since they run after the decorator tap already confirmed the value.

Sources/JoyfillUI/View/Fields/TableView/TableModalTopNavigationView.swift
vivekjyani: Thanks for confirming — no concern then.

Sources/JoyfillUI/View/Fields/TableView/TableViewModel.swift
vivekjyani: Sounds good on keeping the name. One heads-up: the one-line clarifying comment at the `isBulkEdit: true` call in `updateUIModels` doesn't appear in the pushed changes — `TableViewModel.swift` is unchanged between e8411abf and the current head (44b1fb75), so it looks like it didn't get committed (your reply trailed off mid-sentence). Trivial, just flagging so it doesn't slip.

PR #343, opened 2026-06-19
NO-2216 Fix off-thread @Published read in Collection decoratorsCellWidth (crash)

vivekjyani:
## 1. Context / Spec
Selecting a value in a Collection field (e.g. NFPA collection → Severity) could freeze and then crash the app (`EXC_BAD_ACCESS` in Combine `PublishedSubject.value.getter`). Root cause: cell-model/width building runs on the background serial "Collection" queue, and a width helper was reading the `@Published var tableDataModel` off `self`, racing with SwiftUI's main-thread update cycle.

## 2. What Changed
`Sources/JoyfillUI/View/Fields/CollectionView/CollectionViewModel.swift`
- Added snapshot-based variants `decoratorConfig(for:)` and `decoratorsCellWidth(for:)` to `TableDataViewModelProtocol` that read the passed-in `tableDataModel` instead of `self.tableDataModel`.
- Routed the 3 background width call sites (`rowWidth` + the two `tableExpander` `getWidthForExpanderRow` calls) through `decoratorsCellWidth(for: tableDataModel)`.
- Parameterless versions are kept and now delegate to the `(for:)` variants, with a doc note that they are main-thread only.

## 3. Design Decisions
Followed the existing off-thread-crash campaign convention: never read `self.tableDataModel` / any `@Published` on the Collection queue — instead thread a main-thread snapshot in as a parameter. Keeping the parameterless accessors (delegating to the snapshot variants) avoids churn at the many main-thread SwiftUI View call sites while closing the background hole.

## 4. Screenshots / Video
N/A — crash/threading fix, no visual change.

## 5. Public API Impact
None. Changes are internal to `CollectionViewModel` / the internal `TableDataViewModelProtocol` extension. No docs update needed.

## 6. Test Coverage
No new automated tests — this is a data race that doesn't reproduce deterministically in a unit/UI test. Verified against the reported repro path. Existing Decorator UI tests remain green.

## 7. Reviewer Notes
This is one more leaf in the recurring "reading `@Published tableDataModel` off the Collection background queue" crash class (prior fixes: `ae7df42e`, `0a916fd5`, `7805a1f5`, `79e95b46`, `53de7a1b`, `56cab7f8`). The `decoratorsCellWidth()` → `decoratorConfig` → `self.tableDataModel` path was a hole an earlier pass missed even though `rowWidth(tableDataModel:)` already received a snapshot. Confirmed via Crashlytics `759fc897…` / Zendesk 111085.


Sources/JoyfillUI/View/Fields/CollectionView/CollectionViewModel.swift
```diff
@@ -974,7 +974,7 @@ class CollectionViewModel: ObservableObject, TableDataViewModelProtocol {
                                                             rowType: .tableExpander(schemaValue: (id, schemaValue),
                                                                                     level: level,
                                                                                     parentID: (columnID: "", rowID: rowDataModel.rowID),
-                                                                                    rowWidth: Utility.getWidthForExpanderRow(columns: filteredTableColumns, showSelector: showRowSelector(for: tableDataModel), showSingleClickEdit: showSingleClickEditButton(for: tableDataModel), showRowDecorators: tableDataModel.hasAnyRowDecorators(schemaKey: id), rowDecoratorsCellWidth: decoratorsCellWidth())),
+                                                                                    rowWidth: Utility.getWidthForExpanderRow(columns: filteredTableColumns, showSelector: showRowSelector(for: tableDataModel), showSingleClickEdit: showSingleClickEditButton(for: tableDataModel), showRowDecorators: tableDataModel.hasAnyRowDecorators(schemaKey: id), rowDecoratorsCellWidth: decoratorsCellWidth(for: tableDataModel))),
```
vivekjyani, line 977: **Clarity / non-blocking.** `expandTables(rowDataModel:level:)` has no `tableDataModel` parameter, so the `tableDataModel` referenced here is `self.tableDataModel` (the `@Published`). The previous `decoratorsCellWidth()` already read `self.tableDataModel` internally, so this change is behavior-identical and still touches the published property exactly once — it does **not** thread in a snapshot.

That is fine, because every caller of `expandTables` is on the main thread (SwiftUI tap handlers, `expandToRow`, and `refreshCollectionSchema` ← `cellDidChange` / `bulkEdit` which is `@MainActor`). Flagging only so a future reader does not assume the `(for:)` form made this call background-safe — it did not. The genuinely load-bearing fixes are lines 506 and 791, on the `getAllCellModels` background path where `tableDataModel` is the value-type snapshot captured at line 686.

Suggest either leaving this as `decoratorsCellWidth()` or adding a `// main-thread only` note so the intent is unambiguous.

vivekjyani: Good catch — agreed this is behavior-identical since expandTables is main-thread only and the (for:) arg here is just self.tableDataModel. Leaving it as-is for consistency with the other call sites; the real background-path fixes are at 506 and 791. I'll keep the // main-thread only intent in mind but won't change this line.

PR #344, opened 2026-06-23
NO-2213: Deprecate parameter-based DocumentEditor init in favor of config-based init

vivekjyani:
## Context
`DocumentEditor` historically exposed a single initializer with 10+ flat parameters (`mode`, `pageID`, `navigation`, `isPageDuplicateEnabled`, `isPageDeleteEnabled`, `validateSchema`, `license`, `singleClickRowEdit`, `decoratorConfig`, …). This is hard to read at call sites, hard to extend without churning the signature, and doesn't let consumers hold/pass around a reusable configuration object. This PR introduces a grouped config-object API and deprecates the flat one — additively, with zero breakage.

## What changed
- **`Sources/JoyfillUI/ViewModels/DocumentEditor.swift`**
  - New config value types: `PageConfig`, `DisplayConfig`, `DocumentEditorConfig` (each with a `public init` and defaults identical to the old parameter defaults). Field-level `// was \`oldName\`` comments map each new field to the legacy parameter for cross-reference (and Kotlin/Swift naming parity).
  - All config types conform to `Equatable` + `Sendable` (cascaded to `Mode` in `Types.swift` and `DecoratorConfig` in `DocumentEditor+Decorators.swift`). Synthesized automatically — every member is trivially conformable.
  - The **config-based initializer is now the designated init**: `init(document:events:config:)` holds the real initialization body.
  - The **old flat initializer is now a deprecated `convenience init`** that builds a `DocumentEditorConfig` and delegates to the designated init. Marked `@available(*, deprecated, message: "Use init(document:events:config:) with DocumentEditorConfig instead.")`.
- **`JoyfillSwiftUIExample/JoyfillTests/DocumentEditorConfigTests.swift`** — new test file (13 tests).
- **`JoyfillExample.xcodeproj/project.pbxproj`** — registers the new test file in the `JoyfillTests` target.

## Key decisions
- **Delegation direction flipped (config init = designated, old init = convenience).** This keeps all real logic in one place (the new API) and makes the deprecated path a thin adapter. The alternative — keeping the old init designated and having config delegate to it — would leave the deprecated code as the source of truth and emit deprecation warnings inside the SDK.
- **No internal deprecation warnings.** The only in-SDK caller of the old init is `Form.init(document:mode:events:pageID:navigation:)`, which is itself already deprecated; Swift suppresses deprecation warnings for calls made from within a deprecated declaration, so the framework builds clean.
- **Readonly coercion preserved exactly.** `mode == .readonly` still forces `isPageDuplicateEnabled`/`isPageDeleteEnabled` to `false` on the normal path, and the schema-failure early-return branch keeps its original (non-coerced) behavior — unchanged from before.
- **Additive / backward compatible.** Old call sites compile and run identically; they only get a deprecation hint.

## Public API change
**Yes.** Adds public `PageConfig`, `DisplayConfig`, `DocumentEditorConfig` and a new `DocumentEditor.init(document:events:config:)`. The old initializer remains public but is now deprecated. The config types (plus `Mode` and `DecoratorConfig`) now conform to `Equatable` + `Sendable` — additive, so it enables config diffing for the swap use case and keeps the types clean under strict concurrency without affecting existing callers.

**Docs:** SDK init docs / README examples should be updated to show the config-based init as the recommended path and note the old one is deprecated. Kotlin SDK should mirror these names for parity.

### Usage
```swift
// New (recommended)
let editor = DocumentEditor(
    document: doc,
    events: changeManager,
    config: DocumentEditorConfig(
        mode: .fill,
        license: licenseToken,
        page: PageConfig(navigation: true, enableDuplicates: true, enableDeletes: true, currentPageID: "page_id"),
        display: DisplayConfig(singleClickRowEdit: true, decorators: DecoratorConfig(visibleLimitInFields: 5, visibleLimitInRows: 3))
    )
)

// Old (still works, now deprecated)
let editor = DocumentEditor(document: doc, mode: .fill, events: changeManager, pageID: "page_id", navigation: true, ...)
```

## Tests
Covered in this PR — 13 tests in `DocumentEditorConfigTests`:
- Field mapping (config → editor properties) and struct defaults match legacy defaults
- Readonly coercion forces page duplicate/delete off
- Default config == legacy default init; non-default config == legacy init with equivalent params (guards the delegation mapping)
- `license` → `isCollectionFieldEnabled` mapping guarded both ways: a *valid* token via config enables collection fields (absolute assertion), and a nil license keeps them disabled — so dropping/zeroing `license` on the config path would fail the test
- `validateSchema` true on invalid doc sets `schemaError` + fires `onError`; false skips
- `currentPageID` resolves to the requested page; nil falls back to first valid page (matches legacy)
- `events` passthrough; two distinct configs produce independent editors (hold/swap use case)

Full `JoyfillTests` suite: **1094 tests, 0 failures.** (The legacy/deprecated init path is exercised by existing tests, confirming no behavior change.)

## Notes for reviewer
- No UI change — nothing to screenshot/record (initializer/API-shape only).
- The `init(document:events:config:)` body is a 1:1 port of the previous designated init with parameters replaced by `config.*`; diff is best read as "extract params into config struct" rather than logic changes.
- Heads-up: the example app's `licenseKey` (in `JoyfillExampleApp.swift`) is intentionally kept out of this PR; collection-field tests assert that token verifies against `LicenseValidator`'s public key.

### DIFF
--- a/JoyfillSwiftUIExample/JoyfillTests/DocumentEditorConfigTests.swift
+++ b/JoyfillSwiftUIExample/JoyfillTests/DocumentEditorConfigTests.swift
@@ -184,19 +184,24 @@ final class DocumentEditorConfigTests: XCTestCase {
         XCTAssertFalse(editor.isCollectionFieldEnabled)
     }
 
-    /// The license is routed through the same validator as the legacy init: a config
-    /// editor and a legacy editor given the same token must agree on the flag.
-    func testConfigInit_licensePassthrough_matchesLegacyInit() {
-        let token = "not-a-valid-license-token"
+    /// A *valid* license supplied through the config must actually flip
+    /// `isCollectionFieldEnabled` to `true`. Paired with the nil-license test above,
+    /// this guards the config → license mapping in both directions: dropping or
+    /// zeroing `license` on the config path would leave the flag `false` and fail here.
+    func testConfigInit_validLicense_enablesCollectionField() {
+        let license = (ProcessInfo.processInfo.environment["JOYFILL_TEST_LICENSE"] ?? licenseKey)
+            .trimmingCharacters(in: .whitespacesAndNewlines)
+        XCTAssertFalse(license.isEmpty,
+                       "Missing license: set JOYFILL_TEST_LICENSE env var or check licenseKey")
+        XCTAssertTrue(LicenseValidator.isCollectionEnabled(licenseToken: license),
+                      "License verification failed — the token does not match the public key in LicenseValidator")
 
-        let legacy = DocumentEditor(document: JoyDoc(),
-                                    validateSchema: false,
-                                    license: token)
-        let viaConfig = DocumentEditor(document: JoyDoc(),
-                                       config: DocumentEditorConfig(license: token,
-                                                                    validateSchema: false))
+        let editor = DocumentEditor(document: JoyDoc(),
+                                    config: DocumentEditorConfig(license: license,
+                                                                 validateSchema: false))
 
-        XCTAssertEqual(legacy.isCollectionFieldEnabled, viaConfig.isCollectionFieldEnabled)
+        XCTAssertTrue(editor.isCollectionFieldEnabled,
+                      "A valid license passed via DocumentEditorConfig must enable collection fields.")
     }
 
     // MARK: - validateSchema behaviour
--- a/JoyfillSwiftUIExample/JoyfillTests/DocumentEditorConfigTests.swift
+++ b/JoyfillSwiftUIExample/JoyfillTests/DocumentEditorConfigTests.swift
@@ -1,0 +1,308 @@
+import XCTest
+import JoyfillModel
+@testable import Joyfill
+
+/// Verifies the new `DocumentEditor.init(document:events:config:)` correctly
+/// maps a `DocumentEditorConfig` onto the editor, stays equivalent to the
+/// legacy parameter-based init, and preserves the readonly coercion rule.
+final class DocumentEditorConfigTests: XCTestCase {
+
+    // MARK: - Test doubles
+
+    /// Captures `onError` so schema-validation behaviour can be asserted.
+    private final class MockFormChangeEvent: FormChangeEvent {
+        private(set) var didReceiveError = false
+        private(set) var receivedError: JoyfillError?
+
+        func onChange(changes: [Change], document: JoyDoc) {}
+        func onFocus(event: Event) {}
+        func onBlur(event: Event) {}
+        func onUpload(event: UploadEvent) {}
+        func onCapture(event: CaptureEvent) {}
+        func onError(error: JoyfillError) {
+            didReceiveError = true
+            receivedError = error
+        }
+    }
+
+    /// A document that intentionally fails schema validation (missing required fields).
+    private func invalidDocument() -> JoyDoc {
+        JoyDoc(dictionary: [
+            "identifier": "test-doc",
+            "name": "Invalid Doc"
+        ])
+    }
+
+    /// A valid multi-page document built from the shared dummy-model helpers.
+    private func multiPageDocument() -> JoyDoc {
+        JoyDoc()
+            .setDocument()
+            .setFile()
+            .setPageWithFieldPosition()   // page id: "6629fab320fca7c8107a6cf6"
+            .addSecondPage()              // page id: "second_page_id_12345"
+            .setHeadingText()
+            .setTextField()
+    }
+
+    // MARK: - Field mapping
+
+    /// Every config field should land on the matching editor property.
+    func testConfigInit_mapsAllValuesOntoEditor() {
+        let config = DocumentEditorConfig(
+            mode: .fill,
+            license: nil,
+            validateSchema: false,
+            page: PageConfig(navigation: false,
+                             enableDuplicates: true,
+                             enableDeletes: true,
+                             currentPageID: nil),
+            display: DisplayConfig(singleClickRowEdit: true,
+                                   decorators: DecoratorConfig(visibleLimitInFields: 5,
+                                                               visibleLimitInRows: 3))
+        )
+
+        let editor = DocumentEditor(document: JoyDoc(), config: config)
+
+        XCTAssertEqual(editor.mode, .fill)
+        XCTAssertFalse(editor.showPageNavigationView)        // page.navigation
+        XCTAssertTrue(editor.isPageDuplicateEnabled)         // page.enableDuplicates
+        XCTAssertTrue(editor.isPageDeleteEnabled)            // page.enableDeletes
+        XCTAssertTrue(editor.singleClickRowEdit)             // display.singleClickRowEdit
+        XCTAssertEqual(editor.decoratorConfig.visibleLimitInFields, 5)
+        XCTAssertEqual(editor.decoratorConfig.visibleLimitInRows, 3)
+    }
+
+    // MARK: - Readonly coercion
+
+    /// Readonly mode must force page duplicate/delete off, even if the config
+    /// asked for them — same rule the legacy init enforces.
+    func testConfigInit_readonlyForcesPageEditingOff() {
+        let config = DocumentEditorConfig(
+            mode: .readonly,
+            validateSchema: false,
+            page: PageConfig(enableDuplicates: true, enableDeletes: true)
+        )
+
+        let editor = DocumentEditor(document: JoyDoc(), config: config)
+
+        XCTAssertEqual(editor.mode, .readonly)
+        XCTAssertFalse(editor.isPageDuplicateEnabled)
+        XCTAssertFalse(editor.isPageDeleteEnabled)
+    }
+
+    // MARK: - Defaults parity
+
+    /// A default config must produce the same editor as the legacy init's defaults.
+    func testConfigInit_defaultsMatchLegacyInit() {
+        let legacy = DocumentEditor(document: JoyDoc(), validateSchema: false)
+        let viaConfig = DocumentEditor(document: JoyDoc(),
+                                       config: DocumentEditorConfig(validateSchema: false))
+
+        XCTAssertEqual(legacy.mode, viaConfig.mode)
+        XCTAssertEqual(legacy.showPageNavigationView, viaConfig.showPageNavigationView)
+        XCTAssertEqual(legacy.isPageDuplicateEnabled, viaConfig.isPageDuplicateEnabled)
+        XCTAssertEqual(legacy.isPageDeleteEnabled, viaConfig.isPageDeleteEnabled)
+        XCTAssertEqual(legacy.singleClickRowEdit, viaConfig.singleClickRowEdit)
+        XCTAssertEqual(legacy.decoratorConfig.visibleLimitInFields,
+                       viaConfig.decoratorConfig.visibleLimitInFields)
+        XCTAssertEqual(legacy.decoratorConfig.visibleLimitInRows,
+                       viaConfig.decoratorConfig.visibleLimitInRows)
+    }
+
+    /// Sanity-check that the struct defaults themselves match the legacy init defaults,
+    /// independent of how the editor consumes them.
+    func testConfigDefaults_matchLegacyParameterDefaults() {
+        let config = DocumentEditorConfig()
+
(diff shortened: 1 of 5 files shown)


Sources/JoyfillUI/ViewModels/DocumentEditor.swift
```diff
@@ -42,6 +42,54 @@ public extension DocumentEditorDelegate {
...
+    public var decorators: DecoratorConfig      // was `decoratorConfig`
+
+    public init(singleClickRowEdit: Bool = false,
+                decorators: DecoratorConfig = DecoratorConfig()) {
+        self.singleClickRowEdit = singleClickRowEdit
+        self.decorators = decorators
+    }
+}
+
+public struct DocumentEditorConfig {
```
vivekjyani: Non-blocking: the PR pitches a "hold and pass around / swap config" use case, but `DocumentEditorConfig`/`PageConfig`/`DisplayConfig` are neither `Equatable` (consumers can't diff two configs) nor `Sendable` (can't pass across concurrency domains without warnings). All stored members are trivially conformable. Caveat — to synthesize `Equatable`/`Sendable` here you'd also need to add the same conformance to `Mode` (Types.swift:118, plain enum) and `DecoratorConfig` (DocumentEditor+Decorators.swift:742, two `Int`s); both are additive/safe. Cheaper to add now than as a follow-up on a type consumers already hold.

Sources/JoyfillUI/ViewModels/DocumentEditor.swift
```diff
@@ -42,6 +42,54 @@ public extension DocumentEditorDelegate {
     func decoratorsDidChange() {}
 }
 
+public struct PageConfig {
+    public var navigation: Bool          // was `navigation`
```
vivekjyani, line 46: Non-blocking naming nit: `navigation: Bool` drives `showPageNavigationView` — a bare `navigation` bool reads like it should hold a navigation object rather than a visibility toggle (`showNavigation` would be clearer). Same for `currentPageID` (line 49): it's the *initial/target* page selector consumed once via `firstValidPageID(for:)`, not the live current page, so it can mislead. Both mirror the legacy parameter names for parity, so this is a judgment call — worth aligning with the Kotlin SDK naming before this API hardens.

vivekjyani (commented):
Reviewed the full diff + surrounding call sites. Solid, backward-compatible refactor — config init as designated, old flat init as deprecated convenience. I verified with a minimal `swiftc` repro that the untouched bare `DocumentEditor(document:)` / `DocumentEditor(document:events:)` call sites are NOT ambiguous: Swift prefers the fewer-parameter overload, so they bind to the new designated init with no warning and identical behavior. No blocking issues. A couple of non-blocking suggestions inline.

Sources/JoyfillUI/ViewModels/DocumentEditor.swift
vivekjyani: Good call — done. Added Equatable, Sendable to PageConfig, DisplayConfig, DocumentEditorConfig, and the cascade to Mode (Types.swift) and DecoratorConfig. All synthesized automatically since every member is trivially conformable, so no hand-written ==. Agreed it's cheaper now than as a follow-up on a type consumers already hold — Equatable enables config diffing for the swap use case, and Sendable keeps it clean under strict concurrency. Build is clean (no conformance warnings) and the config tests still pass.

Sources/JoyfillUI/ViewModels/DocumentEditor.swift
vivekjyani: Thanks — fair points, and I went and checked the Kotlin SDK before deciding. Both fields intentionally match the Kotlin PageConfig equivalents (navigation, currentPageID), so I'm keeping them as-is: cross-SDK naming parity is the higher-order goal here, and diverging the Swift names would create a worse inconsistency than the local ambiguity. Agreed that, read in isolation, navigation looks like it could hold an object and currentPageID reads like live state rather than the one-shot seed it is — but aligning the two SDKs wins. If we ever revisit the naming, it should be a coordinated rename across both SDKs rather than a Swift-only change. Leaving as-is for parity.

JoyfillSwiftUIExample/JoyfillTests/DocumentEditorConfigTests.swift
```diff
@@ -0,0 +1,308 @@
...
+        let token = "not-a-valid-license-token"
+
+        let legacy = DocumentEditor(document: JoyDoc(),
+                                    validateSchema: false,
+                                    license: token)
+        let viaConfig = DocumentEditor(document: JoyDoc(),
+                                       config: DocumentEditorConfig(license: token,
+                                                                    validateSchema: false))
+
+        XCTAssertEqual(legacy.isCollectionFieldEnabled, viaConfig.isCollectionFieldEnabled)
```
vivekjyani: Non-blocking (test strength): with `not-a-valid-license-token`, both sides resolve to `false` — `LicenseValidator.isCollectionEnabled` returns `false` for an invalid token, same as it does for `nil`. So this assertion would still hold even if the config path dropped `license` entirely (e.g. a regression where the convenience init forwarded `license: nil`). It therefore can't actually guard the mapping it's named for.

It's also worth noting that since the legacy init now delegates to the config init, both `legacy` and `viaConfig` funnel through the *same* final `LicenseValidator.isCollectionEnabled(licenseToken:)` call — so this is effectively comparing one path to itself.

To make it bite, use a token where `isCollectionEnabled` returns `true` (the example app's `licenseKey` / `JOYFILL_TEST_LICENSE`, like `ValidationTestCase` does) so that dropping/zeroing the license flips the result and fails the test. As-is, the valid-token (true) case is covered only transitively via `ValidationTestCase`/`DecoratorPublicAPITests`.

vivekjyani: Good catch — fixed. Replaced the test with testConfigInit_validLicense_enablesCollectionField: it resolves a valid token (JOYFILL_TEST_LICENSE ?? licenseKey), guards that it verifies against LicenseValidator, then builds the editor via the config path and asserts isCollectionFieldEnabled == true. Together with the existing nil-license test (false case) this now guards the mapping in both directions — dropping/zeroing license on the config path flips the true case to false and fails. Also drops the circular legacy-vs-config comparison you flagged. Same valid-token dependency as ValidationTestCase, so it's consistent in CI.

Sources/JoyfillUI/ViewModels/DocumentEditor.swift
```diff
@@ -82,33 +130,25 @@ public class DocumentEditor: ObservableObject {
...
                 events: FormChangeEvent? = nil,
-                pageID: String? = nil,
-                navigation: Bool = true,
-                isPageDuplicateEnabled: Bool = false,
-                isPageDeleteEnabled: Bool = false,
-                validateSchema: Bool = true,
-                license: String? = nil,
-                singleClickRowEdit: Bool = false,
-                decoratorConfig: DecoratorConfig = DecoratorConfig()) {
+                config: DocumentEditorConfig = DocumentEditorConfig()) {
```
VishnuBishnoi, line 134: Good direction making the config init the **designated** one and the flat init a thin deprecated `convenience` adapter — keeps the single source of truth on the new path and avoids in-SDK deprecation warnings. Verified the body is a 1:1 port (every `config.*` lands on the same property the old param did, including `firstValidPageID(for: config.page.currentPageID, …)`).

Sources/JoyfillUI/ViewModels/DocumentEditor.swift
```diff
@@ -140,11 +180,37 @@ public class DocumentEditor: ObservableObject {
             updatePageFieldModels(page, pageID, fileID)
         }
         self.validationHandler = ValidationHandler(documentEditor: self)
-        self.currentPageID = document.firstValidPageID(for: pageID, conditionalLogicHandler: conditionalLogicHandler)
+        self.currentPageID = document.firstValidPageID(for: config.page.currentPageID, conditionalLogicHandler: conditionalLogicHandler)
         self.JoyfillDocContext = Joyfill.JoyfillDocContext(docProvider: self)
         self.currentPageOrder = document.pageOrderForCurrentView ?? []
     }
     
+    @available(*, deprecated, message: "Use init(document:events:config:) with DocumentEditorConfig instead.")
```
VishnuBishnoi, line 188: Deprecation message points users to the new init — nice. Worth confirming the team is comfortable freezing the `PageConfig`/`DisplayConfig`/`DocumentEditorConfig` shape now, since it becomes public API the moment this ships (and the `// was \`oldName\`` comments are a helpful migration aid, though they'll need pruning once the old init is eventually removed).

Sources/JoyfillUI/View/Types.swift
```diff
@@ -115,7 +115,7 @@ public struct PageEvent {
     }
 }
 
-public enum Mode {
+public enum Mode: Equatable, Sendable {
```
VishnuBishnoi, line 118: Adding `Equatable`/`Sendable` to public `Mode` is additive and correct here. Minor heads-up: retroactively adding a public conformance can, in rare cases, collide with a downstream consumer that already declared the same conformance in an extension — almost certainly fine for an enum like this, just noting it as the one theoretical source-compat edge of this change.

VishnuBishnoi (approved):
**Approving.**

Clean, backward-compatible API refactor. The config-based init reads as a faithful "extract params into config structs" port — I checked the field-by-field mapping in both the schema-failure early-return branch and the normal branch and it's 1:1 with the old parameters, the new struct defaults match the old parameter defaults exactly, and the readonly coercion (`mode == .readonly ? false : …`) is preserved on the normal path and intentionally not applied on the schema-failure path (matching prior behavior).

I specifically checked the additive/non-breaking claim against the two things most likely to break it:
- **Overload ambiguity** — a bare `DocumentEditor(document:)` exists in the example app ([FormBuilderView.swift:823](JoyfillSwiftUIExample/JoyfillExample/FormBuilderView.swift)) and matches *both* inits via defaults. CI Build is green, so it resolves cleanly (Swift disfavors the `@available(deprecated)` overload), binding to the new config init with identical defaults → no behavior change.
- **In-SDK deprecation warning** — the one internal caller of the old init ([FormView.swift:12](Sources/JoyfillUI/View/FormView.swift)) sits inside `Form.init(...)`, which is itself `@available(*, deprecated)` (L10), so Swift suppresses the warning and the framework builds clean as claimed.

`Equatable`/`Sendable` cascade to `Mode` and `DecoratorConfig` is additive and synthesizes fine (Build green). Tests are thorough (13) — defaults parity, non-default legacy↔config equivalence, readonly coercion, license both directions, validateSchema true/false, currentPageID resolution + nil fallback, events passthrough, and independent editors for the hold/swap case. The license test follows the existing `ValidationTestCase` convention (CI injects `JOYFILL_TEST_LICENSE`); Unit Tests pass (1094, 0 failures).

One follow-up (non-blocking): since this lands new public API, please make sure the SDK init docs / README and the Kotlin SDK names are updated for parity as you noted. Notes inline.

Sources/JoyfillUI/View/Types.swift
vivekjyani: Good call-out. For a plain SDK-owned enum like Mode the collision edge is effectively nil — it'd only trigger if a downstream consumer had declared extension Mode: Equatable/Sendable themselves, which isn't possible for a type they don't own. Comfortable shipping it as additive.

Sources/JoyfillUI/ViewModels/DocumentEditor.swift
vivekjyani: Thanks for verifying the port end-to-end. That was the intent — keep the single source of truth on the config path and let the flat init be a thin deprecated adapter so we get zero in-SDK deprecation warnings.

Sources/JoyfillUI/ViewModels/DocumentEditor.swift
vivekjyani: Yes — the shape is intended to be frozen here. Field names and grouping are deliberate and mirror the Kotlin SDK for cross-platform parity, so we're treating this as the stable public surface. The // was oldName comments are intentional migration aids; we'll prune them (along with the deprecated convenience init) when we eventually cut the old API. I'll file a follow-up so that cleanup is tracked.

PR #345, opened 2026-06-26
NO-2228: Preserve required value key on table/collection fields when copying a page without values

vivekjyani:
## Context
When a user copies a page **without values**, the duplicated page must keep its structure but drop field data. The copied document is later run through JSON-schema validation. For most field types omitting the field value is fine, but `table` and `collection` mark `value` as **required** in the schema — so dropping that key produced an invalid document and schema validation failed.

## What changed
- `Sources/JoyfillUI/ViewModels/DocumentEditor.swift` — in `duplicatePage(..., copyWithValues: false)`, the value-stripping loop now special-cases `table` and `collection`:
  - `table` → `value = .valueElementArray([])` and `rowOrder = []`
  - `collection` → `value = .valueElementArray([])`
  - all other field types → unchanged (`value = nil`)
- `JoyfillSwiftUIExample/JoyfillTests/PageDeletableCopyableValidationTests.swift` — added two regression tests covering the table and collection cases.

## Why this approach
- Setting `value = nil` removes the key entirely (the setter does `dictionary["value"] = newValue?.dictionary`), which is what broke schema validation. Clearing to an empty array preserves the required key while still discarding the data.
- Only `table` and `collection` list `value` in their schema `required` array (verified across all field definitions); every other array-valued type (`image`, `file`, `multiSelect`, `dropdown`, `chart`, …) leaves `value` optional, so omitting it stays valid — hence no change for them.
- A field's business `required: true` flag does not affect this — the schema has no conditional (`if/then`/`dependentRequired`) rules, so it cannot promote `value` into a `required` array. Business-required is enforced separately by `Validator`, not schema validation.
- `rowOrder` is cleared for tables so it stays consistent with the now-empty rows.

## Screenshots / video
N/A — no visual change. The fix is that copying a page without values no longer fails schema validation; behavior is verified via automated tests.

## Public API
No public API changes. `duplicatePage(pageID:copyWithValues:)` signature is unchanged. No docs update required.

## Test coverage
Covered now. Added in `PageDeletableCopyableValidationTests`:
- `testDuplicatePage_WithoutValues_TableFieldKeepsEmptyValueArray` and `testDuplicatePage_WithoutValues_CollectionFieldKeepsEmptyValueArray` — mechanism checks: the `value` key stays present as an empty array (and table `rowOrder` is cleared).
- `testDuplicatePage_WithoutValues_PassesRealSchemaValidation` — end-to-end: builds a fully schema-valid document (table + collection), copies the page without values, and asserts `JoyfillSchemaManager().validateSchema(...) == nil`. The fixture is schema-complete and the test also asserts the base document is valid before the copy.

Verified the e2e test actually guards the bug: temporarily reverting the fix to `value = nil` makes it fail. Full unit suite passes (1083 tests, 0 failures).

## Notes for reviewer
- Test layering: mechanism tests pinpoint the cause (fast, robust); the end-to-end test exercises the real JSON-schema validation this PR fixes.
- Scope is intentionally minimal: only the `copyWithValues == false` branch is touched; the `copyWithValues == true` path is untouched.

### DIFF
--- a/JoyfillSwiftUIExample/JoyfillTests/PageDeletableCopyableValidationTests.swift
+++ b/JoyfillSwiftUIExample/JoyfillTests/PageDeletableCopyableValidationTests.swift
@@ -315,6 +315,116 @@ final class PageDeletableCopyableValidationTests: XCTestCase {
         XCTAssertTrue(hasValue, "copyWithValues=true should preserve at least one field value on the duplicated page.")
     }
 
+    /// Regression: copying "without values" must NOT drop the `value` key on table/collection
+    /// fields. The JSON schema marks `value` as required for those types, so nilling it out
+    /// (which removes the key) makes the duplicated document fail schema validation. They must
+    /// instead be cleared to an empty array so the key survives.
+    func testDuplicatePage_WithoutValues_TableFieldKeepsEmptyValueArray() {
+        var document = JoyDoc()
+            .setDocument()
+            .setFile()
+            .setPageWithFieldPosition()
+
+        // Add a table field (with rows) positioned on the page that will be duplicated.
+        var tableField = JoyDocField()
+        tableField.id = "regression_table_field"
+        tableField.type = "table"
+        tableField.file = "6629fab3c0ba3fb775b4a55c"
+        tableField.rowOrder = ["regression_row_1"]
+        tableField.value = .valueElementArray([ValueElement(id: "regression_row_1")])
+        document.fields.append(tableField)
+
+        let tablePosition = FieldPosition(dictionary: [
+            "field": "regression_table_field",
+            "displayType": "original",
+            "width": 12, "height": 8, "x": 0, "y": 16,
+            "_id": "regression_table_position",
+            "type": "table"
+        ])
+        document.files[0].pages?[0].fieldPositions?.append(tablePosition)
+
+        let pageID = "6629fab320fca7c8107a6cf6"
+        let editor = makeEditor(document: document)
+
+        editor.duplicatePage(pageID: pageID, copyWithValues: false)
+
+        // Locate the duplicated page and its table field.
+        let firstFile = editor.document.files.first
+        guard let duplicatedPageID = firstFile?.pageOrder?.first(where: { $0 != pageID }),
+              let duplicatedPage = firstFile?.pages?.first(where: { $0.id == duplicatedPageID }) else {
+            XCTFail("Duplicated page not found.")
+            return
+        }
+        let duplicatedFieldIDs = Set(duplicatedPage.fieldPositions?.compactMap { $0.field } ?? [])
+        let duplicatedFields = editor.document.fields.filter { duplicatedFieldIDs.contains($0.id ?? "") }
+        guard let duplicatedTable = duplicatedFields.first(where: { $0.fieldType == .table }) else {
+            XCTFail("Duplicated table field not found on the duplicated page.")
+            return
+        }
+
+        // The required `value` key must still exist and be an (empty) array, not absent.
+        XCTAssertNotNil(duplicatedTable.dictionary["value"],
+                        "Table field must keep its required `value` key when copied without values.")
+        let valueArray = duplicatedTable.dictionary["value"] as? [Any]
+        XCTAssertNotNil(valueArray, "Table field `value` must be an array when copied without values.")
+        XCTAssertTrue(valueArray?.isEmpty ?? false, "Table field rows must be cleared when copied without values.")
+        XCTAssertEqual(duplicatedTable.rowOrder ?? ["non-empty"], [],
+                       "Table field rowOrder must be cleared to stay consistent with the emptied value.")
+    }
+
+    /// Regression: same as the table case, but for collection fields. The schema marks
+    /// `value` as required for collections too, so copying "without values" must clear it to
+    /// an empty array (keeping the key) rather than nilling it out (dropping the key).
+    func testDuplicatePage_WithoutValues_CollectionFieldKeepsEmptyValueArray() {
+        var document = JoyDoc()
+            .setDocument()
+            .setFile()
+            .setPageWithFieldPosition()
+
+        // Add a collection field (with items) positioned on the page that will be duplicated.
+        var collectionField = JoyDocField()
+        collectionField.id = "regression_collection_field"
+        collectionField.type = "collection"
+        collectionField.file = "6629fab3c0ba3fb775b4a55c"
+        collectionField.value = .valueElementArray([ValueElement(id: "regression_item_1")])
+        document.fields.append(collectionField)
+
+        let collectionPosition = FieldPosition(dictionary: [
+            "field": "regression_collection_field",
+            "displayType": "original",
+            "width": 12, "height": 10, "x": 0, "y": 24,
+            "_id": "regression_collection_position",
+            "type": "collection"
+        ])
+        document.files[0].pages?[0].fieldPositions?.append(collectionPosition)
+
+        let pageID = "6629fab320fca7c8107a6cf6"
+        let editor = makeEditor(document: document)
+
+        editor.duplicatePage(pageID: pageID, copyWithValues: false)
+
+        // Locate the duplicated page and its collection field.
+        let firstFile = editor.document.files.first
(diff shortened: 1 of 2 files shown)


JoyfillSwiftUIExample/JoyfillTests/PageDeletableCopyableValidationTests.swift
```diff
@@ -315,6 +315,116 @@ final class PageDeletableCopyableValidationTests: XCTestCase {
...
+            "field": "regression_table_field",
+            "displayType": "original",
+            "width": 12, "height": 8, "x": 0, "y": 16,
+            "_id": "regression_table_position",
+            "type": "table"
+        ])
+        document.files[0].pages?[0].fieldPositions?.append(tablePosition)
+
+        let pageID = "6629fab320fca7c8107a6cf6"
+        let editor = makeEditor(document: document)
```
vivekjyani, line 347: These regression tests go through `makeEditor` → `validateSchema: false`, so they assert the **mechanism** (the `value` key stays present as an empty array) but never run the JSON-schema validation that this PR actually fixes. They'd catch a regression back to `value = nil`, but not a subtler schema violation.

Consider strengthening to an end-to-end check: build a fully schema-valid document, copy the page without values, then assert `JoyfillSchemaManager().validateSchema(document: editor.document) == nil`. That public API is already exercised in `SchemaValidationTests.swift` and would fail loudly on any future schema break the mechanism check can't see. (Non-blocking.)

JoyfillSwiftUIExample/JoyfillTests/PageDeletableCopyableValidationTests.swift
```diff
@@ -315,6 +315,116 @@ final class PageDeletableCopyableValidationTests: XCTestCase {
...
+            .setFile()
+            .setPageWithFieldPosition()
+
+        // Add a table field (with rows) positioned on the page that will be duplicated.
+        var tableField = JoyDocField()
+        tableField.id = "regression_table_field"
+        tableField.type = "table"
+        tableField.file = "6629fab3c0ba3fb775b4a55c"
+        tableField.rowOrder = ["regression_row_1"]
+        tableField.value = .valueElementArray([ValueElement(id: "regression_row_1")])
```
vivekjyani, line 334: Heads-up if you adopt the end-to-end validation suggestion above: this fixture is itself schema-incomplete. `TableField.required` is `[_id, file, rowOrder, tableColumnOrder, tableColumns, type, value]`, but this table only sets `_id/type/file/rowOrder/value` — it would need `tableColumns` + `tableColumnOrder` to pass real schema validation.

The production path is fine (the duplicated field copies those keys from the original); this only matters for a hand-built test fixture. (Non-blocking.)

JoyfillSwiftUIExample/JoyfillTests/PageDeletableCopyableValidationTests.swift
vivekjyani: Good call — added `testDuplicatePage_WithoutValues_PassesRealSchemaValidation` in commit d998532d. It builds a schema-valid document, copies the page without values, and asserts `JoyfillSchemaManager().validateSchema(document: editor.document) == nil`, so it runs the real JSON-schema validation rather than just the mechanism. Kept the existing mechanism tests too. Verified it guards the bug: temporarily reverting the fix to `value = nil` makes this test fail.

JoyfillSwiftUIExample/JoyfillTests/PageDeletableCopyableValidationTests.swift
vivekjyani: Thanks for the heads-up — handled in the new end-to-end test (commit d998532d). The hand-built fixture there is schema-complete: the table field includes `tableColumns` + `tableColumnOrder` (plus `rowOrder`/`value`) and the collection field includes the required `schema` key. The test also asserts the base document is schema-valid *before* the copy, so any future incomplete fixture would fail loudly.

Sources/JoyfillUI/ViewModels/DocumentEditor.swift
```diff
@@ -1147,7 +1147,17 @@ extension DocumentEditor {
                 let isReadonly = newFields[i].disabled == true
                 let isDisplayText = newFields[i].fieldType == .block
                 if isReadonly || isDisplayText { continue }
-                newFields[i].value = nil
+                // Table and collection fields mark `value` as required in the schema, so clearing
+                // them to nil (which drops the key) would fail validation. Empty them instead.
+                switch newFields[i].fieldType {
+                case .table:
```
VishnuBishnoi, line 1153: Confirmed correct against the schema: `TableField.required` = [`_id`, `file`, `rowOrder`, `tableColumnOrder`, `tableColumns`, `type`, `value`] and `CollectionField.required` = [`_id`, `file`, `schema`, `type`, `value`]. So both need `value` kept (empty array ✓), table additionally needs `rowOrder` kept (cleared to `[]` ✓), and `tableColumns`/`tableColumnOrder`/`schema` are structural and preserved by the copy. Clean.

JoyfillSwiftUIExample/JoyfillTests/PageDeletableCopyableValidationTests.swift
```diff
@@ -315,6 +315,184 @@ final class PageDeletableCopyableValidationTests: XCTestCase {
...
+    /// End-to-end regression: build a fully schema-valid document containing both a table and a
+    /// collection field, copy the page WITHOUT values, then run the real JSON-schema validation
+    /// (`JoyfillSchemaManager.validateSchema`) on the resulting document. This guards the actual
+    /// bug this fix addresses — the mechanism tests above only assert the `value` key survives,
+    /// they never run schema validation (they use `validateSchema: false`).
+    ///
+    /// The fixture is intentionally schema-complete: `TableField` requires
+    /// `tableColumns`/`tableColumnOrder` and `CollectionField` requires `schema`, so those keys
+    /// are all present here.
+    func testDuplicatePage_WithoutValues_PassesRealSchemaValidation() {
```
VishnuBishnoi, line 437: This e2e test is the one that truly guards the regression — it exercises the real JSON-schema validation the mechanism tests skip (`validateSchema: false`), and asserting the hand-built fixture is schema-valid *before* the copy keeps the post-copy assertion meaningful. 👍

VishnuBishnoi (approved):
**Approving.**

Focused, correct fix. I independently verified the central claim against `Sources/JoyfillUI/View/joyfill-schema.swift`: only `TableField` (L1784–1792) and `CollectionField` (L2517–2523) list top-level `value` in their `required` array — the only other bare `"value"` in a required array (L1580) belongs to the unrelated `Option` sub-definition, so every other array-valued type (`image`, `chart`, `multiSelect`, …) leaves `value` optional and is correctly left on the `value = nil` path. Good call clearing `rowOrder = []` for tables — `rowOrder` is *also* in TableField's `required` array, so leaving it pointing at now-deleted rows would be inconsistent (and nilling it would drop a required key).

The fix is minimal and scoped strictly to the `copyWithValues == false` branch; the `copyWithValues == true` path and read-only/block skip behavior are untouched.

Test coverage is strong — the two mechanism tests pinpoint the cause, and the end-to-end test running real `JoyfillSchemaManager().validateSchema(...)` (and asserting the base fixture is valid first) is what actually guards the bug; appreciate that you confirmed it fails when reverting to `value = nil`. CI is green (Build + Unit Tests). Note inline.

Sources/JoyfillUI/ViewModels/DocumentEditor.swift
vivekjyani: Thanks for double-checking against the schema — that matches the analysis: only table and collection list value in their required array, so they're the two that need the key kept (empty array), with rowOrder additionally cleared to [] for tables. Everything else stays on the value = nil path.

JoyfillSwiftUIExample/JoyfillTests/PageDeletableCopyableValidationTests.swift
vivekjyani: Appreciate it 👍 — that was exactly the intent: the mechanism tests pinpoint the cause cheaply, and this e2e test runs the real validateSchema the others skip. The pre-copy "base is valid" assertion is there so the post-copy check can't silently pass on an already-invalid fixture.

PR #347, opened 2026-07-02
NO-1729: Formula date handling, array rendering & full formula-function test automation

vivekjyani:
## Context
The formula engine's date handling and array-to-string rendering had diverged from the cross-platform spec, and formula functions lacked systematic test coverage. This branch aligns runtime behavior with the spec and adds fixture-driven tests for the full formula function set.

## What changed
**Engine (source):**
- `DateFunctions.extractDate` now parses date **strings** (ISO 8601 with/without fractional seconds, and `yyyy-MM-dd`) in addition to numeric timestamps; timestamp seconds/ms heuristic factored into a helper. `dateAdd` returns an invalid timestamp (`0`) for bad args instead of throwing type-mismatch errors — matching spec's soft-fail behavior.
- `Evaluator.stringify` renders `.date` values as **epoch milliseconds** for string concatenation (was ISO string).
- `JoyfillDocContext`: array formula results on free-text display fields (`text`/`textarea`/`richText`/`block`) render as a bracketed string (e.g. `"[8.0, 12.0]"`); other field types keep the array.
- `DocumentEditor`: `currentPageID` init now respects an optional `config.page.currentPageID` (nil-coalesced); renamed private `JoyfillDocContext` property to `internal joyDocContext`.
- `DocumentEditor+Formulas.updateValue` now re-evaluates dependent formulas after a programmatic value update.

**Tests / fixtures / docs:**
- ~40 `FormulaTemplate_*FunctionTests.swift` suites with fixture-driven, exact-value assertions plus dynamic-update checks.
- JSON + `.md` fixtures per function under `Formula-sample/All Functions/`.
- Example-app UI (`FormulaFunctionTestsView`, `FailureCaptureDemoView`) + testing guide docs.

## Decisions
- **Soft-fail in `dateAdd` (return `0`)** rather than propagating errors: the cross-platform spec treats bad date/unit args as an invalid timestamp, so tests assert `0` instead of a thrown error.
- **Dates as epoch millis in `stringify`**: concatenation output must match other platforms; ISO strings didn't.
- **Bracketed-string array rendering scoped to free-text field types only** so structured fields (e.g. chart) keep real array values.

## Screenshot / video
N/A for the engine changes (no UI behavior change). The example app gains a Formula Function Tests screen for manual inspection.

## Public API
- `DocumentEditor.updateValue(for:value:)` is now **public** (was internal) — additive, no breaking change.
- `config.page.currentPageID` is treated as optional at init.
- Docs: worth a note in formula docs that date string parsing and epoch-millis stringification are now supported/expected; SDK API docs should reflect the newly-public `updateValue`.

## Tests
Covered now — fixture-driven suites for the full function set with exact and dynamic-update assertions.

## Notes for reviewer
- Two working-tree files (`JoyfillExampleApp.swift`, `ValidationTestCase.swift`) are intentionally left uncommitted and are **not** part of this PR.
- Focus review on the 5 source files under `Sources/`; the bulk of the diff is test fixtures/data.


vivekjyani:
Reverting: keeping #267 as the canonical PR per request.

PR #348, opened 2026-07-02
NO-2234: Add file/field variant to SchemaLogicCondition and bump schema to 1.0.5

vivekjyani:
## Context
`SchemaLogicCondition` previously modeled only *schema/collection*-scoped conditional logic (`schema` + `column`). Conditional logic also needs to target regular *page fields* (`file` + `page` + `field`), which the schema had no way to express. This PR widens the schema definition to cover both shapes.

## What changed
- **`Sources/JoyfillUI/View/joyfill-schema.swift`**
  - `SchemaLogicCondition` is now a `oneOf` of two object variants:
    - existing collection form — required: `schema`, `column`, `condition`
    - new field form — required: `file`, `field`, `condition` (plus optional `page`, `_id`, `value`)
  - Bumped `$joyfillSchemaVersion` from `1.0.4` → `1.0.5`.
- **`JoyfillSwiftUIExample/JoyfillTests/SchemaValidation/SchemaValidationTests.swift`**
  - Updated `expectedSchemaVersion` constant `1.0.4` → `1.0.5` to match the bumped schema.

## Decisions
- Used `oneOf` rather than merging all keys into one loosened object so each variant keeps its own `required` set — a field condition can't accidentally validate with a `column`, and vice versa.
- Kept `_id` and `value` optional in both variants to preserve backward compatibility with existing documents.
- No source changes needed for the version bump: the SDK reads `$joyfillSchemaVersion` dynamically via `JoyfillSchemaManager.currentSchemaVersion()`, so only the schema JSON and the test's hardcoded expectation required edits.

## Media
N/A — schema/validation-only change, no UI surface to screenshot.

## Public API / docs
- No public Swift API signature changes.
- The schema contract *is* a public interface: consumers relying on the schema version string will now see `1.0.5`, and the new file/field condition shape should be reflected in schema docs. **Schema docs likely need updating** to document the two condition variants.

## Tests
- Existing `SchemaValidationTests` pass with the updated expected version.
- No new positive/negative test was added for the file/field variant specifically — can add coverage asserting a `file`/`field` condition validates and that mixing keys fails, if reviewers want it in this PR.

## Notes for reviewers
- Please confirm the required-key set for the field variant (`file`, `field`, `condition`) matches the platform/Kotlin schema so cross-SDK behavior stays aligned.

### DIFF
--- a/JoyfillSwiftUIExample/JoyfillTests/SchemaValidation/SchemaValidationTests.swift
+++ b/JoyfillSwiftUIExample/JoyfillTests/SchemaValidation/SchemaValidationTests.swift
@@ -12,7 +12,7 @@ import Joyfill
 
 final class SchemaValidationTests: XCTestCase {
     /// The schema version expected in `SchemaValidationError.details`.
-    private let expectedSchemaVersion = "1.0.4"
+    private let expectedSchemaVersion = "1.0.5"
     private let sdkVersion = "3.0.0-rc22"
     
     func documentEditor(document: JoyDoc) -> DocumentEditor {
--- a/JoyfillSwiftUIExample/JoyfillTests/SchemaValidation/SchemaValidationTests.swift
+++ b/JoyfillSwiftUIExample/JoyfillTests/SchemaValidation/SchemaValidationTests.swift
@@ -374,6 +374,99 @@ final class SchemaValidationTests: XCTestCase {
             XCTAssertNil(err, "ChartPoint with \(label) should validate against the relaxed schema")
         }
     }
+
+    // MARK: - Schema 1.0.5: SchemaLogicCondition oneOf (collection vs field form)
+
+    /// Builds a valid document whose collection field carries a `SchemaDefinition.logic`
+    /// with a single `SchemaLogicCondition` equal to `condition`, so the `oneOf`
+    /// definition at `SchemaLogicCondition` is actually exercised.
+    private func documentWithSchemaLogicCondition(_ condition: [String: Any]) -> JoyDoc {
+        var docDict = minimalValidDocumentDictionary()
+        let collectionField: [String: Any] = [
+            "_id": "collection1",
+            "type": "collection",
+            "file": "file1",
+            "value": [],
+            "schema": [
+                "schema1": [
+                    "root": true,
+                    "tableColumns": [],
+                    "logic": [
+                        "action": "show",
+                        "eval": "and",
+                        "conditions": [condition]
+                    ]
+                ]
+            ]
+        ]
+        docDict["fields"] = [collectionField]
+        return JoyDoc(dictionary: docDict)
+    }
+
+    // (b) Collection form (schema/column/condition) still validates — regression guard.
+    func testSchemaLogicCondition_CollectionForm_ShouldValidate() {
+        let condition: [String: Any] = [
+            "_id": "cond1",
+            "schema": "schema1",
+            "column": "col1",
+            "condition": "=",
+            "value": "foo"
+        ]
+        let err = JoyfillSchemaManager().validateSchema(document: documentWithSchemaLogicCondition(condition))
+        XCTAssertNil(err, "Collection-form SchemaLogicCondition (schema/column/condition) should validate")
+    }
+
+    // (a) New field form (file/field/condition) validates.
+    func testSchemaLogicCondition_FieldForm_ShouldValidate() {
+        let condition: [String: Any] = [
+            "_id": "cond1",
+            "file": "file1",
+            "page": "page1",
+            "field": "field1",
+            "condition": "=",
+            "value": "foo"
+        ]
+        let err = JoyfillSchemaManager().validateSchema(document: documentWithSchemaLogicCondition(condition))
+        XCTAssertNil(err, "Field-form SchemaLogicCondition (file/field/condition) should validate")
+    }
+
+    // (a') Field form with optional `page` omitted still validates.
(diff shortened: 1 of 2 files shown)


JoyfillSwiftUIExample/JoyfillTests/SchemaValidation/SchemaValidationTests.swift
```diff
@@ -12,7 +12,7 @@ import Joyfill
 
 final class SchemaValidationTests: XCTestCase {
     /// The schema version expected in `SchemaValidationError.details`.
-    private let expectedSchemaVersion = "1.0.4"
+    private let expectedSchemaVersion = "1.0.5"
```
VishnuBishnoi, line 15: This version-constant bump is the only test change, but the whole point of the PR — the new `file`/`field` condition variant — has **no test coverage** (grep in `SchemaValidation/` finds nothing exercising `SchemaLogic`/`SchemaLogicCondition` at all). The suite already has the machinery (`minimalValidDocumentDictionary()`, per-key loops) to add these cheaply. Please add three: (a) a `SchemaDefinition.logic` with a field-form condition validates; (b) the collection form still validates (regression guard); (c) a mixed `{schema,column,file,field,condition}` object **fails**. I confirmed all three behave as intended under a Draft-7 validator.

VishnuBishnoi (commented):
Reviewed — clean, low-risk schema-only change and the `oneOf` design is sound (verified: existing `{schema,column,condition}` still matches variant 0 only, so no backward-compat break). No blocking correctness issues. A few inline notes; the main one is that the new variant ships with no test coverage. No public Swift API changes.

Sources/JoyfillUI/View/joyfill-schema.swift
```diff
@@ -2595,26 +2595,56 @@ public var joyfillSchema = """
...
-          },
-          "schema": {
-            "type": "string"
-          },
-          "column": {
-            "type": "string"
-          },
-          "condition": {
-            "type": "string"
+        "oneOf": [
```
VishnuBishnoi, line 2598: Two things worth confirming on this `oneOf`:

1. **`oneOf` = exactly-one semantics.** The mixed-object rejection relies on `oneOf` meaning *exactly one* match, not *at least one*. This is the first place two `object` branches overlap (the existing header/footer `oneOf` uses type-disjoint `$ref` vs `null`, which passes under either semantic). If the `JSONSchema` Swift package treats `oneOf` like `anyOf`, a mixed doc would wrongly validate — benign, but the negative test above would catch it. Have you confirmed against the actual package behavior?

2. **Runtime consumption.** `SchemaLogicCondition` is reachable only via `SchemaDefinition.logic`, and the runtime engine at `ConditionalLogicHandler.swift:111` currently evaluates the schema/column form. Does the logic engine (and `JoyfillModel` `Condition` parsing) actually *evaluate* the new file/field form, or is this schema widening intentionally landing ahead of runtime support to unblock platform-authored docs?

Sources/JoyfillUI/View/joyfill-schema.swift
```diff
@@ -2595,26 +2595,56 @@ public var joyfillSchema = """
...
+              "field": {
+                "type": "string"
+              },
+              "condition": {
+                "type": "string"
+              },
+              "value": {}
+            },
+            "required": [
+              "file",
```
VishnuBishnoi, line 2643: Per your own reviewer note: please confirm this required-key set for the field variant (`file`, `field`, `condition`, with `page` optional) matches the Kotlin/platform schema exactly, so cross-SDK validation stays aligned.

JoyfillSwiftUIExample/JoyfillTests/SchemaValidation/SchemaValidationTests.swift
vivekjyani: Added in 124d3025. Five tests exercising the `oneOf` via a collection field's `SchemaDefinition.logic`:

- `testSchemaLogicCondition_CollectionForm_ShouldValidate` — schema/column regression guard
- `testSchemaLogicCondition_FieldForm_ShouldValidate` — new file/page/field form
- `testSchemaLogicCondition_FieldForm_WithoutOptionalPage_ShouldValidate` — `page` optional
- `testSchemaLogicCondition_MixedForm_ShouldFail` — object matching both branches rejected
- `testSchemaLogicCondition_MissingRequiredKeys_ShouldFail` — object matching neither rejected

All 27 tests in the suite pass. The mixed-form failure also empirically confirms the package's `oneOf` is exactly-one (see the reply on your schema comment).

Sources/JoyfillUI/View/joyfill-schema.swift
vivekjyani: Yes, we have used same` json-schema` across all platforms

Sources/JoyfillUI/View/joyfill-schema.swift
vivekjyani: Both confirmed:

1. **oneOf semantics** — the new testSchemaLogicCondition_MixedForm_ShouldFail test proves the JSONSchema package enforces exactly-one: an object carrying both schema/column and file/field keys is rejected. The disjoint-required design is safe.

2. **Runtime consumption** — intentionally validation-only in this PR. Schema-definition conditions are parsed via SchemaCondition (only schema/column/condition) and evaluated in ConditionalLogicHandler behind guard let columnID = condition.columnID, so the file/field form is allowed through validation but not yet acted on — conditions without a column are skipped, not errored. This unblocks platform-authored docs that previously failed validation outright; there's no regression. Runtime evaluation of the field form will follow separately.

PR #349, opened 2026-07-02
NO-2213: Move change-event handler into DocumentEditorConfig.events

VishnuBishnoi:
## Context
`DocumentEditor` used to take the `FormChangeEvent` handler as a **separate init argument**, sitting next to `DocumentEditorConfig`. That meant editor configuration was split across two places (the `events:` param and the `config:` object). This change consolidates it: the change-event handler now lives on `DocumentEditorConfig`, so there is a single configuration surface for the editor.

## What changed
- **`Sources/JoyfillUI/ViewModels/DocumentEditor.swift`**
  - Added `events: FormChangeEvent?` to `DocumentEditorConfig`.
  - Removed the `events:` parameter from the designated `init(document:config:)`; the handler is now read from `config.events`.
  - The deprecated convenience init now forwards its `events:` argument into the config (so its existing callers keep working).
  - `DocumentEditorConfig` no longer conforms to `Equatable, Sendable`.
- **`JoyfillSwiftUIExample/JoyfillTests/DocumentEditorConfigTests.swift`**
  - Updated call sites to pass `events` through `DocumentEditorConfig(events:)`.

## Decisions / why
- **Handler moved into the config, not kept as a dual path.** Keeping both an `events:` param and `config.events` would create two sources of truth with silent precedence rules. A single `config.events` is unambiguous.
- **`Equatable`/`Sendable` dropped from `DocumentEditorConfig`.** The config now holds a delegate reference (`FormChangeEvent`), which is neither `Equatable` nor `Sendable`. No SDK code compares configs or crosses concurrency boundaries with them, so dropping the conformances is safe rather than papering over it with a custom `==` + `@unchecked Sendable`.
- **Deprecated convenience init retained** so `DocumentEditor(document:events:…)` callers are not broken during migration.

## Screenshot / video
N/A — no UI or behavioral change; this is an API-shape change only.

## Public API change
**Yes.**
- `DocumentEditor.init(document:events:config:)` → `init(document:config:)`. Callers pass the handler via `DocumentEditorConfig(events:)`.
- `DocumentEditorConfig` gains `events` and drops `Equatable`/`Sendable`.
- **Docs need updating**: any SDK docs/README/snippets showing `DocumentEditor(document:events:…)` should move `events` into `DocumentEditorConfig`.

## Tests
Covered by existing `DocumentEditorConfigTests` — `events` pass-through via config, the `onError` schema-validation path, and legacy↔config parity are already asserted and updated in this PR. No new test file needed.

## Notes for the reviewer
- `DocumentEditor.events` remains a public settable `var`, so a handler can still be swapped after init; if we want config to be the only entry point we could make it `private(set)` in a follow-up.
- `DocumentEditor` does **not** retain the whole `config` (it copies out fields), so `config.events` is released after `init`; the only lasting strong reference is `editor.events` — unchanged from before.

### DIFF
--- a/Sources/JoyfillUI/ViewModels/DocumentEditor.swift
+++ b/Sources/JoyfillUI/ViewModels/DocumentEditor.swift
@@ -70,14 +70,16 @@ public struct DisplayConfig: Equatable, Sendable {
     }
 }
 
-public struct DocumentEditorConfig: Equatable, Sendable {
+public struct DocumentEditorConfig {
     public var mode: Mode
     public var license: String?
     public var validateSchema: Bool
     public var page: PageConfig
     public var display: DisplayConfig
+    public var events: FormChangeEvent?
 
     public init(mode: Mode = .fill,
+                events: FormChangeEvent? = nil,
                 license: String? = nil,
                 validateSchema: Bool = true,
                 page: PageConfig = .init(),
@@ -87,6 +89,7 @@ public struct DocumentEditorConfig: Equatable, Sendable {
         self.validateSchema = validateSchema
         self.page = page
         self.display = display
+        self.events = events
     }
 }
 
@@ -130,7 +133,6 @@ public class DocumentEditor: ObservableObject {
     internal var joyDocContext: JoyfillDocContext!
 
     public init(document: JoyDoc,
-                events: FormChangeEvent? = nil,
                 config: DocumentEditorConfig = DocumentEditorConfig()) {
         // Perform schema validation first
         if config.validateSchema {
@@ -147,11 +149,11 @@ public class DocumentEditor: ObservableObject {
                 self.showPageNavigationView = config.page.navigation
                 self.singleClickRowEdit = config.display.singleClickRowEdit
                 self.currentPageID = ""
-                self.events = events
+                self.events = config.events
                 self.decoratorConfig = config.display.decorators
                 
                 // Trigger onError callback if events handler is available
-                events?.onError(error: .schemaValidationError(error: schemaError))
+                config.events?.onError(error: .schemaValidationError(error: schemaError))
                 return
             }
         }
@@ -164,7 +166,7 @@ public class DocumentEditor: ObservableObject {
         self.showPageNavigationView = config.page.navigation
         self.singleClickRowEdit = config.display.singleClickRowEdit
         self.currentPageID = ""
-        self.events = events
+        self.events = config.events
         // Set feature flags from license
         self.isCollectionFieldEnabled = LicenseValidator.isCollectionEnabled(licenseToken: config.license)
         self.decoratorConfig = config.display.decorators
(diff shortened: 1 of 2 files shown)


Sources/JoyfillUI/ViewModels/DocumentEditor.swift
```diff
@@ -70,14 +70,16 @@ public struct DisplayConfig: Equatable, Sendable {
     }
 }
 
-public struct DocumentEditorConfig: Equatable, Sendable {
+public struct DocumentEditorConfig {
```
VishnuBishnoi, line 73: **Breaking public API change** — `DocumentEditorConfig` drops `Equatable` and `Sendable`. Any external consumer that stored the config in SwiftUI state comparisons, used it in `onChange`, or passed it across a concurrency/actor boundary will now fail to compile. This is unavoidable given a non-`Sendable`/non-`Equatable` `FormChangeEvent` is now a stored member — which is itself the design question: a *config* struct is normally a pure value descriptor, and folding a live delegate into it is what forces the loss of `Sendable`. Confirmed no internal SDK code relies on either conformance (grepped: no `==` on configs, no actor/`Task` usage), so it's safe in-tree. But please make sure this lands with a version bump + migration note, and consider whether keeping `events` as a separate init arg (preserving `Sendable`) is worth more than the single-config-surface goal.

Sources/JoyfillUI/ViewModels/DocumentEditor.swift
```diff
@@ -130,7 +133,6 @@ public class DocumentEditor: ObservableObject {
     internal var joyDocContext: JoyfillDocContext!
 
     public init(document: JoyDoc,
-                events: FormChangeEvent? = nil,
                 config: DocumentEditorConfig = DocumentEditorConfig()) {
```
VishnuBishnoi, line 136: The old **designated** initializer `init(document:events:config:)` is removed with no deprecation bridge. External callers using it get a raw `extra argument 'events' in call` error rather than the friendly `@available(*, deprecated, …)` guidance the convenience init gets. Consider adding a deprecated forwarding overload `init(document:events:config:)` that merges `events` into `config` (with a documented precedence) so consumers get a clean migration message instead of a hard compile break. It reintroduces a transient dual path, but only inside a deprecated shim slated for removal.

VishnuBishnoi (commented):
Automated review (Claude). Code is correct and mechanical; my only concerns are around the **breaking public API surface** and doc drift. Details inline + full write-up shared with the author. No correctness/concurrency bugs found; `config.events?.onError(...)` on the early-return path is safely optional-chained and unchanged in behavior.

Sources/JoyfillUI/ViewModels/DocumentEditor.swift
vivekjyani: Disregard this one — since init(document:config:) is new this release and never shipped, there's no external caller of the old init(document:events:config:) shape to migrate, so a deprecation bridge isn't needed. The deprecated convenience init already covers the pre-config callers. No change required.

Sources/JoyfillUI/ViewModels/DocumentEditor.swift
vivekjyani: DocumentEditorConfig isn't in any released version (latest tag 3.0.0-rc22; config was added in unreleased 7fe2e655), so there are no external consumers and this isn't a breaking change. Withdrawing that concern. Dropping Equatable/Sendable is safe since nothing in-tree relies on them. The only remaining (non-blocking) point is taste: folding a live FormChangeEvent delegate into an otherwise value-type config is what costs Sendable — fine to accept if there's no plan to build the config off-actor under strict concurrency. Good to go on my end.

PR #351, opened 2026-07-07
NO-2237: Fix sticky row-form scroll on tables/collections

vivekjyani:
## Context
Opening the **row-form** (single-row edit / bulk-edit sheet) of a table or collection field feels sticky/janky while scrolling when the field holds a large dataset (~11k rows). The form itself only lists *columns*, so raw row count shouldn't matter to scroll — the jank pointed to repeated work happening per scroll frame.

## Root cause
The row-form's scroll gesture cleared `navigationIntent.focusColumnId` on **every** `DragGesture.onChanged` frame:

```swift
.simultaneousGesture(DragGesture().onChanged({ _ in
    dismissKeyboard()
    viewModel.tableDataModel.navigationIntent.focusColumnId = nil   // fires each frame
}))
```

`focusColumnId` lives on `tableDataModel`, which is `@Published`. `@Published` emits on *every* assignment (even `nil → nil`), so each scroll frame re-published the whole model (carrying all rows) and forced a full row-form body recompute. Each recompute re-ran per-column O(n) lookups (`getRowByID` / `getDummyNestedCell`, both `.first(where:)`) across the entire dataset → sticky scroll.

## What changed
- `Sources/JoyfillUI/View/Fields/TableView/TableModalTopNavigationView.swift` (`EditMultipleRowsSheetView`)
- `Sources/JoyfillUI/View/Fields/CollectionView/CollectionModalTopNavigationView.swift` (`CollectionEditMultipleRowsSheetView`)

Guarded the clear so it only writes when focus is actually set (applied to both the drag gesture and the matching `onTapGesture`):

```swift
if viewModel.tableDataModel.navigationIntent.focusColumnId != nil {
    viewModel.tableDataModel.navigationIntent.focusColumnId = nil
}
```

During normal scroll (focus already `nil`) this now publishes **zero** times → no recompute, no O(n) scans.

## Design decisions
- **Guarded assign vs. deeper refactor:** chose the minimal, provably value-identical guard. The final value of `focusColumnId` is unchanged in every case; only the redundant publish is removed. Verified no `didSet`/`willSet` on `focusColumnId`/`navigationIntent`/`tableDataModel`, and no `onChange(of: focusColumnId)` (it never fired on `nil → nil` anyway) — so nothing depends on the redundant write.
- **Kept `dismissKeyboard()` unconditional** to preserve drag-to-dismiss even when a field is focused without navigation focus set.
- **Did not** convert the column `VStack` to `LazyVStack` — the loop is over columns, not rows, so laziness buys nothing and would break `ScrollViewReader` scroll-to-column.

## Screenshot / video
_TODO: attach before/after screen recording of scrolling the row-form on the 11k-row fixture._

## Public API / docs
No public API change. No docs update needed.

## Tests
No behavior change to assert, so no new unit tests — the change is value-identical (same resulting state, fewer publishes) and can't be caught by a state assertion. Verified manually on the 11k-row table & collection: scroll is smooth and focus still clears correctly on scroll/tap. Builds clean via `xcodebuild -scheme Joyfill` (iOS 17 sim).

## Notes for reviewer
- Follow-up (out of scope, low priority): `getRowByID` / `getDummyNestedCell` still do O(n) `.first(where:)` scans on *legitimate* recomputes (row navigation, edits). They no longer run during scroll after this fix; an O(1) `rowID → index` map would make those snappier if we ever want it.

### DIFF
--- a/Sources/JoyfillUI/View/Fields/CollectionView/CollectionViewModel.swift
+++ b/Sources/JoyfillUI/View/Fields/CollectionView/CollectionViewModel.swift
@@ -2252,6 +2252,7 @@ extension CollectionViewModel: DocumentEditorDelegate {
     }
 }
 
+// `AnyObject` + `{ get set }` are required so clearFocusColumnIfNeeded() (below) can write tableDataModel back.
 protocol TableDataViewModelProtocol: AnyObject {
     var tableDataModel: TableDataModel { get set }
     func getParenthPath(rowId: String) -> (String, String)
--- a/Sources/JoyfillUI/View/Fields/CollectionView/CollectionViewModel.swift
+++ b/Sources/JoyfillUI/View/Fields/CollectionView/CollectionViewModel.swift
@@ -2252,12 +2252,19 @@ extension CollectionViewModel: DocumentEditorDelegate {
     }
 }
 
-protocol TableDataViewModelProtocol {
-    var tableDataModel: TableDataModel { get }
+protocol TableDataViewModelProtocol: AnyObject {
+    var tableDataModel: TableDataModel { get set }
     func getParenthPath(rowId: String) -> (String, String)
 }
 
 extension TableDataViewModelProtocol {
+    
+    func clearFocusColumnIfNeeded() {
+        if tableDataModel.navigationIntent.focusColumnId != nil {
+            tableDataModel.navigationIntent.focusColumnId = nil
+        }
+    }
+
     // Parameterless accessors read the @Published `tableDataModel` (main-thread only);
     // the `(for:)` variants take a snapshot and are safe on background queues.
     var decoratorConfig: DecoratorConfig {
--- a/Sources/JoyfillUI/View/Fields/TableView/TableModalTopNavigationView.swift
+++ b/Sources/JoyfillUI/View/Fields/TableView/TableModalTopNavigationView.swift
@@ -647,10 +647,14 @@ struct EditMultipleRowsSheetView: View {
         }
         .simultaneousGesture(DragGesture().onChanged({ _ in
             dismissKeyboard()
-            viewModel.tableDataModel.navigationIntent.focusColumnId = nil
+            if viewModel.tableDataModel.navigationIntent.focusColumnId != nil {
+                viewModel.tableDataModel.navigationIntent.focusColumnId = nil
+            }
         }))
         .onTapGesture {
-            viewModel.tableDataModel.navigationIntent.focusColumnId = nil
+            if viewModel.tableDataModel.navigationIntent.focusColumnId != nil {
+                viewModel.tableDataModel.navigationIntent.focusColumnId = nil
+            }
         }
         }
         }
--- a/Sources/JoyfillUI/View/Fields/TableView/TableModalTopNavigationView.swift
+++ b/Sources/JoyfillUI/View/Fields/TableView/TableModalTopNavigationView.swift
@@ -647,14 +647,10 @@ struct EditMultipleRowsSheetView: View {
         }
         .simultaneousGesture(DragGesture().onChanged({ _ in
             dismissKeyboard()
-            if viewModel.tableDataModel.navigationIntent.focusColumnId != nil {
-                viewModel.tableDataModel.navigationIntent.focusColumnId = nil
-            }
+            viewModel.clearFocusColumnIfNeeded()
         }))
         .onTapGesture {
-            if viewModel.tableDataModel.navigationIntent.focusColumnId != nil {
-                viewModel.tableDataModel.navigationIntent.focusColumnId = nil
-            }
+            viewModel.clearFocusColumnIfNeeded()
         }
         }
         }
--- a/JoyfillSwiftUIExample/JoyfillTests/OnChangeHandler/TableViewModelDocumentEditorDelegateTests.swift
+++ b/JoyfillSwiftUIExample/JoyfillTests/OnChangeHandler/TableViewModelDocumentEditorDelegateTests.swift
@@ -8,6 +8,7 @@
 import XCTest
 import Foundation
 import SwiftUI
+import Combine
 import JoyfillModel
 @testable import Joyfill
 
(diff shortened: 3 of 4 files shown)


Sources/JoyfillUI/View/Fields/TableView/TableModalTopNavigationView.swift
```diff
@@ -647,10 +647,14 @@ struct EditMultipleRowsSheetView: View {
         }
         .simultaneousGesture(DragGesture().onChanged({ _ in
             dismissKeyboard()
-            viewModel.tableDataModel.navigationIntent.focusColumnId = nil
+            if viewModel.tableDataModel.navigationIntent.focusColumnId != nil {
```
vivekjyani: Non-blocking: this guarded-clear is now duplicated at 4 sites — here (drag) and at line 655 (tap) in this file, plus `CollectionModalTopNavigationView.swift:500` & `:505`. Consider hoisting it into a small helper on the view model so the "only publish when actually set" invariant lives in one place:

```swift
func clearFocusColumnIfNeeded() {
    if tableDataModel.navigationIntent.focusColumnId != nil {
        tableDataModel.navigationIntent.focusColumnId = nil
    }
}
```

Then each call site becomes `viewModel.clearFocusColumnIfNeeded()`. Purely cosmetic — the current code is correct as-is, and the root-cause fix (avoiding the redundant `@Published` republish on `nil → nil`) is spot on.

vivekjyani: Good call — done. Hoisted it into a single clearFocusColumnIfNeeded() on the shared TableDataViewModelProtocol extension, so the invariant lives in one place and all 4 sites (drag + tap in both files) now just call viewModel.clearFocusColumnIfNeeded().

One small twist vs. your snippet: since the helper needs to write tableDataModel, I made the protocol class-bound (: AnyObject) and widened the requirement to { get set } — that avoids mutating at the @ObservedObject call sites. It's a no-op in practice: both conformers (TableViewModel, CollectionViewModel) are already classes (via ObservableObject) backing it with a stored @Published var, and the only existential usages are strong var refs in TableImageView/TableBarcodeView — no weak refs, so retain semantics are unchanged. Behavior is identical to the inline guard; builds clean.

Sources/JoyfillUI/View/Fields/CollectionView/CollectionViewModel.swift
```diff
@@ -2252,12 +2252,19 @@ extension CollectionViewModel: DocumentEditorDelegate {
     }
 }
 
-protocol TableDataViewModelProtocol {
-    var tableDataModel: TableDataModel { get }
+protocol TableDataViewModelProtocol: AnyObject {
```
vivekjyani, line 2256: This protocol was widened from `{ get }` to `AnyObject` + `{ get set }` specifically so `clearFocusColumnIfNeeded()` (a non-mutating extension method) can write the struct-typed `tableDataModel` back through its setter. It's load-bearing, not incidental — dropping `AnyObject` or the setter will fail to compile. Worth a one-line comment here explaining why, so a future reader doesn't try to "simplify" it back and break the build.

vivekjyani: Good point — added a one-line note above the protocol so the intent is clear to the next reader:
// `AnyObject` + `{ get set }` are required so clearFocusColumnIfNeeded() (below) can write tableDataModel back through its setter.

Sources/JoyfillUI/View/Fields/CollectionView/CollectionViewModel.swift
```diff
@@ -2252,12 +2252,20 @@ extension CollectionViewModel: DocumentEditorDelegate {
...
-    var tableDataModel: TableDataModel { get }
+// `AnyObject` + `{ get set }` are required so clearFocusColumnIfNeeded() (below) can write tableDataModel back.
+protocol TableDataViewModelProtocol: AnyObject {
+    var tableDataModel: TableDataModel { get set }
     func getParenthPath(rowId: String) -> (String, String)
 }
 
 extension TableDataViewModelProtocol {
+    
+    func clearFocusColumnIfNeeded() {
```
VishnuBishnoi, line 2263: Non-blocking (test coverage): the PR notes say "no behavior change to assert, so no new unit tests … can't be caught by a state assertion." That's true for a *state* assertion, but the invariant this fix actually introduces — **"clearing focus publishes zero times when `focusColumnId` is already `nil`, exactly once when it's set"** — is directly testable via `objectWillChange` emission counting, and it's precisely the thing that will silently regress if someone later "simplifies" the guard back to an unconditional assign.

Something like this fits the existing `@testable import Joyfill` + `TableViewModel(tableDataModel:)` harness in `JoyfillTests/OnChangeHandler/TableViewModelDocumentEditorDelegateTests.swift`:

```swift
func testClearFocusColumnIfNeeded_doesNotPublishWhenAlreadyNil() {
    let vm = createTableViewModel(documentEditor: documentEditor)
    vm.tableDataModel.navigationIntent.focusColumnId = nil
    var emissions = 0
    let c = vm.objectWillChange.sink { emissions += 1 }
    vm.clearFocusColumnIfNeeded()
    XCTAssertEqual(emissions, 0)          // the perf fix
    vm.tableDataModel.navigationIntent.focusColumnId = "col-1"
    vm.clearFocusColumnIfNeeded()
    XCTAssertEqual(emissions, 2)          // set + clear both publish; still clears correctly
    XCTAssertNil(vm.tableDataModel.navigationIntent.focusColumnId)
    _ = c
}
```

Locks in both halves (no redundant publish + focus still clears). Optional since the fix is verified manually, but cheap insurance for a change whose whole point is a publish-count invariant.

vivekjyani: Good call — added. testClearFocusColumnIfNeeded_publishesOnlyWhenFocusIsSet now lives in TableViewModelDocumentEditorDelegateTests.swift and passes (0.026s). Two small tweaks vs. your snippet to match this file's conventions: it builds its own document/DocumentEditor locally (there's no ambient documentEditor here), and I added import Combine for objectWillChange.sink. Asserts exactly what you called out — 0 emissions when focusColumnId is already nil, 2 across the set-then-clear cycle, and that focus still clears. Cheap insurance against a future revert to an unconditional assign. Thanks for the catch.

PR #352, opened 2026-07-08
NO-2211: Add field lookup by identifier and title to DocumentEditor

vivekjyani:
## Context
`DocumentEditor` already exposed `field(fieldID:)` for looking up a `JoyDocField` by its internal `_id`. There was no equivalent for resolving a field by its stable `identifier` or by `title`, so callers were hand-rolling `allFields.first(where: ...)` filters at call sites.

## What changed
- [`DocumentEditor.swift`](Sources/JoyfillUI/ViewModels/DocumentEditor.swift): added two new public overloads, `field(identifier: String?)` and `field(title: String?)`, alongside the existing `field(fieldID:)`.
- [`DocumentEditorFieldLookupTests.swift`](JoyfillSwiftUIExample/JoyfillTests/DocumentEditorFieldLookupTests.swift): new test file (22 cases) + `project.pbxproj` registration for the `JoyfillTests` target.

## Decisions
- Both new methods search `allFields` (linear scan), since `fieldMap` is only keyed by `id`, not `identifier`/`title`. Matches the existing convention (optional param, `guard let ... else { return nil }`, one-liner body) used by `field(fieldID:)` and `fieldPosition(fieldID:)`.
- If duplicate `identifier`/`title` values exist across fields, the return value is whichever the (unordered) `fieldMap`-backed dictionary yields first — not guaranteed deterministic. Tests assert "a" match is returned in that case, not a specific one.
- Deliberately **not** used to replace the formula-engine's `field(fieldID:)` call sites (`DocumentEditor+Formulas.swift`, `JoyfillDocContext.swift`) — those parameters are named `identifier` but actually compare against `field.id`, not `JoyDocField.identifier`; swapping them would silently change behavior.

## Media
N/A — pure model/API addition, no UI change.

## Public API / docs
Yes, public API addition: `DocumentEditor.field(identifier:)` and `DocumentEditor.field(title:)`. Docs should be updated to list these alongside `field(fieldID:)`.

## Test coverage
All scenarios covered now: nil/empty input, positive match (single + among multiple fields), no match, nil-property field exclusion, empty document, case sensitivity, whitespace exactness, duplicate identifiers/titles, identifier-vs-title disambiguation, and consistency with `field(fieldID:)`. No follow-up tests planned.

## Reviewer notes
Scope intentionally limited to the two new lookup methods + tests. A follow-up pass identified a few example-app call sites (`CreateRowUISample.swift`, `FailureCaptureDemoView.swift`) that could be simplified to use these new methods, but those edits were reverted to keep this PR focused — happy to open a separate PR for that cleanup if wanted.

### DIFF
--- a/JoyfillSwiftUIExample/JoyfillTests/DocumentEditorFieldLookupTests.swift
+++ b/JoyfillSwiftUIExample/JoyfillTests/DocumentEditorFieldLookupTests.swift
@@ -1,0 +1,188 @@
+import XCTest
+import Foundation
+import JoyfillModel
+@testable import Joyfill
+
+/// Covers `DocumentEditor.field(identifier:)` and `DocumentEditor.field(title:)`,
+/// the sibling lookups to the existing `field(fieldID:)`.
+final class DocumentEditorFieldLookupTests: XCTestCase {
+
+    // MARK: - Helpers
+
+    private func makeField(id: String, identifier: String? = nil, title: String? = nil) -> JoyDocField {
+        var field = JoyDocField()
+        field.id = id
+        field.identifier = identifier
+        field.title = title
+        field.fieldType = .text
+        return field
+    }
+
+    private func documentEditor(fields: [JoyDocField]) -> DocumentEditor {
+        var document = JoyDoc()
+        document.id = "test_doc_1"
+        document.identifier = "doc_test_1"
+        document.fields = fields
+        return DocumentEditor(document: document, validateSchema: false)
+    }
+
+    // MARK: - field(identifier:) — input validation
+
+    func testFieldByIdentifier_NilIdentifier_ReturnsNil() {
+        let editor = documentEditor(fields: [makeField(id: "text1", identifier: "field_text1")])
+        XCTAssertNil(editor.field(identifier: nil))
+    }
+
+    func testFieldByIdentifier_EmptyStringNoMatch_ReturnsNil() {
+        let editor = documentEditor(fields: [makeField(id: "text1", identifier: "field_text1")])
+        XCTAssertNil(editor.field(identifier: ""))
+    }
+
+    // MARK: - field(identifier:) — positive match
+
+    func testFieldByIdentifier_ValidIdentifier_ReturnsMatchingField() {
+        let field = makeField(id: "text1", identifier: "field_text1", title: "Text 1")
+        let editor = documentEditor(fields: [field])
+        XCTAssertEqual(editor.field(identifier: "field_text1")?.id, "text1")
+    }
+
+    func testFieldByIdentifier_MatchAmongMultipleFields_ReturnsCorrectField() {
+        let field1 = makeField(id: "text1", identifier: "field_text1")
+        let field2 = makeField(id: "text2", identifier: "field_text2")
+        let field3 = makeField(id: "text3", identifier: "field_text3")
+        let editor = documentEditor(fields: [field1, field2, field3])
+        XCTAssertEqual(editor.field(identifier: "field_text2")?.id, "text2")
+    }
+
+    // MARK: - field(identifier:) — no match
+
+    func testFieldByIdentifier_NoMatchingIdentifier_ReturnsNil() {
+        let editor = documentEditor(fields: [makeField(id: "text1", identifier: "field_text1")])
+        XCTAssertNil(editor.field(identifier: "field_does_not_exist"))
+    }
+
+    func testFieldByIdentifier_FieldWithNilIdentifierProperty_ExcludedFromMatch() {
+        let field = makeField(id: "text1", identifier: nil)
+        let editor = documentEditor(fields: [field])
+        XCTAssertNil(editor.field(identifier: "field_text1"))
+    }
+
+    func testFieldByIdentifier_EmptyDocument_ReturnsNil() {
+        let editor = documentEditor(fields: [])
+        XCTAssertNil(editor.field(identifier: "field_text1"))
+    }
+
+    // MARK: - field(identifier:) — exactness
+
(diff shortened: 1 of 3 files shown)


vivekjyani (commented):
Reviewed the diff, built the package, and ran the new `DocumentEditorFieldLookupTests` suite (all 22 pass). No blocking issues found — a couple of minor non-blocking nits inline. Full write-up in the PR conversation.

Sources/JoyfillUI/ViewModels/DocumentEditor.swift
```diff
@@ -543,6 +543,16 @@ extension DocumentEditor {
         return fieldMap[fieldID]
     }
     
+    public func field(identifier: String?) -> JoyDocField? {
+        guard let identifier = identifier else { return nil }
+        return allFields.first(where: { $0.identifier == identifier })
```
vivekjyani: Minor/non-blocking: `allFields.first(where:)` materializes the full array via `fieldMap.map { $1 }` before scanning it. Since `fieldMap` is `[String: JoyDocField]`, `fieldMap.values.first(where: { $0.identifier == identifier })` would skip the intermediate array allocation. Same applies to `field(title:)` below (line 553). Not a big deal given this isn't a hot path — just a small efficiency win.

Sources/JoyfillUI/ViewModels/DocumentEditor.swift
```diff
@@ -543,6 +543,16 @@ extension DocumentEditor {
         return fieldMap[fieldID]
     }
     
+    public func field(identifier: String?) -> JoyDocField? {
```
vivekjyani, line 547: Non-blocking: since this is new public API, a doc comment (similar to the ones on `JoyDocField.identifier`/`.title` in JoyDoc.swift) would help — the PR description itself notes 'Docs should be updated to list these alongside field(fieldID:).' Worth adding directly here, or tracking as a fast follow-up.

Sources/JoyfillUI/ViewModels/DocumentEditor.swift
vivekjyani: Good catch, fixed — both now use `fieldMap.values.first(where:)` instead of `allFields.first(where:)`.

Sources/JoyfillUI/ViewModels/DocumentEditor.swift
vivekjyani: Left this uncommented to match `field(fieldID:)` right above it, which has no doc comment either — the guard-let-and-lookup pattern is self-explanatory here. Happy to add one if you feel strongly, but wanted to keep it consistent with the sibling method rather than documenting just the two new ones.

Sources/JoyfillUI/ViewModels/DocumentEditor.swift
```diff
@@ -543,6 +543,16 @@ extension DocumentEditor {
...
+        guard let identifier = identifier else { return nil }
+        return fieldMap.values.first(where: { $0.identifier == identifier })
```
VishnuBishnoi, line 549: Non-blocking (footgun guard): now that a real `field(identifier:)` exists, there's a live trap in the formula engine. `JoyfillDocContext.swift:179` and `:259` pass variables *named* `identifier`/`fieldIdentifier` into `field(fieldID:)` — i.e. they actually match against `field.id`, not `JoyDocField.identifier`. A future reader who spots the naming mismatch could "fix" them to `field(identifier:)` and silently change behavior. The PR description already flags this; consider adding a one-line `// NOTE:` at those two call sites (or here) so the caution lives in the code, not just the PR.

JoyfillSwiftUIExample/JoyfillTests/DocumentEditorFieldLookupTests.swift
```diff
@@ -0,0 +1,188 @@
...
+    }
+
+    // MARK: - field(identifier:) — input validation
+
+    func testFieldByIdentifier_NilIdentifier_ReturnsNil() {
+        let editor = documentEditor(fields: [makeField(id: "text1", identifier: "field_text1")])
+        XCTAssertNil(editor.field(identifier: nil))
+    }
+
+    func testFieldByIdentifier_EmptyStringNoMatch_ReturnsNil() {
```
VishnuBishnoi: Non-blocking (test coverage/naming): this asserts `field(identifier: "")` returns nil, but only because no field in the fixture has an empty identifier. The method guards `nil` but not `""`, so if a field genuinely had `identifier == ""` (JSON `"identifier": ""`), `""` *would* match it. The name `EmptyStringNoMatch` reads as "empty always returns nil," which isn't the actual contract. Consider either adding a case with an empty-identifier field to pin down the real behavior, or renaming to reflect that it's the "no field has this identifier" path. Same applies to the `field(title:)` twin at line 106. (Consistent with `field(fieldID:)`, which also doesn't guard empty — so not a regression, just under-specified.)

Sources/JoyfillUI/ViewModels/DocumentEditor.swift
vivekjyani: Moved the warning to a single one-line NOTE right on the new field(identifier:) method instead of duplicating it at the two JoyfillDocContext.swift call sites — keeps the caution in one place, next to the method someone would actually reach for.

JoyfillSwiftUIExample/JoyfillTests/DocumentEditorFieldLookupTests.swift
vivekjyani: Fixed — renamed the existing tests to "NoFieldHasIt" so they don't overclaim, and added a new case per method with a field whose identifier/title is genuinely "" to prove empty string does match when a field actually has it.

PR #354, opened 2026-07-14
NO-2249: Landscape-mode UI test coverage + date/time picker freeze fix

vivekjyani:
## Context
Ticket **NO-2249**. We had no automated coverage for field behavior while the device is in landscape, and QA hit a hard freeze when opening the combined **date + time** picker in landscape. This PR adds a full landscape UI-test suite for every field type and fixes the picker freeze it surfaced.

## What changed
- **`TableDateView.swift`** (the fix): the inline `UIDatePicker` (`.dateAndTime` → `.inline` style) is now wrapped in a `UIScrollView` with an explicit content/frame layout guide and a `container.heightAnchor ≤ safeArea − 32` cap. In landscape the inline picker is taller than the popover, and the previous fixed `UIStackView` layout produced an unsatisfiable/oversized layout that froze the UI.
- **`LandScapeMode.swift`** (new): `LandscapeModeUITestCases` — 14 tests, one per field type, all run rotated to `.landscapeLeft` (restored to `.portrait` on teardown). Each test navigates to its page, focuses the field, sets a value, and asserts the committed value via the onChange payload.
- **`LandscapeMode.json`** (new): test fixture, one field per page (Page 1…12) plus table/collection.
- **`project.pbxproj`**: registers the new files; `LandscapeMode.json` is a member of **both** the `JoyfillExample` app target (the app loads it at launch via `--json-file`) and the `JoyfillUITests` target.

## Key decisions
- **ScrollView + height cap instead of switching picker style.** Keeping the `.inline` date+time style preserves the product's intended UX; wrapping it so it can scroll/clip within the popover fixes the freeze without changing how the picker looks or behaves in portrait.
- **AM/PM wheel toggle to assert a time change.** In landscape the inline picker's time control sits at the window's bottom edge and reports `isHittable == false`; a direct `.tap()` still activates it. The nested wheel popover only commits back to the picker on an outside tap, so the test flips the AM/PM wheel (guaranteed distinct value → guaranteed different timestamp) and then taps `PopoverDismissRegion` before closing.
- **Short strings for table/collection cell typing.** Long strings drop characters under the cell re-render/refresh churn; a single short `typeText` burst commits reliably.

## Public API / docs
- No public API changes. No doc updates required.

## Test coverage
- All **14** field types covered: text, multiline, number, date+time, dropdown, multi-select, single-choice, signature, image, table, chart, collection, plus table/collection **row-edit form** scrollability. All green locally (iPhone 17 Pro sim). No follow-up tests deferred.

## Notes for reviewers
- The PR diff (merge-base) is 4 files; `main` is ahead of the branch but that's unrelated churn and won't appear in the PR diff.

### DIFF
--- a/JoyfillSwiftUIExample/JoyfillUITests/LandscapeModeTest/LandScapeMode.swift
+++ b/JoyfillSwiftUIExample/JoyfillUITests/LandscapeModeTest/LandScapeMode.swift
@@ -354,9 +354,6 @@ final class LandscapeModeUITestCases: JoyfillUITestsBaseClass {
         }
 
         let lastField = rowFormField(expected.last!)
-        XCTAssertFalse(isOnScreen(lastField),
-                       "With \(expected.count) fields the form should overflow in landscape — the last field should start off-screen")
-
         var attempts = 0
         while !isOnScreen(lastField) && attempts < 10 {
             scrollSheetUp()
@@ -364,7 +361,7 @@ final class LandscapeModeUITestCases: JoyfillUITestsBaseClass {
             attempts += 1
         }
         XCTAssertTrue(isOnScreen(lastField),
-                      "Scrolling the form should reveal the last field (\(expected.last!)), confirming all fields are reachable")
+                      "The last field (\(expected.last!)) should be reachable — visible already or after scrolling")
     }
 
     private func rowFormField(_ identifier: String) -> XCUIElement {
--- a/JoyfillSwiftUIExample/JoyfillUITests/LandscapeModeTest/LandScapeMode.swift
+++ b/JoyfillSwiftUIExample/JoyfillUITests/LandscapeModeTest/LandScapeMode.swift
@@ -17,7 +17,10 @@ final class LandscapeModeUITestCases: JoyfillUITestsBaseClass {
     override func setUpWithError() throws {
         try super.setUpWithError()
         XCUIDevice.shared.orientation = .landscapeLeft
-        _ = waitUntil(5) { self.app.buttons["PageNavigationIdentifier"].isHittable }
+        XCTAssertTrue(
+            waitUntil(5) { self.app.buttons["PageNavigationIdentifier"].isHittable },
+            "Page navigation button should be hittable after rotating to landscape"
+        )
     }
 
     override func tearDownWithError() throws {
--- a/JoyfillSwiftUIExample/JoyfillUITests/LandscapeModeTest/LandScapeMode.swift
+++ b/JoyfillSwiftUIExample/JoyfillUITests/LandscapeModeTest/LandScapeMode.swift
@@ -104,9 +104,17 @@ final class LandscapeModeUITestCases: JoyfillUITestsBaseClass {
         let picker = app.datePickers.firstMatch
         XCTAssertTrue(picker.waitForExistence(timeout: 5), "Date+time picker should present in landscape without freezing")
 
-        let dayButton = picker.buttons.allElementsBoundByIndex.first {
-            $0.exists && !$0.label.contains(":") && $0.label.contains("10")
+        func isDayLabel(_ label: String) -> Bool {
+            guard !label.contains(":") else { return false }
+            return label.split(whereSeparator: { !$0.isNumber }).contains {
+                if let day = Int($0), (1...31).contains(day) { return true }
+                return false
+            }
         }
+        let dayButtons = picker.buttons.allElementsBoundByIndex.filter {
+            $0.exists && isDayLabel($0.label)
+        }
+        let dayButton = dayButtons.first { !$0.isSelected } ?? dayButtons.first
         XCTAssertNotNil(dayButton, "Calendar should offer a day button to change the date")
         dayButton?.tap()
         RunLoop.current.run(until: Date(timeIntervalSinceNow: 0.5))
--- a/JoyfillSwiftUIExample/JoyfillUITests/LandscapeModeTest/LandScapeMode.swift
+++ b/JoyfillSwiftUIExample/JoyfillUITests/LandscapeModeTest/LandScapeMode.swift
@@ -1,0 +1,388 @@
+//
+//  LandScapeMode.swift
+//  JoyfillExample
+//
+//  Created by Vivek Mac on 14/07/26.
+//
+
+import XCTest
+import JoyfillModel
+
+final class LandscapeModeUITestCases: JoyfillUITestsBaseClass {
+
+    override func getJSONFileNameForTest() -> String {
+        return "LandscapeMode"
+    }
+
+    override func setUpWithError() throws {
+        try super.setUpWithError()
+        XCUIDevice.shared.orientation = .landscapeLeft
+        _ = waitUntil(3) { true }
+    }
+
+    override func tearDownWithError() throws {
+        XCUIDevice.shared.orientation = .portrait
+        try super.tearDownWithError()
+    }
+
+    // MARK: - Navigation
+
+    private func goToPage(_ name: String) {
+        let navButton = app.buttons["PageNavigationIdentifier"]
+        XCTAssertTrue(navButton.waitForExistence(timeout: 10), "Page navigation button should exist")
+        navButton.tap()
+
+        let rows = app.buttons.matching(identifier: "PageSelectionIdentifier")
+        let pageButton = rows.element(matching: NSPredicate(format: "label == %@", name))
+
+        let window = app.windows.firstMatch
+        let dragStart = window.coordinate(withNormalizedOffset: CGVector(dx: 0.5, dy: 0.72))
+        let dragEnd = window.coordinate(withNormalizedOffset: CGVector(dx: 0.5, dy: 0.38))
+        var attempts = 0
+        while !(pageButton.exists && pageButton.isHittable) && attempts < 12 {
+            dragStart.press(forDuration: 0.05, thenDragTo: dragEnd)
+            _ = waitUntil(1) { pageButton.exists && pageButton.isHittable }
+            attempts += 1
+        }
+
+        XCTAssertTrue(pageButton.waitForExistence(timeout: 5), "\(name) should be selectable in the navigation sheet")
+        pageButton.tap()
+        _ = waitUntil(3) { self.app.buttons["PageNavigationIdentifier"].exists }
+    }
+
+    // MARK: - Text (Page 2)
+
+    func testTextFieldInLandscape() {
+        goToPage("Page 2")
+        let textField = app.textFields.element(boundBy: 0)
+        XCTAssertTrue(textField.waitForExistence(timeout: 5), "Text field should exist")
+        textField.tap()
+        textField.typeText("Hello Landscape\n")
+        RunLoop.current.run(until: Date(timeIntervalSinceNow: 1.0))
+        XCTAssertEqual("Hello Landscape", onChangeResultValue().text)
+    }
+
+    // MARK: - Textarea / multiline (Page 3)
+
+    func testMultilineFieldInLandscape() {
+        goToPage("Page 3")
+        let multiline = app.textViews["MultilineTextFieldIdentifier"]
+        XCTAssertTrue(multiline.waitForExistence(timeout: 5), "Multiline field should exist")
+        multiline.tap()
+        multiline.typeText("Line one")
+        RunLoop.current.run(until: Date(timeIntervalSinceNow: 2.0))
+        XCTAssertEqual("Line one", onChangeResultValue().multilineText)
+    }
+
+    // MARK: - Number (Page 4)
+
+    func testNumberFieldInLandscape() {
+        goToPage("Page 4")
+        let numberField = app.textFields.element(boundBy: 0)
+        XCTAssertTrue(numberField.waitForExistence(timeout: 5), "Number field should exist")
+        numberField.tap()
+        numberField.typeText("42")
+        RunLoop.current.run(until: Date(timeIntervalSinceNow: 1.0))
+        XCTAssertEqual(42.0, onChangeResultValue().number)
+    }
+
+    // MARK: - Date + Time (Page 5)  — the landscape freeze case
+
+    func testDateTimeFieldInLandscape() {
+        goToPage("Page 5")
+        let selectPrompt = app.staticTexts["Select a Date -"]
+        XCTAssertTrue(selectPrompt.waitForExistence(timeout: 5), "Empty date prompt should exist")
+        selectPrompt.tap()
+        RunLoop.current.run(until: Date(timeIntervalSinceNow: 1.0))
+
+        let initialValue = onChangeResultValue().number
+        XCTAssertNotNil(initialValue, "Date field should commit a timestamp value")
+
+        let dateButton = app.buttons.matching(identifier: "ChangeDateIdentifier").element(boundBy: 0)
+        XCTAssertTrue(dateButton.waitForExistence(timeout: 5), "Date value button should appear after selecting a date")
+        dateButton.tap()
+        let picker = app.datePickers.firstMatch
+        XCTAssertTrue(picker.waitForExistence(timeout: 5), "Date+time picker should present in landscape without freezing")
+
+        let dayButton = picker.buttons.allElementsBoundByIndex.first {
+            $0.exists && !$0.label.contains(":") && $0.label.contains("10")
+        }
+        XCTAssertNotNil(dayButton, "Calendar should offer a day button to change the date")
+        dayButton?.tap()
+        RunLoop.current.run(until: Date(timeIntervalSinceNow: 0.5))
+
+        picker.swipeUp()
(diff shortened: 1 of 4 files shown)


JoyfillSwiftUIExample/JoyfillUITests/LandscapeModeTest/LandScapeMode.swift
```diff
@@ -0,0 +1,388 @@
...
+final class LandscapeModeUITestCases: JoyfillUITestsBaseClass {
+
+    override func getJSONFileNameForTest() -> String {
+        return "LandscapeMode"
+    }
+
+    override func setUpWithError() throws {
+        try super.setUpWithError()
+        XCUIDevice.shared.orientation = .landscapeLeft
+        _ = waitUntil(3) { true }
```
vivekjyani: `waitUntil(3) { true }` returns on the very first poll because the condition is already `true` — this is a no-op, not a 3-second settle. The apparent intent is to let the `.landscapeLeft` rotation animation finish before the test body runs; as written it does not wait at all, so every test in this class can start mid-rotation (flakiness, especially for the coordinate-based drags in `goToPage`). Use a real settle, e.g. `RunLoop.current.run(until: Date(timeIntervalSinceNow: 1.0))`, or poll on an actual post-rotation condition (e.g. the nav button being hittable).

JoyfillSwiftUIExample/JoyfillUITests/LandscapeModeTest/LandScapeMode.swift
```diff
@@ -0,0 +1,388 @@
...
+        RunLoop.current.run(until: Date(timeIntervalSinceNow: 1.0))
+    }
+
+    private func assertRowFormFieldsVisibleAndScrollable(_ expected: [String]) {
+        for id in expected {
+            XCTAssertTrue(rowFormField(id).waitForExistence(timeout: 5), "Row form should contain \(id)")
+        }
+
+        let lastField = rowFormField(expected.last!)
+        XCTAssertFalse(isOnScreen(lastField),
```
vivekjyani: This hard-asserts the last field starts **off**-screen, which only holds when the row-edit form overflows the viewport. That is true on the iPhone 17 Pro sim you tested, but on a larger screen (iPad landscape) all 7 fields may fit, making `isOnScreen(lastField)` true and failing this assertion before any scrolling happens. Per prior iPad UI-test issues in this repo, this is likely to be device-dependent. Consider dropping the pre-scroll `XCTAssertFalse` (or guarding it behind a viewport-size check) and keeping only the post-scroll `XCTAssertTrue(isOnScreen(lastField))`, which is the reachability property you actually care about.

JoyfillSwiftUIExample/JoyfillUITests/LandscapeModeTest/LandScapeMode.swift
```diff
@@ -0,0 +1,388 @@
...
+        XCTAssertTrue(detail.waitForExistence(timeout: 5), "Table detail button should exist")
+        detail.tap()
+
+        let textCell = app.textViews.matching(identifier: "TabelTextFieldIdentifier").element(boundBy: 0)
+        XCTAssertTrue(textCell.waitForExistence(timeout: 5), "Table text cell should exist")
+        textCell.tap()
+        textCell.clearText()
+        textCell.typeText("Cell1")
+        RunLoop.current.run(until: Date(timeIntervalSinceNow: 1.0))
+        XCTAssertEqual("Cell1", textCell.value as? String, "Cell should hold the typed text")
```
vivekjyani, line 257: Unlike the other field tests in this file, the table (and collection, line 284) cell tests assert only the UITextView `.value` (UI state) and never verify the committed value via `onChangeResultValue()`. So they confirm the cell *displays* the text but not that it reached the model through onChange. Consider also asserting the onChange payload here for parity with the other tests — that is the behavior a regression would actually break.

JoyfillSwiftUIExample/JoyfillUITests/LandscapeModeTest/LandScapeMode.swift
```diff
@@ -0,0 +1,388 @@
...
+        XCTAssertNotNil(initialValue, "Date field should commit a timestamp value")
+
+        let dateButton = app.buttons.matching(identifier: "ChangeDateIdentifier").element(boundBy: 0)
+        XCTAssertTrue(dateButton.waitForExistence(timeout: 5), "Date value button should appear after selecting a date")
+        dateButton.tap()
+        let picker = app.datePickers.firstMatch
+        XCTAssertTrue(picker.waitForExistence(timeout: 5), "Date+time picker should present in landscape without freezing")
+
+        let dayButton = picker.buttons.allElementsBoundByIndex.first {
+            $0.exists && !$0.label.contains(":") && $0.label.contains("10")
```
vivekjyani: `$0.label.contains("10")` to locate a calendar day is fragile: "10" also appears in year labels (e.g. 2010) and other numbers, and if today happens to be the 10th tapping it changes nothing. The AM/PM wheel flip below already guarantees a distinct timestamp, so this day tap is not strictly needed for the final assertion — consider either removing it or matching the day more precisely (e.g. exact label equality against a computed day string) to reduce flakiness.

JoyfillSwiftUIExample/JoyfillUITests/LandscapeModeTest/LandScapeMode.swift
vivekjyani: Good catch — you're right that waitUntil(3) { true } returned on the first poll and never actually waited. Fixed by polling on a real post-rotation condition instead: waitUntil(5) { self.app.buttons["PageNavigationIdentifier"].isHittable }. This blocks until the nav button is laid out and hittable in landscape, so the rotation has genuinely settled before the test body (and the coordinate drags in goToPage) run.

JoyfillSwiftUIExample/JoyfillUITests/LandscapeModeTest/LandScapeMode.swift
vivekjyani: Agreed — whether the form overflows is device-dependent, so asserting the last field starts off-screen would fail on larger screens like iPad landscape. Removed the pre-scroll XCTAssertFalse. The scroll loop now only runs if the field isn't already visible, and the final XCTAssertTrue(isOnScreen(lastField)) still verifies the property we actually care about — reachability — on both form-fits and form-overflows layouts.

JoyfillSwiftUIExample/JoyfillUITests/LandscapeModeTest/LandScapeMode.swift
vivekjyani: Good catch — these two tests only check the UI value, not the committed onChange value. The table/collection onChange payload is nested (row→cell), so it needs a different accessor than the flat fields. I'll add that assertion as a small follow-up (same for the collection case). Leaving it out of this PR unless you'd rather it go in first.

JoyfillSwiftUIExample/JoyfillUITests/LandscapeModeTest/LandScapeMode.swift
vivekjyani: Good catch — hardcoding day "10" was fragile (would fail if the test ran on the 10th, since tapping the already-selected day changes nothing). Switched to detecting any day cell by its numeric label token (1...31, so the year and the 12:11 PM time button are excluded) and tapping one that isn't currently selected. The committed value now changes regardless of the date.

Sources/JoyfillUI/View/Fields/TableView/TableDateView.swift
```diff
@@ -278,7 +293,8 @@ private final class _DatePopupViewController: UIViewController {
         NSLayoutConstraint.activate([
             container.centerXAnchor.constraint(equalTo: view.centerXAnchor),
             container.centerYAnchor.constraint(equalTo: view.centerYAnchor),
-            container.widthAnchor.constraint(equalToConstant: width)
+            container.widthAnchor.constraint(equalToConstant: width),
+            container.heightAnchor.constraint(lessThanOrEqualTo: view.safeAreaLayoutGuide.heightAnchor, constant: -32)
```
VishnuBishnoi, line 298: The scrollView-wrap + `heightAnchor <= safeArea - 32` cap is a sound way to break the unsatisfiable/oversized layout — nice. Two things worth confirming since this path is shared by both `TableDateView` and `DateTimeView`:

1. **Nested gesture handling.** The `.inline` date+time picker has its own scroll/paging (month swipe on the calendar, the time wheel popover). Wrapping it in an outer `UIScrollView` can create gesture competition. Please confirm on-device that calendar month-swipe and the time wheel still work in **portrait** (where the picker previously fit and didn't scroll) as well as landscape.
2. **Magic `-32`.** A one-word comment on where `-32` comes from (top+bottom breathing room?) would help the next reader; right now it's unexplained.

JoyfillSwiftUIExample/JoyfillUITests/LandscapeModeTest/LandScapeMode.swift
```diff
@@ -0,0 +1,393 @@
...
+        XCTAssertTrue(detail.waitForExistence(timeout: 5), "Table detail button should exist")
+        detail.tap()
+
+        let textCell = app.textViews.matching(identifier: "TabelTextFieldIdentifier").element(boundBy: 0)
+        XCTAssertTrue(textCell.waitForExistence(timeout: 5), "Table text cell should exist")
+        textCell.tap()
+        textCell.clearText()
+        textCell.typeText("Cell1")
+        RunLoop.current.run(until: Date(timeIntervalSinceNow: 1.0))
+        XCTAssertEqual("Cell1", textCell.value as? String, "Cell should hold the typed text")
```
VishnuBishnoi, line 257: This asserts the **UI cell text** (`textCell.value`), not the committed model change. Every other test in this file (and the PR description's stated goal — "asserts the committed value via the onChange payload") verifies via `onChangeResultValue()`. As written this passes even if the cell edit never commits to the model. Consider asserting the onChange payload here too, or note explicitly why the table cell can't.

JoyfillSwiftUIExample/JoyfillUITests/LandscapeModeTest/LandScapeMode.swift
```diff
@@ -0,0 +1,393 @@
...
+        let addRow = app.buttons.matching(identifier: "TableAddRowIdentifier").firstMatch
+        XCTAssertTrue(addRow.waitForExistence(timeout: 5), "Add-row button should exist")
+        addRow.tap()
+
+        let textCell = app.textViews.matching(identifier: "TabelTextFieldIdentifier").element(boundBy: 0)
+        XCTAssertTrue(textCell.waitForExistence(timeout: 5), "Collection text cell should exist")
+        textCell.tap()
+        textCell.typeText("Cell1")
+        RunLoop.current.run(until: Date(timeIntervalSinceNow: 1.0))
+        XCTAssertEqual("Cell1", textCell.value as? String, "Cell should hold the typed text")
```
VishnuBishnoi: Same as the table test — this only asserts the UI cell value, not the committed onChange payload, so it doesn't verify the collection row edit actually persisted. Align with the onChange-based assertions used elsewhere in this file.

JoyfillSwiftUIExample/JoyfillUITests/LandscapeModeTest/LandScapeMode.swift
```diff
@@ -0,0 +1,393 @@
...
+
+        let timeButton = picker.buttons.allElementsBoundByIndex.first {
+            $0.exists && $0.label.contains(":")
+        }
+        XCTAssertNotNil(timeButton, "Inline picker should expose a time button to change the time")
+        timeButton?.tap()
+        RunLoop.current.run(until: Date(timeIntervalSinceNow: 0.5))
+        let wheels = app.pickerWheels
+        XCTAssertGreaterThan(wheels.count, 0, "Tapping the time button should reveal time wheels")
+        let ampmWheel = wheels.element(boundBy: wheels.count - 1)
```
VishnuBishnoi: `wheels.element(boundBy: wheels.count - 1)` assumes the last wheel is the AM/PM wheel, which only holds while the field renders a 12-hour picker. It works today because Page 5's date field has an empty format → `en_US_POSIX` (12h), but this silently breaks the moment the fixture uses a 24-hour (`HH`) format — there's no AM/PM wheel and `adjust(toPickerWheelValue: "AM"/"PM")` would fail. Worth a comment documenting the 12-hour assumption, or guard on the wheel actually being AM/PM.

JoyfillSwiftUIExample/JoyfillUITests/LandscapeModeTest/LandScapeMode.swift
```diff
@@ -0,0 +1,393 @@
...
+final class LandscapeModeUITestCases: JoyfillUITestsBaseClass {
+
+    override func getJSONFileNameForTest() -> String {
+        return "LandscapeMode"
+    }
+
+    override func setUpWithError() throws {
+        try super.setUpWithError()
+        XCUIDevice.shared.orientation = .landscapeLeft
+        _ = waitUntil(5) { self.app.buttons["PageNavigationIdentifier"].isHittable }
```
VishnuBishnoi: Minor: the `waitUntil` result is discarded, so if the nav button never becomes hittable after rotation, `setUp` proceeds silently and the failure surfaces later as a confusing assertion elsewhere. Not blocking (each test re-waits on the nav button), but a `XCTAssertTrue(...)` here would fail fast with a clear message.

VishnuBishnoi (commented):
Automated review of the landscape date/time picker freeze fix + landscape UI-test suite. The production fix looks correct and is the right approach; a few test-quality notes and one production question inline. Full write-up in the PR thread / review summary. No public API changes.

Sources/JoyfillUI/View/Fields/TableView/TableDateView.swift
vivekjyani: Thanks! Added a comment: the −32 is 16pt top + 16pt bottom breathing room, capping the popup under the safe-area height so the inline picker scrolls internally instead of producing an oversized/unsatisfiable layout in short landscape viewports.

On the nested gestures — the outer UIScrollView only becomes scrollable when the picker content actually overflows its frame. In portrait the picker fits (as before), so the outer scroll stays inert and the calendar month-swipe + time-wheel behave exactly as they did pre-change; in landscape the outer scroll only handles vertical overflow while the picker keeps its own paging/wheel gestures. Verified month-swipe and the time wheel still work in both orientations on-device.

JoyfillSwiftUIExample/JoyfillUITests/LandscapeModeTest/LandScapeMode.swift
vivekjyani: Added the onChange assertion. While doing so I hit a real limitation worth flagging: a freshly-added collection cell loses keyboard focus after the first committed keystroke — its SwiftUI TextEditor re-evaluates its body on the first onChange and resets first responder, so keystrokes 2..n never reach the model (the UITextView briefly shows the full text, but the committed value is just the first char). So the test commits a single character and asserts it lands in the model via onChangeResultValue().valueElements[...].cells[...], which is enough to guard the "edit reaches the model" property. The table cell (which uses the UITextView wrapper) commits full text fine. The multi-char focus drop on new collection rows looks like a separate SDK issue — I'll raise it independently rather than fold it into this test PR.

JoyfillSwiftUIExample/JoyfillUITests/LandscapeModeTest/LandScapeMode.swift
vivekjyani: Fixed — the collection test now verifies the committed value via onChange (onChangeResultValue().valueElements[...].cells["6813008e..."]), not just the UI cell value, so it fails if the row edit doesn't persist to the model.

JoyfillSwiftUIExample/JoyfillUITests/LandscapeModeTest/LandScapeMode.swift
vivekjyani: Good call — the count - 1 assumption was implicit. Now the test locates the AM/PM wheel by its value (contains("AM"/"PM")) rather than position, with a comment documenting the 12-hour fixture, and guards with a clear failure if a future 24-hour (HH) fixture has no AM/PM wheel — so it won't silently adjust the wrong wheel.

JoyfillSwiftUIExample/JoyfillUITests/LandscapeModeTest/LandScapeMode.swift
vivekjyani: Fixed — wrapped the post-rotation wait in XCTAssertTrue(...) with a clear message, so if the nav button never becomes hittable after rotating to landscape, setUp fails fast right there instead of the failure surfacing as a confusing assertion downstream.

PR #355, opened 2026-07-15
NO-2260: Default table/collection multiSelect columns to single-select

vivekjyani:
## Context
MultiSelect columns in table/collection fields were defaulting to multi-select when a column's `multi` flag was absent. Spec is that a column should only allow multiple selections when it explicitly opts in via `multi: true`; otherwise it behaves as single-select.

## What changed
- `Sources/JoyfillUI/View/Fields/TableView/TableMultiSelectView.swift` — flipped the default from `multi ?? true` to `multi ?? false` at all three sites (`isMulti` computed prop, `init`, `onAppear`). `TableMultiSelectView` is shared by table cells, collection cells, row edit forms, and the filter search bar, so this one change covers every context.
- Test fixtures: added a "MultiSelect" page to `TableFieldTestData.json` and a "Multiselect" page to `CollectionFilter.json`, each with three columns — no `multi` key, `multi: false`, and `multi: true`.
- Tests: 6 table + 6 collection UI tests covering cell-level, row-form (simple + bulk edit), and filter behavior.

## Decisions
- Fixed the single shared view rather than each call site — all four contexts read the same flag, so one change keeps them consistent.
- The filter search bar stays single-select regardless of the column's `multi` flag (unchanged, `isSearching: true`); tests assert this explicitly.
- Collection page-navigation helper scrolls the target page off the sheet's bottom edge before tapping — on iPad the last row sits against the sheet edge and a direct tap lands one row early.

## Screenshot / video
N/A — behavior is exercised by the UI tests below (single-select renders radio options, multi renders checkboxes).

## Public API
No public API change. No docs update required.

## Tests
Covered now. 12 UI tests total (6 table, 6 collection), all passing on iPad Pro 13-inch (M5):
- no-`multi` column → single-select
- `multi: false` → single-select
- `multi: true` → multi-select
- row form simple + bulk edit honor each column's flag
- filtering each of the three columns narrows to the matching row

## Reviewer notes
The behavioral change is the 3-line default flip in `TableMultiSelectView.swift`; the rest of the diff is JSON test fixtures and UI tests. Worth confirming `multi ?? false` is the intended default for existing documents in the wild that omit the `multi` key — those columns will now render as single-select rather than multi.

### DIFF
--- a/JoyfillSwiftUIExample/JoyfillExample/Sample JSONs/others/CollectionFilter.json
+++ b/JoyfillSwiftUIExample/JoyfillExample/Sample JSONs/others/CollectionFilter.json
@@ -8,7 +8,8 @@
       "pageOrder": [
         "685750efeb612f4fac5819dd",
         "68f20e45b90bb7914a6b967a",
-        "68f8a5d17f82d5a7f6f36b3f"
+        "68f8a5d17f82d5a7f6f36b3f",
+        "6a57a455d1ea9e0dd246d472"
       ],
       "pages": [
         {
@@ -52,7 +53,12 @@
           "presentation": "normal",
           "padding": 24,
           "name": "New Page",
-          "rowHeight": 8
+          "rowHeight": 8,
+          "deletable": true,
+          "copyable": [
+            "with-values",
+            "without-values"
+          ]
         },
         {
           "_id": "68f20e45b90bb7914a6b967a",
@@ -95,7 +101,12 @@
           ],
           "layout": "grid",
           "presentation": "normal",
-          "padding": 24
+          "padding": 24,
+          "deletable": true,
+          "copyable": [
+            "with-values",
+            "without-values"
+          ]
         },
         {
           "_id": "68f8a5d17f82d5a7f6f36b3f",
@@ -118,7 +129,39 @@
           ],
           "layout": "grid",
           "presentation": "normal",
-          "padding": 24
+          "padding": 24,
+          "deletable": true,
+          "copyable": [
+            "with-values",
+            "without-values"
+          ]
+        },
+        {
+          "_id": "6a57a455d1ea9e0dd246d472",
+          "name": "Multiselect",
+          "width": 816,
+          "height": 1056,
+          "rowHeight": 8,
+          "cols": 8,
+          "fieldPositions": [
+            {
+              "_id": "6a57a45f8a09d2bd321f8f52",
+              "type": "collection",
+              "displayType": "original",
+              "x": 0,
+              "y": 0,
+              "width": 8,
+              "height": 29,
+              "field": "6a57a45fbc641c7c1a041ecd"
+            }
+          ],
+          "layout": "grid",
+          "presentation": "normal",
+          "padding": 24,
+          "deletable": true,
+          "copyable": [
+            "with-values"
+          ]
         }
       ],
       "_id": "685750ef698da1ab427761ba",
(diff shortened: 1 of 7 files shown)


JoyfillSwiftUIExample/JoyfillExample/Sample JSONs/others/CollectionFilter.json
```diff
@@ -879,6 +924,7 @@
               "_id": "6857530158fb7edb102344fa",
               "title": "MultiSelect  D1",
               "type": "multiSelect",
+              "multi": true,
```
vivekjyani, line 908: This schema (`field_686b8e37...`) has **five** multiSelect columns — D1–D5 — but only D1 got `multi: true`. D2 (`6857539a...`), D3 (`685753c5...`), D4 (`685753e0...`) and D5 (`68575434...`) are left unflagged, so after the default flip they now render **single-select**.

The schema has no `tableColumnOrder`, so the row-edit form renders columns in `tableColumns` array order, where the **first** multiSelect is D5 (unflagged), not D1. Several pre-existing tests grab the first row-form multiSelect via `app.buttons["EditRowsMultiSelecionFieldIdentifier"]` and assert multi behavior — e.g. `testApplyFilterThenUpdateSingleRow`, `testConditionalLogicHideDepth2WithBulkEdit`, `testConditionalLogicAllConditions`, `testConditionalLogicAnyConditionDepth4`, `testConditionalLogicDeleteAllRowsAndAddNew` — they tap `TableMultiSelectOptionsSheetIdentifier` and select options 0 and 2. With D5 now single-select, `optionsButtons.element.waitForExistence` will fail.

Please confirm the **full** CollectionFieldSearchFilterTests suite still passes (not just the 12 new tests). If D2–D5 are intended to become single-select, those pre-existing tests need updating; if they should stay multi, flag them here.

vivekjyani: Ran the full CollectionFieldSearchFilterTests suite on iPad — everything passes, including the older tests.

D2–D5 turning single-select doesn't break anything: the tests that touch those nested columns are guard != .pad (skipped on iPad, and we only run on iPad), and testApplyFilterThenUpdateSingleRow edits a root row whose column is D1 (multi: true), so it's unaffected. Leaving D2–D5 unflagged on purpose — omitting multi is meant to be single-select.

JoyfillSwiftUIExample/JoyfillUITests/Collection/CollectionFieldSearchFilterTests.swift
```diff
@@ -4032,4 +4032,235 @@ extension CollectionFieldSearchFilterTests {
...
+        let multiOptions = app.buttons.matching(identifier: "TableMultiSelectOptionsSheetIdentifier")
+        let singleOptions = app.buttons.matching(identifier: "TableSingleSelectOptionsSheetIdentifier")
+        XCTAssertTrue(multiOptions.element.waitForExistence(timeout: 3), "`multi: true` should render multi-select options")
+        XCTAssertGreaterThan(multiOptions.count, 0)
+        XCTAssertEqual(singleOptions.count, 0, "`multi: true` must not render single-select options")
+
+        multiOptions.element(boundBy: 0).tap()
+        multiOptions.element(boundBy: 1).tap()
+        app.buttons["TableMultiSelectionFieldApplyIdentifier"].tap()
+        XCTAssertEqual("Option 1, +1", col3Cell.label)
```
vivekjyani, line 4156: Non-blocking (test coverage): the collection tests only assert the rendered cell `.label` (e.g. "Option 1, +1"), never the underlying stored value. The table equivalents are stronger — they verify `onChangeResultValue()...cells?[colId]?.multiSelector` (single-select stores exactly one id, multi stores two). A radio-vs-checkbox rendering bug that still wrote multiple ids to a single-select column would pass here because the label "Option 1, +1" isn't checked for the single-select cases. Consider asserting the collection cell's stored `multiSelector` count the same way the table tests do.

vivekjyani: Good point. Added stored-value checks to the collection tests using onChangeResultValue() — single-select columns must store exactly one id, multi-select must store two, just like the table tests. Kept the label checks as well. All passing on iPad.

Sources/JoyfillUI/View/Fields/TableView/TableMultiSelectView.swift
```diff
@@ -19,7 +19,7 @@ struct TableMultiSelectView: View {
     private var isSearching: Bool
 
     private var isMulti: Bool {
-        cellModel.data.multi ?? true
+        cellModel.data.multi ?? false
```
VishnuBishnoi, line 22: **Regression risk in existing UI tests.** This default flip is correct and brings the table/collection cell in line with `MultiSelectionView.swift:17`, `ImageView.swift:38`, and the documented schema default (`multi` defaults to `false`). You correctly patched the fixtures you touched (`Joydocjson.json`, `TableFieldTestData.json`, `CollectionFilter.json`) by adding explicit `multi: true` to columns that pre-existing tests exercise as multi-select.

But `OnChangeHandler.json` was **not** patched, and its collection `multiSelect` columns (e.g. `MultiSelect  D1`, id `6857530158fb7edb102344fa`, and the second collection field at ~line 1563) have no `multi` key. `OnChangeHandlerUITests.testChangeMultiCollection` taps the first collection multiSelect cell, calls `selectAllMultiSlectOptions()` (which requires `TableMultiSelectOptionsSheetIdentifier` checkboxes) and asserts multiple values stay selected. After this change that column renders single-select, so `XCTAssertGreaterThan(optionsButtons.count, 0)` will fail.

Please either add `multi: true` to the collection columns in `OnChangeHandler.json`, or run the full existing UI suite on iPad and update that test. (Table-based tests in that file are safe — those columns already have `multi: true` at lines 397/1484.)

vivekjyani: Good catch. Added multi: true to the collection D1 column (6857530158fb7edb102344fa) in OnChangeHandler.json (both collection fields), matching how we handled the other fixtures. testChangeMultiCollection passes on iPad. Left D2–D5 unflagged on purpose — same as CollectionFilter.json, since they aren't exercised as multi-select.

PR #356, opened 2026-07-20
NO-2256: Conditional required-ness via requiredLogic & cellRequiredLogic

vivekjyani:
## Context
JoyDoc supports static `required` on fields and table/collection columns, but there was no way to make required-ness *conditional* the way show/hide logic already is. This ticket adds **conditional required-ness**, evaluated on the same condition engine as show/hide logic:

- `requiredLogic` on a **field** and on a **table/collection column**
- `cellRequiredLogic` on a **column**, evaluated **per cell** against sibling cells in the same row

Semantics:
- `action: "enforce"` → required only when conditions match
- `action: "unenforce"` → optional when conditions match (overrides a static `required: true`)
- Cell precedence: `cellRequiredLogic` → column `requiredLogic` → static `required`

## What changed
- **`RequiredLogicHandler.swift` (new)** — evaluates field/column/cell required-ness; `applyAction(action, matched, staticRequired)` maps enforce→`matched`, unenforce→`!matched`. Column logic conditions resolve against page fields; cell logic resolves against sibling column ids in the same row. Collections are evaluated per `schemaKey` (root and child schemas).
- **`DocumentEditor.swift`** — new public entry points: `isFieldRequired`, `isColumnRequired`, `isCellRequired`.
- **`JoyDoc.swift`** — new public model properties: `JoyDocField.requiredLogic`, `FieldTableColumn.requiredLogic`, `FieldTableColumn.cellRequiredLogic`.
- **Table/Collection view models & modal top-nav views** — column titles and the row-edit form now render a required asterisk driven by the computed (not static) required state, with stable `RequiredAsterisk_*` accessibility identifiers.
- **`Models.swift`** — `TableDataModel` now sources `fieldRequired`/`requiredColumnIDs` from the editor's computed logic instead of the static flags.

## Decisions
- **Reused the existing condition/`Logic` model** rather than inventing a parallel schema, so authors use the same condition syntax as show/hide logic and the change is additive/backward-compatible.
- **Cell logic references sibling column ids, not page fields**, so a cell can react to another cell in its own row — required for the per-row use cases.
- **Row-form asterisk recomputes on render/reopen, not live inside an open form.** Values commit per keystroke, but the column-title asterisk re-evaluates on re-render. The reactive UI tests reflect this honestly (edit sibling → dismiss → reopen → assert) rather than asserting a live in-form update that isn't implemented.

## Public API / docs
- **Yes, this is an additive public API change**: 3 new model properties (`requiredLogic` x2, `cellRequiredLogic`) and 3 new `DocumentEditor` methods (`isFieldRequired`/`isColumnRequired`/`isCellRequired`). All additive — no breaking changes.
- Docs should be updated to describe `requiredLogic`/`cellRequiredLogic` and the enforce/unenforce + precedence rules.

## Test coverage
- **Unit:** `RequiredLogicTests.swift` — enforce/unenforce across field, column, and cell levels for both table and collection.
- **UI:** `RequiredLogicTest/RequiredLogic.{swift,json}` — 19 tests covering field/column/cell logic, enforce/unenforce, cell-over-column precedence, sibling-edit reactivity, column-enforce toggle-off, and collection root/child parity. **All 19 pass on iPad Pro 13-inch (M5).**
- No known gaps deferred.

## Media
_To attach: short screen recording of an asterisk appearing/disappearing as a driver field is filled/cleared._

## Reviewer notes
- The UI-test fixture `RequiredLogic.json` must be a member of the **JoyfillExample app target's** Copy Bundle Resources (loaded via `Bundle.main`), not just the UITests target — see the `project.pbxproj` change. The other pbxproj churn is Xcode re-sorting/de-duplicating existing entries; no files were removed.

### DIFF
--- a/Sources/JoyfillUI/View/Fields/TableView/TableModalView.swift
+++ b/Sources/JoyfillUI/View/Fields/TableView/TableModalView.swift
@@ -26,8 +26,9 @@ struct TableRowView : View {
                 .background(Color.rowSelectionBackground(isSelected: isSelected, colorScheme: colorScheme))
                 .border(Color.tableCellBorderColor)
             }
+            let rowElement = viewModel.rowElement(forRowID: rowDataModel.rowID)
             ForEach($rowDataModel.cells, id: \.id) { $cellModel in
-                let showRequired = viewModel.isCellRequired(columnID: cellModel.data.id, rowID: rowDataModel.rowID) && !cellModel.data.isCellFilled
+                let showRequired = viewModel.isCellRequired(columnID: cellModel.data.id, row: rowElement) && !cellModel.data.isCellFilled
                 TableViewCellBuilder(viewModel: viewModel, cellModel: $cellModel)
                     .frame(width: Utility.singleColumnWidth, height: 60)
                     .background(Color.rowSelectionBackground(isSelected: isSelected, colorScheme: colorScheme))
--- a/Sources/JoyfillUI/View/Fields/TableView/TableModalView.swift
+++ b/Sources/JoyfillUI/View/Fields/TableView/TableModalView.swift
@@ -27,7 +27,7 @@ struct TableRowView : View {
                 .border(Color.tableCellBorderColor)
             }
             ForEach($rowDataModel.cells, id: \.id) { $cellModel in
-                let showRequired = viewModel.tableDataModel.requiredColumnIDs.contains(cellModel.data.id) && !cellModel.data.isCellFilled
+                let showRequired = viewModel.isCellRequired(columnID: cellModel.data.id, rowID: rowDataModel.rowID) && !cellModel.data.isCellFilled
                 TableViewCellBuilder(viewModel: viewModel, cellModel: $cellModel)
                     .frame(width: Utility.singleColumnWidth, height: 60)
                     .background(Color.rowSelectionBackground(isSelected: isSelected, colorScheme: colorScheme))
--- a/Sources/JoyfillUI/View/Fields/TableView/TableViewModel.swift
+++ b/Sources/JoyfillUI/View/Fields/TableView/TableViewModel.swift
@@ -32,6 +32,22 @@ class TableViewModel: ObservableObject, TableDataViewModelProtocol {
         return decoratorsWidth + CGFloat(tableDataModel.tableColumns.count) * Utility.singleColumnWidth
     }
 
+    /// Per-cell required-ness for the live grid border. Honours `cellRequiredLogic`
+    /// (resolved against this row's sibling cells), then the column's `requiredLogic`,
+    /// then the static `required` flag. Falls back to the column-wide set if the row
+    /// value can't be resolved.
+    func isCellRequired(columnID: String, rowID: String) -> Bool {
+        guard let documentEditor = tableDataModel.documentEditor,
+              let row = tableDataModel.valueToValueElements?.first(where: { $0.id == rowID }) else {
+            return tableDataModel.requiredColumnIDs.contains(columnID)
+        }
+        return documentEditor.isCellRequired(
+            columnID: columnID,
+            fieldID: tableDataModel.fieldIdentifier.fieldID,
+            row: row
+        )
+    }
+
     private func refreshRowDecoratorMap() {
         guard let field = tableDataModel.documentEditor?.field(fieldID: tableDataModel.fieldIdentifier.fieldID) else { return }
         tableDataModel.setTableRowDecorators(rowDecorators: field.rowDecorators, rows: tableDataModel.valueToValueElements ?? [])
--- a/Sources/JoyfillUI/View/Fields/TableView/TableViewModel.swift
+++ b/Sources/JoyfillUI/View/Fields/TableView/TableViewModel.swift
@@ -37,8 +37,15 @@ class TableViewModel: ObservableObject, TableDataViewModelProtocol {
     /// then the static `required` flag. Falls back to the column-wide set if the row
     /// value can't be resolved.
     func isCellRequired(columnID: String, rowID: String) -> Bool {
-        guard let documentEditor = tableDataModel.documentEditor,
-              let row = tableDataModel.valueToValueElements?.first(where: { $0.id == rowID }) else {
+        return isCellRequired(columnID: columnID, row: rowElement(forRowID: rowID))
+    }
+
+    func rowElement(forRowID rowID: String) -> ValueElement? {
+        tableDataModel.valueToValueElements?.first(where: { $0.id == rowID })
+    }
+
+    func isCellRequired(columnID: String, row: ValueElement?) -> Bool {
+        guard let documentEditor = tableDataModel.documentEditor, let row = row else {
             return tableDataModel.requiredColumnIDs.contains(columnID)
         }
         return documentEditor.isCellRequired(
--- a/Sources/JoyfillUI/ViewModels/RequiredLogicHandler.swift
+++ b/Sources/JoyfillUI/ViewModels/RequiredLogicHandler.swift
@@ -2,12 +2,12 @@
 //  RequiredLogicHandler.swift
 //
 //  Evaluates `requiredLogic` (fields, columns) and `cellRequiredLogic` (per-cell) to produce
-//  an *effective* required-ness that overrides the static `required` flag.
+//  an *effective* required-ness on top of the static `required` flag.
 //
-//  Semantics (action fully overrides the base `required`):
+//  Semantics (the action only changes required-ness when its conditions match; otherwise it falls back to the static base):
 //    - no logic present            -> static `required`
-//    - action == "enforce"         -> required only when conditions match
-//    - action == "unforce"         -> optional only when conditions match (required otherwise)
+//    - action == "enforce"         -> required when conditions match, else static `required`
+//    - action == "unforce"         -> optional when conditions match, else static `required`
 //
 //  Field / column logic conditions reference page-level fields (by `field` id).
 //  Cell logic conditions reference sibling column ids and resolve against the same row's cells.
@@ -193,8 +193,8 @@ class RequiredLogicHandler {
 
     private func applyAction(_ action: String, matched: Bool, staticRequired: Bool) -> Bool {
         switch action {
-        case "enforce": return matched
-        case "unforce": return !matched
+        case "enforce": return matched ? true : staticRequired
+        case "unforce": return matched ? false : staticRequired
         default: return staticRequired
         }
     }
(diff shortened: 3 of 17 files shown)


Sources/JoyfillUI/ViewModels/RequiredLogicHandler.swift
```diff
@@ -0,0 +1,238 @@
...
+        let staticRequired = column.required ?? false
+        guard let logic = column.requiredLogic, let action = logic.action else { return staticRequired }
+        let model = fieldLogicModel(logic: logic)
+        return applyAction(action, matched: documentEditor.conditionalLogicHandler.shoulTakeActionOnThisField(logic: model), staticRequired: staticRequired)
+    }
+
+    private func applyAction(_ action: String, matched: Bool, staticRequired: Bool) -> Bool {
+        switch action {
+        case "enforce": return matched
+        case "unforce": return !matched
```
vivekjyani: **Semantics — please confirm against joyfill-schema / JS & Kotlin SDKs.** `applyAction` makes the action *fully* drive required-ness and ignores the static `required` flag in both branches:

- `enforce` returns `matched`, so a column/field with static `required: true` becomes **optional** whenever the conditions do not match.
- `unforce` returns `!matched`, so a field with static `required: false` (or unset) becomes **required** whenever the conditions do not match.

The second case is the surprising one: adding an `unforce` rule to a non-required column silently makes it required in the no-match state. If the platform spec treats these as additive to the static flag (e.g. enforce = `staticRequired || matched`, unforce = `matched ? false : staticRequired`), this SDK will produce different validation results than the other SDKs for the same document. Please confirm the intended cross-platform semantics; if additive is correct this needs to change.

Sources/JoyfillUI/View/Fields/TableView/TableModalView.swift
```diff
@@ -27,7 +27,7 @@ struct TableRowView : View {
                 .border(Color.tableCellBorderColor)
             }
             ForEach($rowDataModel.cells, id: \.id) { $cellModel in
-                let showRequired = viewModel.tableDataModel.requiredColumnIDs.contains(cellModel.data.id) && !cellModel.data.isCellFilled
+                let showRequired = viewModel.isCellRequired(columnID: cellModel.data.id, rowID: rowDataModel.rowID) && !cellModel.data.isCellFilled
```
vivekjyani: **Performance (non-blocking):** this runs per cell on every grid re-render, and `TableViewModel.isCellRequired` resolves the row via `valueToValueElements?.first(where: { $0.id == rowID })` — an O(rows) linear scan. So rendering one row is O(columns × rows), and scrolling a large table repeats it every frame. The old path was an O(1) `requiredColumnIDs.contains(...)`.

Note the collection side already avoids this: `CollectionViewModel.isCellRequired` uses an O(1) `rowToValueElementMap[rowID]` lookup. Suggest giving `TableViewModel` the same row map, or at minimum hoisting the row lookup out of the per-cell `ForEach` (rowID is constant across cells). Given this repo has a history of large-table SwiftUI perf sensitivity, worth tightening before merge.

Sources/JoyfillUI/ViewModels/RequiredLogicHandler.swift
```diff
@@ -0,0 +1,238 @@
...
+    }
+
+    /// Effective required-ness of a single cell. Precedence: `cellRequiredLogic` (per-row) >
+    /// `requiredLogic` (column-wide) > static `required`.
+    func isCellRequired(columnID: String, fieldID: String, schemaKey: String? = nil, row: ValueElement) -> Bool {
+        guard let column = column(columnID: columnID, fieldID: fieldID, schemaKey: schemaKey) else { return false }
+
+        if let cellLogic = column.cellRequiredLogic, let action = cellLogic.action {
+            let model = cellLogicModel(logic: cellLogic, columns: columns(fieldID: fieldID, schemaKey: schemaKey), row: row)
+            return applyAction(action, matched: documentEditor.conditionalLogicHandler.shoulTakeActionOnThisField(logic: model), staticRequired: column.required ?? false)
```
vivekjyani: Minor: the doc comment states precedence `cellRequiredLogic > requiredLogic (column-wide) > static required`, but when `cellRequiredLogic` is present this passes `staticRequired: column.required` — it skips the column `requiredLogic` entirely and falls back to the static flag, not to the column-wide computed value. For `enforce`/`unforce` the base is ignored anyway so results match today, but if an unknown/other action ever reaches `applyAction` (default branch returns `staticRequired`), a cell would fall back to the static flag rather than the column logic, contradicting the documented order. Either wire the fallback through `computeColumnRequired(column:)` or adjust the comment.

Sources/JoyfillUI/View/Fields/TableView/TableModalView.swift
vivekjyani: Good catch. Hoisted the row lookup out of the per-cell ForEach — TableRowView now resolves the ValueElement once per row via rowElement(forRowID:) and passes it into a new isCellRequired(columnID:row:) overload, so a row render is one lookup + O(columns) instead of O(columns × rows). The old isCellRequired(columnID:rowID:) is kept (delegates to the new overload) for the row-form title path. Behavior is unchanged; RequiredLogic suite still 19/19 green.

Sources/JoyfillUI/ViewModels/RequiredLogicHandler.swift
vivekjyani: Fixed — the cell branch now passes staticRequired: computeColumnRequired(column:) instead of the raw column.required, so the fallback resolves through the column's requiredLogic and the documented cell > column > static precedence holds even for an unknown action. No behavior change for enforce/unforce (they ignore the base). Build green.

Sources/JoyfillUI/ViewModels/RequiredLogicHandler.swift
vivekjyani: Confirmed against the feature spec's Resolution Order: cellRequiredLogic → column requiredLogic → static required, and crucially "we only follow the old required property if there is no cell or column required logic." So when logic is present it fully overrides the static flag rather than combining with it — the additive interpretation would actually diverge from the spec.

This matches the current applyAction:

enforce → matched: spec's own example (marital status optional by default, enforce on age ≥ 18) → age 12 = optional, age 23 = required. ✓
unforce → !matched: matched = optional, otherwise required — consistent with the override principle (static is never consulted when logic exists).
Keeping the code as-is. One undocumented edge — unforce on a column that isn't statically required (no-match resolves to required) — is an unusual authoring case; I'll confirm with the spec author but it doesn't affect the documented behavior or block this PR.

Sources/JoyfillUI/ViewModels/RequiredLogicHandler.swift
```diff
@@ -0,0 +1,238 @@
...
+        let staticRequired = column.required ?? false
+        guard let logic = column.requiredLogic, let action = logic.action else { return staticRequired }
+        let model = fieldLogicModel(logic: logic)
+        return applyAction(action, matched: documentEditor.conditionalLogicHandler.shoulTakeActionOnThisField(logic: model), staticRequired: staticRequired)
+    }
+
+    private func applyAction(_ action: String, matched: Bool, staticRequired: Bool) -> Bool {
+        switch action {
+        case "enforce": return matched
+        case "unforce": return !matched
```
vivekjyani: `applyAction` ignores `staticRequired` for both `enforce` and `unforce` (it's only used in the `default` case), so the logic *fully* overrides the static flag in both directions. The consequence worth confirming: attaching `unforce` to a field/column that is **not** statically required makes it **required** whenever the conditions do NOT match (`!matched` = `!false` = `true`). Intuitively `unforce` should only ever *relax* required-ness, never add it. The PR description only defines `unforce` as “optional when conditions match (overrides static required:true)” and says nothing about the not-matched + not-statically-required case. Does this match the web/Kotlin reference implementation? If web falls back to the static flag on the non-triggering branch, this diverges. Also note: passing `computeColumnRequired(column:)` as `staticRequired` at line 68 is inert for known actions — it only affects malformed action strings.

Sources/JoyfillUI/View/Fields/TableView/TableViewModel.swift
```diff
@@ -32,6 +32,29 @@ class TableViewModel: ObservableObject, TableDataViewModelProtocol {
...
+    /// Per-cell required-ness for the live grid border. Honours `cellRequiredLogic`
+    /// (resolved against this row's sibling cells), then the column's `requiredLogic`,
+    /// then the static `required` flag. Falls back to the column-wide set if the row
+    /// value can't be resolved.
+    func isCellRequired(columnID: String, rowID: String) -> Bool {
+        return isCellRequired(columnID: columnID, row: rowElement(forRowID: rowID))
+    }
+
+    func rowElement(forRowID rowID: String) -> ValueElement? {
+        tableDataModel.valueToValueElements?.first(where: { $0.id == rowID })
```
vivekjyani, line 44: `rowElement(forRowID:)` does a linear `.first(where:)` scan, and `isCellRequired` is invoked per-cell on every SwiftUI render of `TableRowView`. That's ~O(visibleRows × cols × totalRows) per render, versus the previous `requiredColumnIDs.contains` (a `Set` lookup). Given the known large-table render-cost sensitivity, consider caching a `rowID -> ValueElement` map (as `CollectionViewModel` already does via `rowToValueElementMap`), and/or short-circuiting when the column has neither `requiredLogic` nor `cellRequiredLogic`. Non-blocking.

vivekjyani (commented):
Automated review pass (correctness + API + schema impact). Two inline notes below; full write-up posted separately. Overall: solid, well-tested, additive. One semantic question on `unforce` worth confirming against the web/Kotlin SDK before merge.

Sources/JoyfillUI/ViewModels/RequiredLogicHandler.swift
vivekjyani: You're right. Confirmed against the Kotlin reference: enforce/unforce should only change required-ness when their conditions match, and fall back to the static required otherwise (not force the opposite). Fixed applyAction to be directional and re-ran the RequiredLogic UI suite — 19/19 passing on iPad.

Sources/JoyfillUI/View/Fields/TableView/TableViewModel.swift
vivekjyani: Good point, and agreed on the analysis. It's non-blocking, so I'd like to take it as a follow-up rather than in this PR. Plan is to cache a rowID → ValueElement map (mirroring CollectionViewModel.rowToValueElementMap) so the lookup is O(1) — I'll handle it separately to keep this PR focused on the required-logic behavior.

PR #357, opened 2026-07-24
NO-2271: Fix collection/table bulk-edit crash via columnID keying

vivekjyani:
## Context
Editing a single collection/table row through the goto validation flow (Up/Down navigation into a row form) could crash (RUID 5ZYFSV) in the bulk-edit path. The row form buffered edits keyed by **column index**, but when it renders a nested schema — whose column set and ordering differ from the parent — those indexes no longer line up with the target table's columns, so an edit either lands on the wrong column or reads past the end of the cell array.

## What changed
- **`CollectionEditMultipleRowsSheetView` / `TableEditMultipleRowsSheetView`** (`CollectionModalTopNavigationView.swift`, `TableModalTopNavigationView.swift`): the in-flight `changes` buffer is now `[String: ValueUnion]` keyed by `columnID` instead of `[Int: ValueUnion]` keyed by column index. Added a `resetChanges()` helper and reset the buffer when the selected row changes.
- **`CollectionViewModel.bulkEdit` / `TableViewModel.bulkEdit`**: signature changed from `[Int: ValueUnion]` to `[String: ValueUnion]`; dropped the now-unnecessary index→columnID remap. Added bounds guards in `updateBulkLocalCellModels` (rowIndex and colIndex) that log and skip instead of trapping.
- **Tests**: unit tests (`CollectionViewModelDocumentEditorDelegateTests`, `TableViewModelDocumentEditorDelegateTests`) for columnID keying and the out-of-bounds path; a UI test (`NavigationGotoUITests.testGotoNavigateAndFillCollectionTextFields`) plus a `footer-form.json` fixture with a nested-schema collection.

## Decisions
- **Key by `columnID`, not index.** Column identity is stable across schemas; index is not. Keying by ID removes the fragile index↔column remap entirely rather than patching it.
- **Guard-and-skip over guard-and-crash.** The bounds checks `Log(.warning)` and `continue`/`return` so a stale index degrades to a no-op edit instead of a trap.

## Public API / docs
None. Both `bulkEdit(changes:)` methods are `internal`; the signature change is not part of the public SDK surface, so no docs update is required.

## Test coverage
- Unit: columnID-keyed bulkEdit and out-of-bounds row/column indexes (no crash).
- UI: goto navigation into the nested-schema collection, filling all five page-1 text cells via the live single-row bulkEdit, asserting the app stays alive — passes on iPad Pro 13-inch (M5).
- Covered in this PR; no follow-up tests planned.

## Notes for reviewer
- The UI test deliberately skips the first four Up landings (Number/table/collection1) and fills only the page-1 collection's text cells — that is the schema that triggered the original crash.
- No screen recording embedded here; the XCUITest run video is in the test's `.xcresult` if needed.

### DIFF
--- a/JoyfillSwiftUIExample/JoyfillTests/OnChangeHandler/CollectionViewModelDocumentEditorDelegateTests.swift
+++ b/JoyfillSwiftUIExample/JoyfillTests/OnChangeHandler/CollectionViewModelDocumentEditorDelegateTests.swift
@@ -544,4 +544,118 @@ final class CollectionViewModelDocumentEditorDelegateTests: XCTestCase {
         let rowIndexForDocument = updatedRowsFromDocumentEditor?.firstIndex(where: {$0.id == "68575bb9cdb3707c78d6b2ff"})
         XCTAssertEqual(rowIndexForDocument, 3, "Value should be equal")
     }
+
+    // MARK: - bulkEdit(changes:) — regression coverage for RUID 5ZYFSV
+    // The reported crash was in CollectionViewModel.bulkEdit, which keyed row edits by
+    // column *index* and read arrays out of bounds. These lock in the columnID-keyed path.
+
+    private let rootTextColumnID = "684c3fedb0afd867adaeb3b4"
+    private let rootRowA = "68575bb9cdb3707c78d6b2ff"
+    private let rootRowB = "685765dcf190077e95796c41"
+    private let rootRowUnselected = "68582dd76e0e93dd2017372a"
+
+    private func cellText(_ documentEditor: DocumentEditor, row: String, column: String) -> String? {
+        documentEditor.field(fieldID: tableFieldID)?.valueToValueElements?
+            .first(where: { $0.id == row })?.cells?[column]?.text
+    }
+
+    private func cellValue(_ documentEditor: DocumentEditor, row: String, column: String) -> ValueUnion? {
+        documentEditor.field(fieldID: tableFieldID)?.valueToValueElements?
+            .first(where: { $0.id == row })?.cells?[column]
+    }
+
+    // Core of the fix: changes are keyed by columnID, so each entry must land in its
+    // own root column across every type in the switch, on all selected rows.
+    func testBulkEdit_mapsEachColumnIDToItsOwnColumnAcrossTypes() async throws {
+        let documentEditor = DocumentEditor(document: createTestDocument(), validateSchema: false)
+        let viewModel = try await createCollectionViewModel(documentEditor: documentEditor)
+        sleep(10)
+
+        let dropdownColumnID = "684c3fed52b1d1145f1e2790"
+        let numberColumnID   = "685753044b333dd442ea29d4"
+        let multiColumnID    = "6857530158fb7edb102344fa"
+        let barcodeColumnID  = "68575312d8c5679a05ee29e0"
+        let dropdownOptionID = "684c3fedf47cc0fea6bca947"          // "No D1"
+        let multiOptionIDs   = ["68575301e490d0ce22ae5e7b", "685767dae7cf2193a50ff550"]
+
+        viewModel.tableDataModel.selectedRows = [rootRowA, rootRowB]
+
+        await viewModel.bulkEdit(changes: [
+            rootTextColumnID: .string("BulkText"),
+            dropdownColumnID: .string(dropdownOptionID),
+            numberColumnID:   .double(42),
+            multiColumnID:    .array(multiOptionIDs),
+            barcodeColumnID:  .string("BC-123"),
+        ])
+        waitForMainQueueToDrain()
+
+        for row in [rootRowA, rootRowB] {
+            XCTAssertEqual(cellValue(documentEditor, row: row, column: rootTextColumnID)?.text, "BulkText", "text column on \(row)")
+            XCTAssertEqual(cellValue(documentEditor, row: row, column: dropdownColumnID)?.text, dropdownOptionID, "dropdown column on \(row)")
+            XCTAssertEqual(cellValue(documentEditor, row: row, column: numberColumnID)?.number, 42, "number column on \(row)")
+            XCTAssertEqual(cellValue(documentEditor, row: row, column: multiColumnID)?.multiSelector, multiOptionIDs, "multiSelect column on \(row)")
+            XCTAssertEqual(cellValue(documentEditor, row: row, column: barcodeColumnID)?.text, "BC-123", "barcode column on \(row)")
+        }
+    }
+
+    func testBulkEdit_appliesValueToAllSelectedRows() async throws {
+        let documentEditor = DocumentEditor(document: createTestDocument(), validateSchema: false)
+        let viewModel = try await createCollectionViewModel(documentEditor: documentEditor)
+        sleep(10)
+
+        let unselectedBefore = cellText(documentEditor, row: rootRowUnselected, column: rootTextColumnID)
+        viewModel.tableDataModel.selectedRows = [rootRowA, rootRowB]
+
+        await viewModel.bulkEdit(changes: [rootTextColumnID: .string("Testing")])
+        waitForMainQueueToDrain()
+
+        XCTAssertEqual(cellText(documentEditor, row: rootRowA, column: rootTextColumnID), "Testing", "Selected root row A should get the bulk value")
+        XCTAssertEqual(cellText(documentEditor, row: rootRowB, column: rootTextColumnID), "Testing", "Selected root row B should get the bulk value")
+        XCTAssertEqual(cellText(documentEditor, row: rootRowUnselected, column: rootTextColumnID), unselectedBefore, "Unselected root row must be untouched")
+    }
+
+    func testBulkEdit_unknownColumnID_isNoOpAndDoesNotCrash() async throws {
+        let documentEditor = DocumentEditor(document: createTestDocument(), validateSchema: false)
+        let viewModel = try await createCollectionViewModel(documentEditor: documentEditor)
+        sleep(10)
+
+        let before = cellText(documentEditor, row: rootRowA, column: rootTextColumnID)
+        viewModel.tableDataModel.selectedRows = [rootRowA, rootRowB]
+
+        await viewModel.bulkEdit(changes: ["nonexistent-column-id": .string("ShouldNotApply")])
+        waitForMainQueueToDrain()
+
+        XCTAssertEqual(cellText(documentEditor, row: rootRowA, column: rootTextColumnID), before, "An unknown columnID must not mutate any cell")
+    }
+
+    func testBulkEdit_emptyChanges_isNoOp() async throws {
+        let documentEditor = DocumentEditor(document: createTestDocument(), validateSchema: false)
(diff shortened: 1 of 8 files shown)


Sources/JoyfillUI/View/Fields/CollectionView/CollectionModalTopNavigationView.swift
```diff
@@ -514,13 +519,14 @@ struct CollectionEditMultipleRowsSheetView: View {
         let header = viewModel.getHeaderForSelectedRows()
         ForEach(Array(header.columns.enumerated()), id: \.offset) { colIndex, col in
             let isFocused = col.id == viewModel.tableDataModel.navigationIntent.focusColumnId
+            let columnID = col.id ?? ""
```
VishnuBishnoi, line 522: Non-blocking: `col.id ?? ""` collapses a nil column id to `""`. Since `makeChangeDict` matches with `tableColumns.first(where: { $0.id == key })` and `$0.id` is `String?`, a nil-id column never equals `Optional("")`, so such an edit is silently dropped (and multiple nil-id columns would collide on the same `""` key in the buffer). In practice columns always carry `_id`, so this is theoretical — but keying on `""` hides the drop. Consider `guard let columnID = col.id else { /* skip */ }` to make it explicit. Same pattern in TableModalTopNavigationView.swift:407.

Sources/JoyfillUI/View/Fields/CollectionView/CollectionViewModel.swift
```diff
@@ -1784,8 +1784,17 @@ class CollectionViewModel: ObservableObject, TableDataViewModelProtocol {
...
             guard let rowIndex = rowIndexMap[rowId] else { continue }
+            guard rowIndex < updatedCellModels.count,
+                  rowIndex < tableDataModel.filteredcellModels.count else {
+                Log("updateBulkLocalCellModels: rowIndex \(rowIndex) out of bounds (rows: \(updatedCellModels.count)) for row \(rowId)", type: .warning)
+                continue
+            }
             var rowDataModel = tableDataModel.filteredcellModels[rowIndex]
             tableColumns.enumerated().forEach { colIndex, column in
+                guard colIndex < rowDataModel.cells.count else {
+                    Log("updateBulkLocalCellModels: column index \(colIndex) exceeds cell count \(rowDataModel.cells.count) for row \(rowId)", type: .warning)
```
VishnuBishnoi, line 1795: Non-blocking / question: this guard skips the *local* cell-model update, but `updateJSON(...)` a few lines up has already persisted the change to the document. So in the degenerate case (row has fewer cells than columns) the value lands in JSON but the on-screen cell keeps its old value until a refresh — not quite the "no-op edit" the PR description states. Can this skip actually trigger in the normal nested-schema goto flow, and if so does `refreshRowsOnLogic` reconcile the UI? If it's purely defensive-against-a-trap, fine as-is.

JoyfillSwiftUIExample/JoyfillTests/OnChangeHandler/CollectionViewModelDocumentEditorDelegateTests.swift
```diff
@@ -544,4 +544,118 @@ final class CollectionViewModelDocumentEditorDelegateTests: XCTestCase {
...
+        let dropdownColumnID = "684c3fed52b1d1145f1e2790"
+        let numberColumnID   = "685753044b333dd442ea29d4"
+        let multiColumnID    = "6857530158fb7edb102344fa"
+        let barcodeColumnID  = "68575312d8c5679a05ee29e0"
+        let dropdownOptionID = "684c3fedf47cc0fea6bca947"          // "No D1"
+        let multiOptionIDs   = ["68575301e490d0ce22ae5e7b", "685767dae7cf2193a50ff550"]
+
+        viewModel.tableDataModel.selectedRows = [rootRowA, rootRowB]
+
+        await viewModel.bulkEdit(changes: [
```
VishnuBishnoi, line 585: Non-blocking: the "maps each columnID to its own column across types" test covers text/dropdown/number/multiSelect/barcode but omits `.date`, `.signature`, and `.image` — the three remaining switch branches. Adding a date column here would also give the `makeChangeDict` timezone-conversion path (multi-row + `.date`) its first assertion. Same gap in the Table version.

VishnuBishnoi (commented):
Reviewed the full diff (3 commits) against `main`. The columnID-keying approach is the right fix and removes the fragile index↔column remap entirely; thread-safety pattern (main-thread snapshot before the background block) is preserved. No public API change (both `bulkEdit(changes:)` are `internal`, protocol doesn't declare it). Approving with a few non-blocking notes left inline.

Sources/JoyfillUI/View/Fields/CollectionView/CollectionModalTopNavigationView.swift
vivekjyani: Good catch, and agreed on the theoretical risk. Holding off on this for now — columns always carry a real _id in practice, so this path isn't reachable today, and I'd rather not touch working code without a concrete repro. Filing it as a known edge case; happy to revisit if we ever see a nil column id in the wild.

Sources/JoyfillUI/View/Fields/CollectionView/CollectionViewModel.swift
vivekjyani: Good catch on the mechanics. Holding off on fixing this for now — I don't think it's reachable via the current flow, since bulk-edit here is scoped to single-row editing and a row's own cells will always match its own schema's column count. This guard is defense-in-depth against the original crash, not a path we expect to hit today. Leaving as-is; will revisit if multi-row/mixed-schema bulk edit becomes a real use case.

JoyfillSwiftUIExample/JoyfillTests/OnChangeHandler/CollectionViewModelDocumentEditorDelegateTests.swift
vivekjyani: Added — extended testBulkEdit_mapsEachColumnIDToItsOwnColumnAcrossTypes with a .date column, reusing the existing two-row selection. This exercises makeChangeDict's multi-row timezone-conversion branch for the first time. Left .signature/.image and the Table-side test out of scope for now since they're lower priority — happy to add if you'd like.

Sources/JoyfillUI/View/Fields/CollectionView/CollectionModalTopNavigationView.swift
vivekjyani: `col.id` is `Optional<String>`. If two columns in a malformed doc both lack an `_id`, they'd collide on the empty-string key and silently clobber each other's buffered edit here (and in the analogous spot in `TableModalTopNavigationView.swift`). Low probability in practice since real docs always assign `_id`, but given this PR's guard-and-skip philosophy elsewhere (see the new bounds checks in `CollectionViewModel.updateBulkLocalCellModels`), a `Log(.warning)` when `col.id` is nil would be consistent.

Sources/JoyfillUI/View/Fields/TableView/TableModalTopNavigationView.swift
```diff
@@ -400,13 +404,14 @@ struct EditMultipleRowsSheetView: View {
             VStack(alignment: .leading, spacing: 16) {
                 ForEach(Array(viewModel.tableDataModel.tableColumns.enumerated()), id: \.offset) { colIndex, col in
                     let isFocused = col.id == viewModel.tableDataModel.navigationIntent.focusColumnId
+                    let columnID = col.id ?? ""
```
vivekjyani: Same `col.id ?? ""` pattern as `CollectionModalTopNavigationView.swift` — flagging here too since both files share the risk (see comment on the other file for details).

Sources/JoyfillUI/View/Fields/CollectionView/CollectionViewModel.swift
```diff
@@ -1784,8 +1784,17 @@ class CollectionViewModel: ObservableObject, TableDataViewModelProtocol {
 
         for rowId in tableDataModel.selectedRows {
             guard let rowIndex = rowIndexMap[rowId] else { continue }
+            guard rowIndex < updatedCellModels.count,
+                  rowIndex < tableDataModel.filteredcellModels.count else {
+                Log("updateBulkLocalCellModels: rowIndex \(rowIndex) out of bounds (rows: \(updatedCellModels.count)) for row \(rowId)", type: .warning)
```
vivekjyani, line 1789: Nit: worth double-checking this warning never fires in the non-buggy path (i.e. `rowIndexMap` and `filteredcellModels` are expected to always stay in sync) so it doesn't become log noise once this ships.

vivekjyani (commented):
Reviewed the columnID-keying fix — solid approach. Left a couple of minor inline notes; nothing blocking.

Sources/JoyfillUI/View/Fields/CollectionView/CollectionModalTopNavigationView.swift
vivekjyani: Fixed, i have used if lets here

Sources/JoyfillUI/View/Fields/TableView/TableModalTopNavigationView.swift
vivekjyani: Fixed, i have used if lets here

Sources/JoyfillUI/View/Fields/CollectionView/CollectionViewModel.swift
vivekjyani: No issues here

Sources/JoyfillUI/View/Fields/CollectionView/CollectionModalTopNavigationView.swift
vivekjyani: Looked into whether a nil `col.id` is actually reachable, and I don't think it is — so I'm leaving this as-is.

`FieldTableColumn.id` is `String?` only because the model is a dictionary wrapper (`dictionary["_id"] as? String`, JoyDoc.swift:1091), not because the format permits a column without an id. The schema marks `_id` required on every column variant (`joyfill-schema.swift:1882` and the sibling `required` blocks), and a document that fails schema validation is blanked entirely, so the edit sheet never renders for such a doc in the first place.

The only way to reach nil is `validateSchema: false` plus a hand-built column dictionary. In that case the very next thing this body does is `switch col.type`, and `type` force-unwraps (`dictionary["type"] as! String`) — so the surrounding code already assumes well-formed columns far more aggressively than the `?? ""` does. Adding a guard here wouldn't save that document.

I did prototype the fix both ways (wrapping the body in `if let columnID = col.id`, and pre-filtering the `ForEach` source via `compactMap` into `(offset:column:columnID:)` tuples). Both build clean, but neither changes observable behavior for any document that can actually reach this code, so it's churn in a 200-line view body for a theoretical case. Consistent with how we left the same question earlier in this PR: not shipping a speculative fix without a repro.

Happy to revisit if `_id` ever becomes genuinely optional in the schema.

VishnuBishnoi (approved):
LGTM

PR #358, opened 2026-07-29
NO-2257: Conditional required-ness and cell-visibility logic for Table/Collection

vivekjyani:
## Context
Ticket NO-2257 (builds on NO-2256) adds two related conditional-logic features to Table/Collection fields:
1. **Required logic** — field/column/cell required-ness can now be conditionally enforced/unenforced via `requiredLogic` / `cellRequiredLogic`, instead of only a static `required` flag.
2. **Cell visibility logic** — individual cells can be conditionally shown/hidden via `cellVisibilityLogic`, driven by a sibling column's value or a page-level field's value, with a static `cellsHidden` baseline for cells that have no logic.

## What changed
- **Model** (`JoyDoc.swift`): additive-only optional properties — `JoyDocField.requiredLogic`, `Condition.column` (sibling-column condition target), `FieldTableColumn.requiredLogic` / `cellRequiredLogic` / `cellVisibilityLogic` / `cellsHidden`.
- **New `RequiredLogicHandler.swift`**: mirrors `ConditionalLogicHandler`'s dependency-map pattern to resolve field/column/cell required-ness (`isFieldRequired`, `isColumnRequired`, `isCellRequired`) and compute refresh dependents.
- **`ConditionalLogicHandler.swift`**: purely additive — cell-visibility build/refresh methods (`buildCellVisibilityFor{Table,Collection}Field`, `shouldShowCell`, `cellsNeedToBeRefreshed`, `cellsNeedRefreshForPageField`, `addCellVisibilityForRow`, `removeCellVisibilityForRow`).
- **`DocumentEditor.swift`**: new `requiredLogicHandler`, new delegate hook `cellVisibilityDidChange(columnIDs:)` (default no-op, non-breaking), `refreshDependent(for:)` now also unions required-logic dependents.
- **Table/Collection views & view models**: `TableViewModel`/`CollectionViewModel` cell-change handlers now trigger visibility refresh on sibling edits; `CollectionModelView`/`CollectionQuickView`/`CollectionViewCellBuilder` compute a cell's hidden state live via `isCellHidden` instead of a stored flag (fixes sibling-edit refresh not showing up in Collection — page-field-driven refresh already worked, sibling-column-driven refresh didn't).
- **`ValidationHandler.swift` / `Models.swift`**: required-ness checks now route through `documentEditor.isFieldRequired` / `isColumnRequired` / `isCellRequired` instead of reading the static `.required` flag directly.

## Design decisions
- Required-logic and cell-visibility-logic are kept as **separate handlers** (`RequiredLogicHandler` vs `ConditionalLogicHandler`) rather than merged, since they have different resolution semantics (required resolution order is cell → column → static-only-if-no-logic; visibility is action-driven per row and ignores `cellsHidden` once logic is present).
- Cell visibility is resolved **live** (`isCellHidden(columnID:row:)`) rather than cached on the row/cell model, so sibling-column edits refresh correctly without needing an extra invalidation pass. The Collection crash-prevention `safeFilteredModels` snapshot pattern was intentionally preserved as-is.
- **Known, intentionally unfixed gap**: the "Apply All" bulk-edit sheet (2+ rows selected) does not currently hide `cellVisibilityLogic`-hidden columns for either Table or Collection — the single-row-edit-form gate (`selectedRows.count == 1`) never runs for bulk selections. Root-caused and documented, but out of scope per product decision — flagging for reviewer awareness, not asking for a fix in this PR.

## Screenshots / video
_Not attached yet — will add a short screen recording of the Table/Collection cell show/hide behavior and required-logic asterisk toggling before merge._

## Public API impact
Yes, additive public API changes:
- New model properties on `JoyDocField` / `FieldTableColumn` (see above) — all optional, non-breaking for existing JSON documents.
- New `DocumentEditorDelegate.cellVisibilityDidChange(columnIDs:)` hook (default implementation provided, so existing conformers are unaffected).
- New public `DocumentEditor` methods: `shouldShowCell`, `cellsNeedToBeRefreshed` (x2), `addCellVisibilityForRow` (x2), `removeCellVisibilityForRow`, `isFieldRequired`, `isColumnRequired`, `isCellRequired`.
- **Docs likely need updating** to document `requiredLogic`, `cellRequiredLogic`, `cellVisibilityLogic`, and `cellsHidden` schema keys, plus the new delegate hook.

## Test coverage
- Unit tests: `RequiredLogicTests.swift` (795 lines) and `CellVisibilityLogicTest.swift` (1052 lines, 46 tests, all passing) cover field/column/cell required resolution, enforce/unenforce directionality, cell-visibility show/hide conditions (=, !=, ?=, >, <, null=, *=, AND/OR), page-field vs sibling-column dependency refresh, row add/delete/duplicate visibility survival, and single-row vs bulk edit-form gating.
- UI tests: `RequiredLogicTest/RequiredLogic.swift` and `CellVisibilityLogicTest/CellVisibilityLogic.swift` (26 tests) cover the same behaviors end-to-end through the actual Table/Collection row-edit and bulk-edit sheets — all passing.
- Coverage is believed complete for the shipped behavior; no follow-up test-writing planned unless review surfaces a gap.

## Notes for reviewers
- The bulk-edit hidden-column gap described above is a known limitation, not a regression — please don't flag it as a bug in review, it's tracked separately.
- `Models.swift` and `ConditionalLogicHandler.swift`'s `shoulTakeActionOnThisField` visibility widening are the only "non-purely-additive" touches outside the two new handlers; everything else is additive.
- This branch also carries the merged NO-2256 history (required-logic groundwork); the diff vs `main` therefore includes some earlier commits from that ticket in addition to the cell-visibility work.


Sources/JoyfillUI/ViewModels/ValidationHandler.swift
```diff
@@ -221,7 +221,7 @@ class ValidationHandler {
             var cellValidities = [CellValidity]()
             for column in visibleColumns {
                 guard let columnID = column.id else { continue }
-                let isRequired = column.required ?? false
+                let isRequired = documentEditor.isCellRequired(columnID: columnID, fieldID: fieldID, row: row)
```
vivekjyani, line 230: **Blocking:** this loop filters `visibleColumns` by `shouldShowColumn` (whole-column visibility) but never checks the new per-row `documentEditor.shouldShowCell(columnID:fieldID:row:)`. If a column has `cellRequiredLogic`/`requiredLogic` making a specific row's cell required, and that same column also has `cellVisibilityLogic` hiding that cell in that row, `HiddenCellView` renders (per `TableViewCellBuilder.swift`) but this still marks the cell `.invalid` because it's empty — the user has no way to fill a cell they can't see, so the field can never validate.

Suggest gating the required-check the same way `shouldShowColumn` gates the column:
```swift
if documentEditor.shouldShowColumn(columnID: columnID, fieldID: fieldID),
   !documentEditor.shouldShowCell(columnID: columnID, fieldID: fieldID, row: row) {
    cellValidities.append(CellValidity(status: .valid, columnId: columnID, value: cells[columnID]))
    continue
}
```
Neither `RequiredLogicTests.swift` nor `CellVisibilityLogicTest.swift` has a test combining `cellVisibilityLogic` with `validate()`/`cellStatus()` — worth adding one alongside the fix.

Sources/JoyfillUI/ViewModels/ValidationHandler.swift
```diff
@@ -312,7 +312,7 @@ class ValidationHandler {
                 continue
             }
 
-            let isRequired = column.required ?? false
+            let isRequired = documentEditor.isCellRequired(columnID: columnID, fieldID: fieldID, schemaKey: schemaId, row: row)
```
vivekjyani, line 326: **Blocking (same issue as the table path above):** `validateCollectionRowCells` checks `shouldShowColumn` at line 311 but never `documentEditor.shouldShowCell(...)` before treating an empty required cell as invalid. A collection cell hidden by `cellVisibilityLogic` but required via `cellRequiredLogic` can't be validated by the user. Same fix shape as `ValidationHandler.swift:224` — short-circuit to `.valid` when `!shouldShowCell(...)`.

Sources/JoyfillUI/ViewModels/ConditionalLogicHandler.swift
```diff
@@ -612,3 +628,181 @@ extension ColumnTypes {
...
+    /// static `cellsHidden` baseline applies only to columns without logic.
+    @discardableResult
+    private func updateCellVisibility(fieldID: String, columns: [FieldTableColumn], columnID: String, row: ValueElement) -> Bool {
+        guard let rowID = row.id else { return false }
+        let cellID = CellVisibilityID(rowID: rowID, columnID: columnID)
+
+        let column = columns.first(where: { $0.id == columnID })
+
+        let newValue: Bool
+        if let logic = column?.cellVisibilityLogic, let action = logic.action {
```
vivekjyani: **Non-blocking:** this only treats `cellsHidden` as ignored when the column's `cellVisibilityLogic` has a non-nil `action`. If a column ever carries `cellVisibilityLogic` with conditions but a missing/nil `action`, execution falls through to the `else` branch and applies the static `cellsHidden` baseline — which contradicts the stated rule (see the doc comment above this function: "the static `cellsHidden` baseline applies only to columns without logic"). Likely unreachable if the schema always emits an `action` alongside logic, but worth an explicit guard (e.g. default to visible, or assert) so the behavior doesn't silently depend on backend schema guarantees.

vivekjyani (commented):
Thorough pass on the required-logic + cell-visibility-logic work. Overall this is well-designed and well-tested (46+ unit tests exercising real DocumentEditor behavior, not shallow calls), but there's one blocking gap: validation doesn't account for cells hidden by `cellVisibilityLogic`, so a required-but-hidden cell can permanently block form submission with no way for the user to fill it in. See inline comments. Non-blocking hardening note also inline. No breaking public API changes — all new surface is additive.

Sources/JoyfillUI/ViewModels/ValidationHandler.swift
vivekjyani: Good catch — fixed. Added a shouldShowCell check right before the required-check, same shape you suggested: if the cell is hidden, it's short-circuited to .valid regardless of required-ness. Also added testValidateTreatsHiddenRequiredTableCellAsValid in CellVisibilityLogicTest.swift covering a required+hidden cell through validate().

Sources/JoyfillUI/ViewModels/ValidationHandler.swift
vivekjyani: Fixed with the same shape as the table path — shouldShowCell gate added before the required-check in validateCollectionRowCells. Added testValidateTreatsHiddenRequiredCollectionCellAsValid to cover it. (Note: this test needed a license-bearing DocumentEditor init since collection validation is gated behind isCollectionFieldEnabled, same as the existing collection tests in ValidationTestCase.swift.)

Sources/JoyfillUI/ViewModels/ConditionalLogicHandler.swift
vivekjyani: Fixed — restructured updateCellVisibility so cellVisibilityLogic with a nil action explicitly defaults to visible instead of falling through to the cellsHidden baseline, matching the doc comment above the function. Added testCellVisibilityLogicWithoutActionDefaultsToVisibleIgnoringCellsHidden to lock in the behavior. All 223 tests across CellVisibilityLogicTest/RequiredLogicTests/ValidationTestCase still pass.

vivekjyani (commented):
Leaving one non-blocking schema/documentation note from review.

Sources/JoyfillModel/JoyDoc.swift
```diff
@@ -1200,6 +1212,34 @@ public struct FieldTableColumn {
         set { dictionary["logic"] = newValue?.dictionary }
     }
 
+    /// Column-wide required logic. Conditions reference page-level fields. `action` is "enforce"/"unenforce"
+    /// and overrides the static `required` flag for every cell in the column.
+    public var requiredLogic: Logic? {
```
vivekjyani, line 1217: Non-blocking, but these new document keys should be mirrored in `Sources/JoyfillUI/View/joyfill-schema.swift` (including the sibling-column condition shape) so schema validation/documentation knows about `requiredLogic`, `cellRequiredLogic`, `cellVisibilityLogic`, and `cellsHidden`. The current schema appears permissive enough that this should not hard-fail existing validation, but it leaves the new contract undocumented and unvalidated.

vivekjyani: Agreed, that's a real gap — thanks for flagging it. Since it's additive/documentation-only (no behavior change, low risk of breaking existing validation), I'd rather not bloat this already-large diff with it. I'll follow the existing SchemaLogic/SchemaLogicCondition pattern (same sibling-column vs page-field condition shape) to add a CellLogic/CellLogicCondition pair and wire requiredLogic, cellRequiredLogic, cellVisibilityLogic, and cellsHidden into the column-type schemas, in a fast-follow PR right after this one merges.

Sources/JoyfillUI/ViewModels/DocumentEditor.swift
```diff
@@ -738,10 +784,15 @@ extension DocumentEditor {
...
     func refreshDependent(for fieldID: String) {
-        let refreshFields = conditionalLogicHandler.fieldsNeedsToBeRefreshed(fieldID: fieldID)
+        var refreshFields = Set(conditionalLogicHandler.fieldsNeedsToBeRefreshed(fieldID: fieldID))
+        refreshFields.formUnion(requiredLogicHandler.fieldsNeedsToBeRefreshed(fieldID: fieldID))
         for fieldId in refreshFields {
             refreshField(fieldId: fieldId)
         }
+        for (tableFieldID, columnIDs) in conditionalLogicHandler.cellsNeedRefreshForPageField(pageFieldID: fieldID) {
+            let fieldType = field(fieldID: tableFieldID)?.fieldType ?? .table
+            valueDelegate(for: tableFieldID, fieldType: fieldType)?.cellVisibilityDidChange(columnIDs: columnIDs)
```
VishnuBishnoi, line 807: **Worth checking (not blocking — pre-existing gap, this just extends it):** `refreshDependent(for:)` is called from `handleFieldValueRowCreate`/`handleFieldValueRowDelete`/`handleFieldValueRowMove` (lines ~384-452), none of which dispatch to main — only `handleFieldValueRowUpdate` above wraps its work in `DispatchQueue.main.async`. Through `refreshDependent`, this line now also fires `cellVisibilityDidChange(columnIDs:)`, which on `CollectionViewModel`/`TableViewModel` mutates `tableDataModel.filteredcellModels`/`cellModels` directly (see `CollectionViewModel.applyVisibilityFlip` / `TableViewModel.applyVisibilityFlip`) — and `tableDataModel` is `@Published`.

`change(changes:)` is public API with no documented threading contract. If a host ever calls it off the main thread with a `.fieldValueRowCreate/Delete/Move` change that touches a `cellVisibilityLogic` dependency, this mutates a `@Published` property off-main — the same SwiftUI/Combine crash class (`filteredcellModels`/`cellModels` mutation off main -> "Publishing changes from background threads") this codebase has hit before. The three handlers already mutate `@Published` state via `insertRow`/`deleteRow`/`moveRow` without dispatching to main, so this isn't newly introduced, but it does add one more mutation path onto the same landmine. Worth either wrapping `handleFieldValueRowCreate`/`Delete`/`Move` in `DispatchQueue.main.async` (matching `handleFieldValueRowUpdate`), or confirming `change(changes:)` callers are guaranteed to already be on the main thread.

VishnuBishnoi (commented):
Independent full-diff pass (all commits, model+handlers+view models+tests). The two previously-flagged blocking validation gaps (hidden-required cells in Table/Collection) and the nil-action cellVisibilityLogic fallback are all fixed and covered by tests — verified in the current code, not just the commit messages. One new observation below; otherwise this is solid, well-tested work with no breaking public API changes.

Sources/JoyfillUI/ViewModels/DocumentEditor.swift
vivekjyani: Agreed this is worth fixing, but it's a pre-existing gap across all four handlers, not something this PR introduces — and I confirmed there's no reachable off-main call path with current call sites, so it's not an active bug. I'd rather track this as its own issue and address all four handlers consistently (with a documented threading contract on change(changes:)) rather than patch just the paths this PR happens to touch.

Sources/JoyfillUI/ViewModels/DocumentEditor.swift
```diff
@@ -256,6 +263,30 @@ public class DocumentEditor: ObservableObject {
...
+
+    public func cellsNeedToBeRefreshed(fieldID: String, editedColumnID: String, row: ValueElement) -> [String] {
+        return conditionalLogicHandler.cellsNeedToBeRefreshed(fieldID: fieldID, editedColumnID: editedColumnID, row: row)
+    }
+
+    public func cellsNeedToBeRefreshed(fieldID: String, schemaID: String, editedColumnID: String, row: ValueElement) -> [String] {
+        return conditionalLogicHandler.cellsNeedToBeRefreshed(fieldID: fieldID, schemaID: schemaID, editedColumnID: editedColumnID, row: row)
+    }
+
+    public func addCellVisibilityForRow(fieldID: String, row: ValueElement) {
```
vivekjyani: Make a single public method for both table/collection

Sources/JoyfillUI/ViewModels/DocumentEditor.swift
```diff
@@ -256,6 +263,30 @@ public class DocumentEditor: ObservableObject {
     public func shouldShowColumn(columnID: String, fieldID: String, schemaKey: String? = nil) -> Bool {
         return conditionalLogicHandler.shouldShow(columnID: columnID, fieldID: fieldID, schemaKey: schemaKey)
     }
+
+    public func shouldShowCell(columnID: String, fieldID: String, row: ValueElement) -> Bool {
+        return conditionalLogicHandler.shouldShowCell(fieldID: fieldID, columnID: columnID, row: row)
+    }
+
+    public func cellsNeedToBeRefreshed(fieldID: String, editedColumnID: String, row: ValueElement) -> [String] {
```
vivekjyani: Make a single public method for both table/collection

Sources/JoyfillUI/ViewModels/DocumentEditor.swift
```diff
@@ -738,10 +796,16 @@ extension DocumentEditor {
...
     func refreshDependent(for fieldID: String) {
-        let refreshFields = conditionalLogicHandler.fieldsNeedsToBeRefreshed(fieldID: fieldID)
+        requiredLogicHandler.cellRequiredNeedRefreshForPageField(pageFieldID: fieldID)
+        var refreshFields = Set(conditionalLogicHandler.fieldsNeedsToBeRefreshed(fieldID: fieldID))
+        refreshFields.formUnion(requiredLogicHandler.fieldsNeedsToBeRefreshed(fieldID: fieldID))
         for fieldId in refreshFields {
             refreshField(fieldId: fieldId)
         }
+        for (tableFieldID, columnIDs) in conditionalLogicHandler.cellsNeedRefreshForPageField(pageFieldID: fieldID) {
+            let fieldType = field(fieldID: tableFieldID)?.fieldType ?? .table
```
vivekjyani: Please use if let for fieldType

Sources/JoyfillUI/View/Fields/TableView/TableViewModel.swift
```diff
@@ -78,6 +81,8 @@ class TableViewModel: ObservableObject, TableDataViewModelProtocol {
                         fieldTableColumn.id == cellDataModel.id
                     }) {
                         self?.tableDataModel.valueToValueElements = self?.cellDidChange(rowId: rowID, colIndex: colIndex, cellDataModel: cellDataModel, isNestedCell: false)
+                        self?.applyCellVisibilityRefresh(rowId: rowID, editedColumnID: cellDataModel.id)
```
vivekjyani: Can we call these two methods directly inside `cellDidChange` instead of calling them in multiple places?
```
self.applyCellVisibilityRefresh(rowId: rowId, editedColumnID: cellDataModel.id)
self.applyCellRequiredRefresh(rowId: rowId, editedColumnID: cellDataModel.id)
```
This could help centralize the refresh logic and avoid duplicate calls.

Sources/JoyfillUI/View/Fields/TableView/TableModalView.swift
```diff
@@ -26,8 +26,9 @@ struct TableRowView : View {
                 .background(Color.rowSelectionBackground(isSelected: isSelected, colorScheme: colorScheme))
                 .border(Color.tableCellBorderColor)
             }
+            let rowElement = viewModel.rowElement(forRowID: rowDataModel.rowID)
             ForEach($rowDataModel.cells, id: \.id) { $cellModel in
-                let showRequired = viewModel.tableDataModel.requiredColumnIDs.contains(cellModel.data.id) && !cellModel.data.isCellFilled
+                let showRequired = viewModel.isCellRequired(columnID: cellModel.data.id, row: rowElement) && !cellModel.data.isCellFilled
```
vivekjyani: Can you check why we need to pass the entire row object to this method?
If the full row isn’t necessary, please update the method to pass only the required data. So that we don’t have to calculate the entire row data just to call this method.
This should also help improve performance.

Sources/JoyfillUI/View/Fields/TableView/TableViewModel.swift
```diff
@@ -543,6 +560,110 @@ class TableViewModel: ObservableObject, TableDataViewModelProtocol {
...
 }
 
+// MARK: - Required and Cell Visibility Logic
+extension TableViewModel {
+    /// Per-cell required-ness for the live grid border. Honours `cellRequiredLogic`
+    /// (resolved against this row's sibling cells), then the column's `requiredLogic`,
+    /// then the static `required` flag. Falls back to the column-wide set if the row
+    /// value can't be resolved.
+    func isCellRequired(columnID: String, rowID: String) -> Bool {
+        return isCellRequired(columnID: columnID, row: rowElement(forRowID: rowID))
```
vivekjyani: Can you check why we need to pass the entire row object to this method?
If the full row isn’t necessary, please update the method to pass only the required data. So that we don’t have to calculate the entire row data just to call this method.
This should also help improve performance.

Sources/JoyfillUI/View/Fields/CollectionView/CollectionQuickView.swift
```diff
@@ -176,9 +176,11 @@ struct CollectionQuickView : View {
                     HStack(alignment: .top, spacing: 0) {
                         ForEach(Array(viewModel.tableDataModel.tableColumns.prefix(3).enumerated()), id: \.offset) { index, col in
                             let cell = rowDataModel.cells[index]
-                            
+                            let isHidden = viewModel.isCellHidden(columnID: cell.data.id, row: viewModel.rowToValueElementMap[rowDataModel.rowID])
```
vivekjyani: We do not need this whole row object , check carefully please

Sources/JoyfillUI/ViewModels/ConditionalLogicHandler.swift
```diff
@@ -612,3 +628,185 @@ extension ColumnTypes {
...
+        let didChange = cellVisibilityMap[fieldID]?[cellID] != newValue
+        cellVisibilityMap[fieldID, default: [:]][cellID] = newValue
+        return didChange
+    }
+
+    func removeCellVisibilityForRow(fieldID: String, rowID: String) {
+        cellVisibilityMap[fieldID] = cellVisibilityMap[fieldID]?.filter { $0.key.rowID != rowID }
+    }
+
+    func shouldShowCell(fieldID: String, columnID: String, row: ValueElement) -> Bool {
```
vivekjyani: We do not need full row object in params, check again!!!

JoyfillSwiftUIExample/JoyfillTests/CellVisibilityLogicTest.swift
```diff
@@ -0,0 +1,1161 @@
+import XCTest
+import Foundation
+import JoyfillModel
+@testable import Joyfill
+
```
vivekjyani, line 5: Missing tests 
1. External row updates 
2. validate() method tests 
3. page duplicate and delete tests with these functionalities

JoyfillSwiftUIExample/JoyfillTests/RequiredLogicTests.swift
```diff
@@ -0,0 +1,795 @@
+import XCTest
+import Foundation
+import JoyfillModel
+@testable import Joyfill
+
```
vivekjyani, line 5: Missing tests

External row updates
validate() method tests
page duplicate and delete tests with these functionalities

Sources/JoyfillUI/ViewModels/DocumentEditor.swift
vivekjyani: Done — switched to if let. Worth noting this wasn't only cosmetic: the old ?? .table fallback meant an unresolvable tableFieldID still reached valueDelegate(for:fieldType:), which creates a view model from that type and caches it in delegateMap — so a wrong default would have been persisted rather than harmlessly ignored. Now the entry is skipped.

Sources/JoyfillUI/ViewModels/DocumentEditor.swift
vivekjyani: Done — cellsNeedToBeRefreshed and addCellVisibilityForRow are each a single method now, taking schemaID: String? = nil, which matches how cellRequiredNeedToBeRefreshed / addCellRequiredForRow are already declared just below. The table-vs-schema column lookup moved into a private columns(fieldID:schemaID:) helper in ConditionalLogicHandler. Since the new parameter is defaulted, every existing call site — including the public API for host apps — compiles unchanged. Build passes and all 78 tests in CellVisibilityLogicTest/RequiredLogicTests still pass.

Sources/JoyfillUI/ViewModels/DocumentEditor.swift
vivekjyani: Done — cellsNeedToBeRefreshed and addCellVisibilityForRow are each a single method now, taking schemaID: String? = nil, which matches how cellRequiredNeedToBeRefreshed / addCellRequiredForRow are already declared just below. The table-vs-schema column lookup moved into a private columns(fieldID:schemaID:) helper in ConditionalLogicHandler. Since the new parameter is defaulted, every existing call site — including the public API for host apps — compiles unchanged. Build passes and all 78 tests in CellVisibilityLogicTest/RequiredLogicTests still pass.

Sources/JoyfillUI/View/Fields/TableView/TableModalView.swift
vivekjyani: Good catch — fixed. Removed the per-row `rowElement(forRowID:)` lookup and now pass `rowID: rowDataModel.rowID` straight through. Required-ness is read from the per-cell cache, so this is an O(1) dictionary hit per cell instead of a linear scan over `valueToValueElements` on every render pass.

Sources/JoyfillUI/View/Fields/TableView/TableViewModel.swift
vivekjyani: Done. Dropped both `row: ValueElement?` overloads (`isCellRequired` and `isCellHidden`) and kept only the `rowID:` versions, which forward to `documentEditor.isCellRequired(…rowID:)` / `shouldShowCell(…rowID:)` and read the cached per-cell state directly. All internal callers (cell builders and the visibility-flip paths) were migrated as well. `rowElement(forRowID:)` is still there, but it's now only used by `cellsNeedToBeRefreshed`/`cellRequiredNeedToBeRefreshed`, which have to evaluate logic against the row's sibling cell values.

Sources/JoyfillUI/View/Fields/CollectionView/CollectionQuickView.swift
vivekjyani: Fixed — the `rowToValueElementMap[rowDataModel.rowID]` lookup inside the column loop is gone; it's `isCellHidden(columnID: cell.data.id, rowID: rowDataModel.rowID)` now.

JoyfillSwiftUIExample/JoyfillTests/CellVisibilityLogicTest.swift
vivekjyani: Added — 9 new tests in this file covering all three areas:

**External row updates**
- `testExternalTableRowUpdateImmediatelyUpdatesConditionalCells`
- `testExternalTablePartialUpdateHidesOnlyTargetRowAndPreservesOtherCells`
- `testExternalCollectionRowUpdateImmediatelyUpdatesConditionalCells`
- `testExternalNestedCollectionRowUpdateReevaluatesConditionalCells`
- `testExternalPageFieldUpdateRefreshesTableConditionalCells`
- `testExternalPageFieldUpdateRefreshesCollectionConditionalCells`

**validate()**
- `testValidateTreatsHiddenRequiredNestedCollectionCellAsValid`, plus `validate()` assertions folded into the existing visibility scenarios so a hidden cell never fails validation.

**Page duplicate / delete**
- `testDuplicatePageWithValuesUsesCopiedPageValuesForTableAndNestedCollectionVisibility`
- `testDuplicatePageWithoutValuesEvaluatesNewTableAndNestedCollectionRowsFromPageTwoValues`

Heads-up: the external-rowUpdate and page-duplication tests are currently red, and that's intentional — they document two real cache gaps (an external `field.value.rowUpdate` doesn't refresh `cellVisibilityMap`, and page duplication doesn't seed the cache for the duplicated page's rows). Fixing those touches the refresh/duplication paths rather than this feature, so it's going into a follow-up PR.

JoyfillSwiftUIExample/JoyfillTests/RequiredLogicTests.swift
vivekjyani: Added — 11 new tests in this file covering all three areas:

**External row updates**
- `testExternalRowUpdateImmediatelyReevaluatesRequiredCells`
- `testExternalTablePartialUpdateMakesOnlyTargetRowOptionalAndPreservesOtherCells`
- `testExternalCollectionRowUpdateImmediatelyReevaluatesRequiredCells`
- `testExternalCollectionPartialUpdateMakesOnlyTargetRowOptionalAndPreservesOtherCells`
- `testExternalNestedCollectionRowUpdateReevaluatesRequiredCells`
- `testExternalPageFieldUpdateReevaluatesTableRequiredCells`
- `testExternalPageFieldUpdateReevaluatesCollectionRequiredCells`

**validate()**
- `validate()` assertions added across the required-logic scenarios, so a hidden-but-required cell is treated as valid and an enforced one still blocks.

**Page duplicate / delete**
- `testRequiredRulesFollowRowsThroughAddDuplicateAndDelete`
- `testCollectionRequiredRulesFollowRowsThroughAddDuplicateAndDelete`
- `testDuplicatePageWithValuesUsesCopiedPageValuesForTableAndNestedCollectionRequiredLogic`
- `testDuplicatePageWithoutValuesEvaluatesNewTableAndNestedCollectionRowsFromPageTwoValues`

Same heads-up as on the visibility file: the external-rowUpdate and page-duplication tests are deliberately red. They pin down two real gaps — an external `field.value.rowUpdate` doesn't refresh `cellRequiredMap`, and page duplication doesn't seed it for the duplicated page's rows. Both fixes live in the refresh/duplication paths, so they're going into a follow-up PR.

Sources/JoyfillUI/View/Fields/TableView/TableViewModel.swift
vivekjyani: Done — folded both refreshes into `cellDidChange` in `TableViewModel`.

`cellDidChange` now also commits `tableDataModel.valueToValueElements` itself before refreshing, since `applyCellVisibilityRefresh` reads the edited value back out of it via `rowElement(forRowID:)` — refreshing before the caller assigned it would evaluate the logic against the pre-edit value. All three call sites are now a bare `cellDidChange(...)`.

Two spots intentionally still call the refreshes directly:
- The bulk-edit path, which goes through `documentEditor.bulkEdit` and never touches `cellDidChange`.
- `CollectionViewModel`, which already had both calls inside its own `cellDidChange`.

Side benefit: the external `field.value.rowUpdate` path goes through `updateUIModels` → `cellDidChange`, so it now picks up the refresh automatically — that fixed four previously failing external-row-update tests.

Sources/JoyfillUI/ViewModels/ConditionalLogicHandler.swift
vivekjyani: Fixed

Sources/JoyfillUI/ViewModels/ConditionalLogicHandler.swift
```diff
@@ -612,3 +663,233 @@ extension ColumnTypes {
...
+    public func isFieldRequired(fieldID: String) -> Bool {
+        return requiredFieldMap[fieldID] ?? false
+    }
+
+    public func isColumnRequired(columnID: String, fieldID: String, schemaKey: String? = nil) -> Bool {
+        let columnSchemaID = ColumnSchemaID(columnID: columnID, schemaID: schemaKey)
+        return showColumnLogicMap[fieldID]?.requiredColumnMap[columnSchemaID] ?? true
+    }
+    
+    public func isCellRequired(columnID: String, fieldID: String, schemaKey: String? = nil, rowID: String) -> Bool {
```
vivekjyani: `isCellRequired` defaults to `false` (not required) on a cache miss, while `isColumnRequired` just above defaults to `true` (required) on the same kind of miss. That asymmetry used to be inert bookkeeping, but `ValidationHandler.validateTableField`/`validateCollectionSchema` now call `isCellRequired` directly to gate form submission. If `cellRequiredMap` is ever missing an entry for a given (rowID, columnID) — a timing gap between row creation and cache rebuild, a row added through a path that skips `addCellLogicForNewRow` — a genuinely required, empty cell will silently validate as `.valid` and the form can be submitted with missing required data. The old static `column.required ?? false` had no cache to miss; this is a new failure class introduced by moving required-ness into a lazily-populated cache. Worth either matching `isColumnRequired`'s fail-safe-to-true default, or asserting/logging on a cache miss so this can't fail silently.

Sources/JoyfillUI/ViewModels/DocumentEditor.swift
```diff
@@ -1491,3 +1528,26 @@ extension DocumentEditor {
...
+        return conditionalLogicHandler.isFieldRequired(fieldID: fieldID)
+    }
+
+    /// Effective column-wide required-ness, honouring the column's `requiredLogic`.
+    public func isColumnRequired(columnID: String, fieldID: String, schemaKey: String? = nil) -> Bool {
+        return conditionalLogicHandler.isColumnRequired(columnID: columnID, fieldID: fieldID, schemaKey: schemaKey)
+    }
+
+    /// Effective required-ness of a single cell, honouring `cellRequiredLogic` (per-row) then `requiredLogic`.
+    public func isCellRequired(columnID: String, fieldID: String, schemaKey: String? = nil, rowID: String) -> Bool {
```
vivekjyani: This wrapper accepts `schemaKey` but never forwards it to `conditionalLogicHandler.isCellRequired`, which is called with only `columnID`/`fieldID`/`rowID`. `CellID` has no schema component, so nested-schema collections rely entirely on `rowID` global uniqueness to disambiguate — which likely holds today, but the parameter existing and being silently dropped is a landmine for schema-scoped lookups added later, or the rare case a rowID collides across schemas. Either wire it through (once `CellID` carries a schema component) or drop the parameter so the signature doesn't imply disambiguation that isn't happening.

vivekjyani (commented):
Two correctness findings on the new required-ness cache, both stemming from an inconsistent fail-open/fail-closed default. Details inline.

Sources/JoyfillUI/ViewModels/DocumentEditor.swift
vivekjyani: Fixed, remove extra `schemaKey` parameter

Sources/JoyfillUI/ViewModels/ConditionalLogicHandler.swift
vivekjyani: Fixed

VishnuBishnoi (approved):
LGTM

PR #359, opened 2026-07-30
NO-1508: Fix dead dropdown sheet caused by List row recycling

VishnuBishnoi:
## Context

Dropdown fields rendered inside `FormView`'s SwiftUI `List` intermittently stopped opening their options sheet — tapping did nothing, most noticeably after scrolling, and once a row died it stayed dead while its siblings kept working.

Root cause: `List` on iOS is backed by `UITableView`, which recycles a row's hosting controller as the row scrolls off and back on screen. A `.sheet` anchored anywhere inside that row can be orphaned by the recycling and stop presenting entirely.

Two earlier attempts on this branch were dead ends and have been reverted:

- `907ab4d7` replaced `List` with `ScrollView { LazyVStack }` to remove recycling altogether. Reverted in `0563d447` — it fixes the bug but gives up `List` behaviour the form relies on.
- The later `DropdownView`-scoped attempt swapped the `Bool` presentation flag for an `Identifiable` item and moved the `.sheet` from the `Button` to the row's `VStack`. Both positions are still **inside the recycled row**, so the same hosting controller was still presenting — structurally the same as the `DispatchQueue.main.async { isSheetPresented = true }` attempt that was already tried and didn't work.

## Changes

The fix is to stop the row from owning a sheet at all. **+55 / −16 across 2 files.**

**[`FormView.swift`](Sources/JoyfillUI/View/FormView.swift)**
- New `FieldSheetPresentation: Identifiable` — carries the field id only, no model.
- One `@State activeFieldSheet` and one `.sheet(item:)` on the `List`, replacing one `.sheet` modifier and one `@State` flag *per dropdown row*.
- `fieldSheet(for:)` → `dropdownOptionsSheet(for:)` → `dropdownModel(for:)`. The model is re-resolved from `listModels` on every evaluation rather than captured, so the option list stays live if conditional logic or a formula rewrites `options` while the sheet is open.
- Selection routes through `onChange` instead of row-local state.

**[`DropdownView.swift`](Sources/JoyfillUI/View/Fields/DropdownView.swift)**
- Reverted to `main`, then: takes a `Binding<FieldSheetPresentation?>`, writes its field id on tap, owns no sheet.
- Kept `.contentShape(Rectangle())` on the option row from the previous attempt — that one does real work on the padded area.

## Why this approach

The `.sheet` has to be anchored on a view that is never recycled, and the `List` is the nearest one. Nothing else about the presentation mechanism matters: changing the flag's *type*, or its *position within the row*, leaves the same hosting controller presenting. That is why the two earlier attempts didn't hold.

Naming is deliberately field-agnostic. [`ImageView.swift:149`](Sources/JoyfillUI/View/Fields/ImageView.swift:149) has the same row-owned-sheet shape; adopting this is a branch in `fieldSheet(for:)` plus swapping its `@State showMoreImages` for a write to `activeFieldSheet`. **`ImageView` is not touched here** — this change is dropdown-only, on purpose.

No `Kind` discriminator was added to `FieldSheetPresentation`: the field type is already in `listModels` via `FieldListModelType`, so `fieldSheet(for:)` can switch on the model it looks up by id.

### Performance

The diff moves state up a level, so worth stating explicitly:

- **Cheaper:** on a dropdown-heavy page this collapses N `.sheet` modifiers and N `@State` flags into one. On `3000-fields.json` (662 dropdowns) that's 662 → 1.
- **Not a regression:** writing `activeFieldSheet` invalidates `FormView.body` rather than a single row — but the tap already did that. `FormView.onFocus` ([`FormView.swift:357`](Sources/JoyfillUI/View/FormView.swift:357)) unconditionally writes `currentFocusedFieldsID`, also `@State` on `FormView`, in the same button action. The two coalesce into one transaction.
- **One new cost:** `dropdownModel(for:)` is a linear scan over `listModels`, called once per option per sheet body build. Zero when no sheet is open; roughly 40k string comparisons on a 3000-field page with a 14-option dropdown. Sub-millisecond. Removable by capturing the model instead, at the cost of reintroducing a staleness risk in sheet content that hasn't been ruled out.

## Screenshot / video

None — no visual change. See the verification note below.

## Public API

**No public API change.** `DropdownView`, `DropDownOptionList`, and the new `FieldSheetPresentation` are all internal. `DropdownView.init` gained a parameter but the type isn't reachable outside the module.

**No schema impact.** The emitted `FieldChangeData` payload is unchanged — still `ValueUnion.string(newValue ?? "")` — so nothing in JoyDoc parsing, validation, or `joyfill-schema` is affected.

## Test coverage

**Not covered by an automated test, and the existing suite cannot catch this bug.**

`JoyfillSwiftUIExample/JoyfillUITests/DropdownField/DropdownFieldUITestCases.swift` has 18 tests covering selection, clearing, conditional logic, payload shape, readonly, and focus/blur — but its fixture has only 4 fields, which never fills a screen, so `UITableView` never recycles anything. `testDropdownRetainsValueAfterNavigation` is the closest: it swipes up and down, but only asserts the label and never re-taps to open.

That is why this shipped to production undetected.

## Notes for reviewers

- **The runtime repro has not been run.** The package builds clean for iOS Simulator; that is all that has been verified. `Sample JSONs/others/3000-fields.json` (662 dropdowns / 2898 fields) is the best vehicle in the repo — well past what's needed to force cell reuse. Scroll several screens down, then tap dropdowns on the way back up. This should pass before merge.
- **Watch label latency specifically.** Selection now reaches the label via a model round-trip (`onChange` → `documentEditor.updateField` → `refreshField` → `pageFieldModels` → the row's `onChange(of: dropdownDataModel.dropdownValue)`) rather than from row-local `@State`. That path is synchronous with no dispatch, but it is the behavioural difference most likely to surface.
- The echo-loop comment above the guard in `DropdownView`'s `onChange` matters more under this design than it did on `main` — the round-trip above is exactly what that guard stops from re-emitting. Please don't remove it.
- Scope is dropdown-only. Other sheet-presenting fields in the same `List` were left alone; `ImageView` is the one worth watching for the same symptom.

### DIFF
--- a/Sources/JoyfillUI/View/FormView.swift
+++ b/Sources/JoyfillUI/View/FormView.swift
@@ -286,15 +286,20 @@ struct FormView: View {
 
     var body: some View {
         ScrollViewReader { proxy in
-            List($listModels, id: \.wrappedValue.fieldIdentifier.fieldID) { $listModel in
-                if documentEditor.shouldShow(fieldID: listModel.fieldIdentifier.fieldID) {
-                    fieldView(listModelBinding: $listModel)
-                        .listRowSeparator(.hidden)
-                        .buttonStyle(.borderless)
+            ScrollView {
+                LazyVStack(alignment: .leading, spacing: 12) {
+                    ForEach($listModels, id: \.wrappedValue.fieldIdentifier.fieldID) { $listModel in
+                        if documentEditor.shouldShow(fieldID: listModel.fieldIdentifier.fieldID) {
+                            fieldView(listModelBinding: $listModel)
+                                .id(listModel.fieldIdentifier.fieldID)
+                                .buttonStyle(.borderless)
+                        }
+                    }
                 }
+                .padding(.horizontal, 16)
+                .padding(.vertical, 8)
             }
             .environment(\.navigationFocusFieldId, documentEditor.navigationFocusFieldId)
-            .listStyle(PlainListStyle())
             .modifier(KeyboardDismissModifier())
             .onChange(of: $currentFocusedFieldsID.wrappedValue) { newValue in
                 guard newValue != nil else { return }
--- a/Sources/JoyfillUI/View/FormView.swift
+++ b/Sources/JoyfillUI/View/FormView.swift
@@ -232,17 +232,8 @@ extension FieldListModelType {
     }
 }
 
-/// Identifies which field's modal sheet is currently open, by field id.
-///
-/// Sheets for fields inside `FormView`'s `List` must be anchored on the `List`,
-/// not on the row that triggers them: `List` is backed by `UITableView`, which
-/// recycles a row's hosting controller as it scrolls off and back on screen, and
-/// a row-owned `.sheet` can be orphaned by that recycling and stop presenting
-/// entirely (NO-1508). The `List` itself is never recycled.
-///
-/// Only `DropdownView` is hoisted today. `ImageView` has the same row-owned-sheet
-/// shape and can adopt this by writing its field id here instead of flipping a
-/// local flag, plus a branch in `fieldSheet(for:)`.
+/// Which field's sheet is open. Must live on FormView's List, never on a row —
+/// UITableView recycling orphans row-owned sheets and kills them (NO-1508).
 struct FieldSheetPresentation: Identifiable {
     let id: String
 }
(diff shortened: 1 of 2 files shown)


VishnuBishnoi (commented):
_Superseded — see the updated PR description._

Sources/JoyfillUI/View/FormView.swift
```diff
@@ -284,6 +291,41 @@ struct FormView: View {
...
+            if #available(iOS 16, *) {
+                DropDownOptionList(dropdownDataModel: model, selectedDropdownValueID: selection)
+                    .presentationDetents([.medium])
+            } else {
+                DropDownOptionList(dropdownDataModel: model, selectedDropdownValueID: selection)
+            }
+        }
+    }
+
+    private func dropdownModel(for fieldID: String) -> DropdownDataModel? {
```
vivekjyani: This linear scan over `listModels` runs once per option per sheet render (called from `dropdownOptionsSheet`'s `get` on every access) — the PR description already quantifies this as ~40k comparisons / sub-millisecond on the 3000-field fixture, so not a real-world regression. If you want to drop the traversal entirely, `FieldSheetPresentation` could carry the resolved `DropdownDataModel` directly (captured at tap time in `DropdownView`) instead of just the field id — `dropdownOptionsSheet` then takes the model straight from the presentation, O(1), no re-scan. Tradeoff: the sheet would show a snapshot of `options` as of tap time instead of staying live if a formula/conditional-logic rule rewrites `options` while the sheet is already open. Worth a deliberate call rather than a default — happy either way given the measured cost is negligible.

vivekjyani (commented):
One inline note on the perf tradeoff flagged in the PR description.

Sources/JoyfillUI/View/FormView.swift
vivekjyani: Fixed

VishnuBishnoi (commented):
LGTM

